PR #2, opened 2021-03-29
Fix compilation in rust 1.46+

abizjak:
## Purpose

Enable compilation with rust versions 1.46 and above.

## Changes

- The main change that is needed to enable compilation with rust 1.46+ is to add 
+[profile.release.build-override]
+opt-level = 3
into the Cargo.toml file of the wallet to make it build the dependencies only once.

- Move the wallet library out of rust-src into mobile_wallet. This accounts for the wast majority of the changed lines. The rationale for this is that rust-src is the core library, and the wallet is just a wrapper for the mobile wallet, so
there is no need for it to be where it was. This should have no external effect since the wallet library's API stays the same.

Other changes are to make clippy happy with newer rust versions.

## Checklist

      hard-to-understand areas.


mobile_wallet/src/lib.rs
```diff
@@ -1,4 +1,767 @@
...
+                sig: signature.to_bytes().to_vec(),
+            });
+        }
+        out.insert(cred_index, cred_sigs);
+    }
+    TransactionSignature { signatures: out }
+}
+
+/// Create a JSON encoding of an encrypted transfer transaction.
+fn create_encrypted_transfer_aux(input: &str) -> anyhow::Result<String> {
```
td202, line 63: Why not add `use anyhow::Result`?

td202 (approved):

mobile_wallet/src/lib.rs
abizjak: It's because Result is an `std` type that is automatically imported, so I think it is confusing to import anyhow::Result unqualified.

And this is how it is in the rest of the project now, so this follows the code style as well.

td202: If `anyhow::Result` wasn't an alias for `std::result::Result`, I'd agree it was confusing.

abizjak: Right, but it's not just an alias. It has a default value for the error type. Lots of other crates define their own Result type (e.g., io::Result) which is a similar alias, but with a fixed error type. The error messages you get when using them incorrectly are needlessly confusing when looking at function signatures that look the same. The common way to use such types is to qualify them.

PR #4, opened 2021-03-30
BLS hash follows BLS12381G1_XMD:SHA-256_SSWU_RO_

SimonKamp:
## Purpose

The purpose of this pull request was to change the way we hash to the G1 subgroup of the BLS-12-381 curve to follow the suite BLS12381G1_XMD:SHA-256_SSWU_RO_ specified at https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-10#section-8.8.1. 

## Changes

The hash_to_group function was changed to follow https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-10#section-8.8.1. 
Concretely, it now uses the Simplified SWU specified at https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-10#section-6.6.3.
Test were added that compares the output with https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-10#appendix-J.9.1.

## Checklist

      hard-to-understand areas.


rust-src/curve_arithmetic/benches/hash_bench.rs
```diff
@@ -40,5 +40,5 @@ pub fn bench_hash_to_curve(c: &mut Criterion) {
 // }
 
 // criterion_group!(hash_to_fq, bench_hash_to_fq);
-criterion_group!(hash_to_curve, bench_hash_to_curve);
-criterion_main!(hash_to_curve);
+criterion_group!(hash_to_curve_, bench_hash_to_curve);
+criterion_main!(hash_to_curve_);
```
abizjak, line 44: Why the underscore?

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -496,41 +485,139 @@ pub(crate) const K4: [[u64; 6]; 16] = [
...
+/// It follows the steps
+///    1. u = hash_to_field(msg, 2)
+///    2. Q0 = map_to_curve(u[0])
+///    3. Q1 = map_to_curve(u[1])
+///    4. R = Q0 + Q1              
+///    5. P = clear_cofactor(R) = h_eff * R   # Clearing cofactor
+///    6. return P,
+/// where the choices of hash_to_field, map_to_curve and h_eff are as described in https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-10#section-8.8.1.
+pub fn hash_to_curve(msg: &[u8]) -> G1 {
+    let dst = "QUUX-V01-CS02-with-BLS12381G1_XMD:SHA-256_SSWU_RO_".as_bytes();
```
abizjak, line 498: If you just need the bytes I'd recommend using `b"QUUX-V01-CS02-with-BLS12381G1_XMD:SHA-256_SSWU_RO_"`.

Is the final underscore intended?

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -496,41 +485,139 @@ pub(crate) const K4: [[u64; 6]; 16] = [
...
+    r.mul_assign(15132376222941642753); // Clearing cofactor with h_eff = 15132376222941642753
+    r // This now guarantied to be in G1
+}
+
+/// Implements https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-10#section-6.6.3
+/// It uses the function sswu_3mod4 to get xn, xd, y such that (xn/xd : y : 1)
+/// is a point on E' (written in Jacobian coordinates). Since inversions are
+/// expensive, we use the fact that (xn/xd : y : 1) on E' <=> (xn xd : xd^3y :
+/// xd) on E' when using Jacobian coordinates. It returns iso_11((xn xd : xd^3y
+/// : xd)) which is then guarentied to lie on E (but not necessarily in G1).
```
abizjak, line 515:
```suggestion
/// : xd)) which is then guaranteed to lie on E (but not necessarily in G1).
```

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -496,41 +485,139 @@ pub(crate) const K4: [[u64; 6]; 16] = [
...
+fn map_to_curve(u: Fq) -> G1 {
+    let (xn, xd, mut y, _) = sswu_3mod4(u);
+    let mut x = xn;
```
abizjak, line 518: You should just write
```rust
let (mut x, xd, mut y, _)
```
to reduce the mental overhead.

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -496,41 +485,139 @@ pub(crate) const K4: [[u64; 6]; 16] = [
...
+    x.mul_assign(&xd);
+    y.mul_assign(&xd);
+    y.mul_assign(&xd);
+    y.mul_assign(&xd);
+    let (xiso, yiso, z) = iso_11(x, y, xd);
+    from_coordinates_unchecked(xiso, yiso, z).expect("should not happen")
```
abizjak, line 524: Perhaps document why it should not happen.

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -496,41 +485,139 @@ pub(crate) const K4: [[u64; 6]; 16] = [
...
+fn hash_to_field(msg: &[u8], dst: &[u8]) -> (Fq, Fq) {
+    let (u_0, u_1, u_2, u_3) = expand_message_xmd(msg, dst);
+
+    (fq_from_bytes(&u_0, &u_1), fq_from_bytes(&u_2, &u_3))
+}
+
+// Interpret input as integers (big endian)
+// Return (left*2^256 + right) as Fq
+fn fq_from_bytes(left_bytes: &[u8; 32], right_bytes: &[u8; 32]) -> Fq {
+    fn le_u64s_from_be_bytes(bytes: &[u8]) -> Fq {
```
abizjak, line 538: This function is always applied to an array of 32 elements, so we can reflect this in its type.

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -496,41 +485,139 @@ pub(crate) const K4: [[u64; 6]; 16] = [
...
+        let mut digits = [0u64; 6];
+
+        for i in 0..4 {
+            digits[3 - i] = u64::from_be_bytes(
+                bytes[i * 8..(i + 1) * 8]
+                    .try_into()
+                    .expect("Chunk Size is always 8."),
+            );
+        }
```
abizjak, line 547: I think it'd be clearer to use the `chunks` method here like we do when deserializing field elements.

```rust
for (place, chunk) in digits.iter_mut().zip(bytes.chunks(8).rev()) {
    *place = u64::from_be_bytes(chunk.try_into().expect("Chunk Sie is always 8"))
}
```

abizjak: But you decide.

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -496,41 +485,139 @@ pub(crate) const K4: [[u64; 6]; 16] = [
...
+    // b_4 = H(strxor(b_2, b_3)  || I2OSP(4, 1) || DST_prime)
+    let mut h = Sha256::new();
+    let xor: Vec<u8> = b_0.iter().zip(b_3.iter()).map(|(x, y)| x ^ y).collect();
+    h.update(xor);
+    h.update(vec![4]);
+    h.update(dst_prime);
+    let mut b_4: [u8; 32] = [0u8; 32];
+    b_4.copy_from_slice(h.finalize().as_slice());
+
+    // todo use loop
```
abizjak, line 618: What is this for?

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -496,41 +485,139 @@ pub(crate) const K4: [[u64; 6]; 16] = [
...
+fn expand_message_xmd(msg: &[u8], dst: &[u8]) -> ([u8; 32], [u8; 32], [u8; 32], [u8; 32]) {
+    // DST_prime = DST || I2OSP(len(DST), 1)
+    let mut dst_prime = dst.to_vec();
+    dst_prime.push(dst.len().try_into().unwrap()); // panics if dst is more than 255 bytes
+                                                   // msg_prime = Z_pad || msg || l_i_b_str || I2OSP(0, 1) || DST_prime
+
+    // b_0 = H(msg_prime)
+    let mut h = Sha256::new();
+    // todo a possible optimization here would be to save the state of H(Z_pad)
+    h.update(vec![0; 64]); // z_pad = I2OSP(0, 64), 64 is the input block size of Sha265
```
abizjak, line 575: I think we should avoid the vector allocation here and just use the array `[0u8; 64]` instead.

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -496,41 +485,139 @@ pub(crate) const K4: [[u64; 6]; 16] = [
...
+    h.update(msg);
+    h.update(vec![0, 128]); // l_i_b_str = I2OSP(128, 2)
```
abizjak, line 577: And the same here.

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -496,41 +485,139 @@ pub(crate) const K4: [[u64; 6]; 16] = [
...
+    h.update(vec![0]);
```
abizjak, line 578: And here. I think it is also quite useful to document that `0` is a `u8` type, i.e., write `0u8` explictly.

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -496,41 +485,139 @@ pub(crate) const K4: [[u64; 6]; 16] = [
...
+    h.update(&dst_prime);
+    let mut b_0: [u8; 32] = [0u8; 32];
+    b_0.copy_from_slice(h.finalize().as_slice());
+
+    // b_1 = H(b_0 || I2OSP(1, 1) || DST_prime)
+    let mut h = Sha256::new();
+    h.update(b_0);
+    h.update(vec![1]);
```
abizjak, line 586: Same comment as above, please be explicit about the type of `1`.

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -496,41 +485,139 @@ pub(crate) const K4: [[u64; 6]; 16] = [
...
+    h.update(&dst_prime);
+    let mut b_1: [u8; 32] = [0u8; 32];
+    b_1.copy_from_slice(h.finalize().as_slice());
+
+    // b_2 = H(strxor(b_0, b_1)  || I2OSP(2, 1) || DST_prime)
+    let mut h = Sha256::new();
+    let xor: Vec<u8> = b_0.iter().zip(b_1.iter()).map(|(x, y)| x ^ y).collect();
+    h.update(xor);
+    h.update(vec![2]);
```
abizjak, line 595: And replace the `vec` with an array.

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -572,141 +659,163 @@ fn from_coordinates_unchecked(x: Fq, y: Fq, z: Fq) -> Result<G1, CurveDecodingEr
...
+    neg_a.negate();
+    xd.mul_assign(&neg_a);
+
+    // 8.   e1 = xd == 0
+    let e1 = xd.is_zero();
+
+    // 9.   xd = CMOV(xd, Z * A, e1)  # If xd == 0, set xd = Z * A
+    if e1 {
+        xd = z;
+        xd.mul_assign(&a); // todo should we use equivalent of CMOV?
```
abizjak, line 720: I think for now we should just document that we do not care about whether this function is constant time or not.

The way it is used it does not matter.

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
```diff
@@ -762,16 +871,16 @@ fn iso_11(x: Fq, y: Fq, z: Fq) -> (Fq, Fq, Fq) {
...
-// Donald E. Knuth. Seminumerical Algorithms, volume 2 of The Art of Computer
-// Programming, chapter 4.6.4. Addison-Wesley, 3rd edition, 1997
-//
-// Evaluates the polynomial with the given coefficients where the i'th
-// coefficient has been multiplied by z^(degree - i) where degree is the degree
-// of the polynomial.
-// z_powers is an array of the even powers of z, ordered [z^2, z^4, ...] The
-// polynomial is evaluated in 'variable'. Note: It's a prerequisite that
-// Fq::from_repr(FqRepr(coefficient[i])) doesn't produce an error!!!
+/// Macro for evaluating polynomials using Horner's rule
```
abizjak, line 874: This comment is a bit out of date.

This is a function, not a macro.

abizjak (commented):
I only have minor comments related to style, documentation, and some minor optimizations.

Otherwise this looks very good.

There need to be related PRs in `concordium-node`, concordium-client, and wallet-proxy to propagate the new hashing function.

rust-src/curve_arithmetic/src/bls12_381_g1hash.rs
eb-concordium: Regarding the underscore, the test(s) will fail without it, since the testvectors from https://tools.ietf.org/html/draft-irtf-cfrg-hash-to-curve-10#appendix-J.9.1 are computed using the underscore.

abizjak: Good, then it is intended :)

SimonKamp: Hardcoding the domain separation string from the test case was intended as a quick way of testing without changing any code relying on hash_to_curve. I think we should consider changing it to a parameter and replace QUUX with concordium or similar. (Or something specific to each context if it used in more than one.)

abizjak: I see, that does indeed make sense. Would you propose some string and parametrize the function instead of hardcoding.

In tests we can then use this one.

rust-src/curve_arithmetic/src/bls12_381_instance.rs
```diff
@@ -196,7 +196,12 @@ impl Curve for G1 {
 
     fn generate_scalar<T: Rng>(csprng: &mut T) -> Self::Scalar { Fr::random(csprng) }
 
-    fn hash_to_group(bytes: &[u8]) -> Self { hash_to_g1(bytes) }
+    fn hash_to_group(bytes: &[u8]) -> Self {
+        hash_to_curve(
+            bytes,
+            b"CONCORDIUM-hashtoG1-with-BLS12381G1_XMD:SHA-256_SSWU_RO",
```
abizjak, line 202: I think it would be useful to factor this out into a constant.

rust-src/curve_arithmetic/src/bls12_381_instance.rs
```diff
@@ -277,7 +282,13 @@ impl Curve for G1Affine {
 
     fn generate_scalar<T: Rng>(csprng: &mut T) -> Self::Scalar { Fr::random(csprng) }
 
-    fn hash_to_group(b: &[u8]) -> Self { hash_to_g1(b).into_affine() }
+    fn hash_to_group(b: &[u8]) -> Self {
+        hash_to_curve(
+            b,
+            b"CONCORDIUM-hashtoG1-with-BLS12381G1_XMD:SHA-256_SSWU_RO",
```
abizjak, line 288: So it is not repeated.

abizjak (approved):
If you could just factor out the constants into an actual constant that is documented.

eb-concordium (approved):

PR #5, opened 2021-03-31
Update genesis dockerfile build for github sources.

abizjak:
Use the public repositories and public docker images.

### DIFF
--- a/scripts/genesis-tools.Dockerfile
+++ b/scripts/genesis-tools.Dockerfile
@@ -10,14 +10,14 @@
 # The builder is based on a base image that has stack, haskell, and rust installed.
 # The only argument is the branch name of a concordium-base repository branch, given as BASE_REF.
 
-FROM 192549843005.dkr.ecr.eu-west-1.amazonaws.com/concordium/base:0.18 AS builder
+FROM concordium/base:0.19 AS builder
 
 # Which branch of concordium-base to build the tools from.
-ARG BASE_REF=master
+ARG BASE_REF=main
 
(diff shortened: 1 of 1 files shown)


scripts/genesis-tools.Dockerfile
```diff
@@ -10,14 +10,14 @@
...
 
 # Which branch of concordium-base to build the tools from.
-ARG BASE_REF=master
+ARG BASE_REF=main
 
-RUN mkdir -p -m 0600 ~/.ssh && ssh-keyscan gitlab.com >> ~/.ssh/known_hosts
+RUN mkdir -p -m 0600 ~/.ssh && ssh-keyscan github.com >> ~/.ssh/known_hosts
 
-RUN --mount=type=ssh git clone --depth 1 --branch ${BASE_REF} git@gitlab.com:Concordium/concordium-base.git
+RUN --mount=type=ssh git clone --depth 1 --branch ${BASE_REF} git@github.com:Concordium/concordium-base.git
```
bisgardo, line 20: `git clone --depth 1 --branch main https://github.com/Concordium/concordium-base.git` would be sufficient.

bisgardo: (can't make multiline suggestions)

abizjak: I can change. 

But the reason I did not is that there might be a case where we'll need a private fork for security fixes, in which case we'll need this again.

Are you strongly against keeping it like this?

bisgardo: Not strongly against it, no.

abizjak: Ok, then I'll keep it if only for myself to not forget how to do this in the future.

PR #6, opened 2021-04-06
Add jenkins pipeline for genesis-tools

bisgardo:
Based on the build pipeline for the base image.

### DIFF
--- a/genesis-tools.Jenkinsfiles
+++ b/genesis-tools.Jenkinsfiles
@@ -2,6 +2,7 @@
 // 'https://gitlab.com/Concordium/infra/jenkins-jobs/-/blob/master/docker_image_genesis_tools.groovy':
 // - image_tag (default: "latest")
 // - base_image_tag
+// - base_ref
 
 pipeline {
     agent any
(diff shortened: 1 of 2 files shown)


genesis-tools.Jenkinsfiles
```diff
@@ -0,0 +1,40 @@
...
+            }
+            steps {
+                sh 'echo $CRED_PSW | docker login --username $CRED_USR --password-stdin'
+            }
+        }
+        stage('build') {
+            steps {
+                sh '''\
+                    docker build \
+                        --build-arg BASE_IMAGE_TAG="$base_image_tag" \
```
abizjak, line 25: Doesn't this mean you'd have to update the `genesis-tools.Dockerfile` by parametrizing it with this image?

genesis-tools.Jenkinsfiles
```diff
@@ -0,0 +1,40 @@
...
+                        --label BASE_IMAGE_TAG="$base_image_tag" \
+                        --label GIT_COMMIT="$GIT_COMMIT" \
```
abizjak, line 27: The argument in the dockerfile is `BASE_REF` at the moment, and it should be a branch, not a commit.

bisgardo: We're passing both, one is the base image tag and the other the concordium-base ref.

We'll pass the base ref as a jenkins param also.

Thahara (approved):

PR #9, opened 2021-04-08
Run workflows when switching status from draft to ready for review.

abizjak:
Run workflows when marking a pull request as ready for review.

The default types are only `opened, synchronize, reopened`.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -16,6 +16,7 @@ name: Build and test
 on:
   pull_request:
     branches: main
+    types: [opened, synchronize, reopened, ready_for_review]
     paths:
     - '.github/workflows/build-test.yaml'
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test.yaml
```diff
@@ -26,6 +26,7 @@ on:
     - 'testdata/**'
   push:
     branches: main
+    types: [opened, synchronize, reopened, ready_for_review]
```
Thahara, line 29: This was probably meant to be under the pull request section above
```suggestion
```

Thahara (changes requested):

.github/workflows/build-test.yaml
abizjak: Yes, indeed. Fixed now

Thahara (approved):

PR #28, opened 2021-04-27
Tool for generating public keys for genesis.

eb-concordium:
## Purpose

To create a tool generating public keys used for genesis. 

## Changes

Such a tool has been added

## Checklist

      hard-to-understand areas.


rust-bins/src/bin/keygen-genesis.rs
```diff
@@ -0,0 +1,241 @@
...
+
+use pairing::bls12_381::Bls12;
+use sha2::Sha512;
+use std::path::PathBuf;
+use structopt::StructOpt;
+
+use curve25519_dalek::edwards::CompressedEdwardsY;
+use ed25519_dalek::*;
+use pairing::bls12_381::{G1, G2};
+use std::fs;
```
abizjak, line 17: Please merge all imports (i.e., now newlines between them) and then format to group them.

rust-bins/src/bin/keygen-genesis.rs
```diff
@@ -0,0 +1,241 @@
...
+#[derive(StructOpt)]
+struct KeygenIp {
+    #[structopt(long = "rand-input", help = "File with randomness.")]
```
abizjak, line 20: Maybe it would be better to call this something else. The argument is not really randomness. Maybe seed.

rust-bins/src/bin/keygen-genesis.rs
```diff
@@ -0,0 +1,241 @@
...
+    )]
+    bound: u32,
+    #[structopt(long = "out-pub", help = "File to output the public keys to.")]
+    out_pub: PathBuf,
+}
+
+#[derive(StructOpt)]
+struct KeygenAr {
+    #[structopt(long = "rand-input", help = "File with randomness.")]
+    rand_input: PathBuf,
```
abizjak, line 46: And likewise here.

rust-bins/src/bin/keygen-genesis.rs
```diff
@@ -0,0 +1,241 @@
...
+    #[structopt(
+        long = "ar-identity",
+        help = "The integer identifying the anonymity revoker"
+    )]
+    ar_identity: u32,
```
abizjak, line 51: Do we have a `FromStr` instance for `ArIdentity`? If so we should use ArIdentity here directly. IF not, disregard this comment, but we do need to make sure this is not 0.

rust-bins/src/bin/keygen-genesis.rs
```diff
@@ -0,0 +1,241 @@
...
+    description: String,
+    #[structopt(long = "out-pub", help = "File to output the public keys to.")]
+    out_pub: PathBuf,
+}
+
+#[derive(StructOpt)]
+#[structopt(
+    about = "Tool for generating keys for genesis",
+    author = "Concordium",
+    version = "0.2"
```
abizjak, line 66: Maybe just version 1.

rust-bins/src/bin/keygen-genesis.rs
```diff
@@ -0,0 +1,241 @@
...
+    url: String,
+    #[structopt(long = "description", help = "Description of anonymity revoker")]
+    description: String,
+    #[structopt(long = "out-pub", help = "File to output the public keys to.")]
+    out_pub: PathBuf,
+}
+
+#[derive(StructOpt)]
+#[structopt(
+    about = "Tool for generating keys for genesis",
```
abizjak, line 64: "Keys for identity providers and anonymity revokers for genesis"

rust-bins/src/bin/keygen-genesis.rs
```diff
@@ -0,0 +1,241 @@
...
+            Some(x) => x,
+            None => return Err($s.to_owned()),
+        }
+    };
+}
+
+fn handle_generate_ar_keys(kgar: KeygenAr) -> Result<(), String> {
+    let bytes_from_file = succeed_or_die!(fs::read(kgar.rand_input), e => "Could not read random input from provided file because {}");
+    if bytes_from_file.len() < 64 {
+        return Err("Provided randomness should be of size at least 64 bytes".to_string());
```
abizjak, line 114: I don't think we should have any such restrictions in this. In principle empty string is fine as a seed.

rust-bins/src/bin/keygen-genesis.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+    let mut bytes = bytes_from_file;
+    bytes.extend_from_slice(b"concordium_generator");
+    let generator = G1::hash_to_group(&bytes);
+    bytes.extend_from_slice(b"concordium_ar_public_key");
```
abizjak, line 119: I'm not sure it adds value to add these strings. Maybe we should just follow the principle you stated, namely that the first one is generated from the input, the second one is generated from the serialization of the first one, etc.

rust-bins/src/bin/keygen-genesis.rs
```diff
@@ -0,0 +1,241 @@
...
+    let ar_public_key = PublicKey {
+        generator,
+        key: G1::hash_to_group(&bytes),
+    };
+    let id = kgar.ar_identity;
+    let ar_identity = ArIdentity::try_from(id).unwrap();
```
abizjak, line 125: I think it'd be useful to handle errors here and inform the user that 0 is not acceptable.

rust-bins/src/bin/keygen-genesis.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+    Ok(())
+}
+
+fn handle_generate_ip_keys(kgip: KeygenIp) -> Result<(), String> {
+    let bytes_from_file = succeed_or_die!(fs::read(kgip.rand_input), e => "Could not read random input from provided file because {}");
+    if bytes_from_file.len() < 64 {
+        return Err("Provided randomness should be of size at least 64 bytes".to_string());
+    }
+    let ip_public_key = generate_ps_pk(kgip.bound, &bytes_from_file);
```
abizjak, line 156: Same comments as for the AR, I don't think we need to restrict data.

abizjak (changes requested):

rust-bins/docs/keygen-genesis.md
```diff
@@ -0,0 +1,85 @@
+# Key generation tool
+The key generation tool can generate encryption keys for anonymity revokers and signature keys for identity providers. The tool can be found in [../src/bin/keygen.rs](../src/bin/keygen.rs).
```
abizjak, line 2: I think this is a copy-paste error. It should refer to `keygen-genesis`.

abizjak (changes requested):

abizjak (approved):

PR #29, opened 2021-04-28
Print the entire genesis in the genesis tool.

abizjak:
## Purpose

Make the genesis tool output the entire genesis.

Example output: [genesis-block.txt](https://github.com/Concordium/concordium-base/files/6397777/genesis-block.txt)

## Changes

- Output the parts of genesis that was not yet printed.
- Try to pretty print as much as possible.
- Modify Show instances of UpdateKeysThreshold and AccountThreshold to simply print the number.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-base.cabal
+++ b/concordium-base.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.24
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 7eaa7403793b3e5ab92a451a158a922442f53c3399265692e643144111450e6c
+-- hash: bee12c1a1aa1edcfaed5a30df0cb5d539332b071c1d34ad598ee6edfcc35b64b
 
 name:           concordium-base
 version:        0.1.0.0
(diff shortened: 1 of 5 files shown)


td202:
The access structures for the baker stake threshold, adding anonymity revokers, and adding identity providers need to be added.

td202:
It would probably also be worth printing the genesis block hash. It might be nice to have a "summary" mode that doesn't print out all of the cryptographic values.

abizjak:
> It would probably also be worth printing the genesis block hash. It might be nice to have a "summary" mode that doesn't print out all of the cryptographic values.

I added printing of the hash.

For some use-cases it might be useful to have a summary mode, but for the genesis ceremony it is very important that those cryptographic values are also checked. If you don't have line-wrapping enabled they are pretty unobtrusive as well.

I don't think it makes sense to spend much time on this.

td202 (approved):
I have checked, and it appears to me that this indeed prints all of the genesis data.

tschudid (commented):
Looks good.

PR #31, opened 2021-05-02
Keygen randomness generation

chrmatt:
## Purpose

Added functionality for randomness generation to be used in anonymity revoker key generation. This combines randomness from the system with randomness input by the user via SHA256-based HKDF extraction.

## Changes

- Implemented functions for generating and verifying BIP39 phrases.
- Added a subcommand gen-rand for generating a randomness file that can be used as input to keygen-ip.
- Changed keygen-ar to use new randomness generation method.

## Checklist

      hard-to-understand areas.

## CLA acceptance


rust-bins/src/bin/keygen.rs
```diff
@@ -301,6 +472,228 @@ pub fn generate_ed_sk(
...
+/// Read vector of words from file.
+pub fn read_words_from_file(
+    path: PathBuf,
+    verify_bip39: bool,
+    bip39_map: &HashMap<&str, usize>,
+) -> Result<Vec<String>, String> {
+    let word_string = succeed_or_die!(
+        fs::read_to_string(path),
+        e => "Could not read input from provided file because {}"
+    );
```
abizjak, line 510: While it might not be too terrible in this case, I think it is bad practice to read entire  files into memory when you don't know their size.

Maybe it would be better to do something along the lines of
```rust
let file = succeed_or_die!(fs::File::open(path), ...);
let reader = BufReader::new(file);
let word_list: Vec<String> = reader.lines().flat_map(|x|  x.split_whitespace().map(str::to_owned)).collect();
```

rust-bins/src/bin/keygen.rs
```diff
@@ -301,6 +472,228 @@ pub fn generate_ed_sk(
...
+    bip39_map: &HashMap<&str, usize>,
+) -> Result<Vec<String>, String> {
+    // Ensure that input length is in allowed set if verification is enabled.
+    if verify_bip39 {
+        match num {
+            12 | 15 | 18 | 21 | 24 => (),
+            _ => {
+                return Err(
+                    "The input length for a valid BIP39 sentence must be in {12, 15, 18, 21, 24}."
+                        .to_string(),
```
abizjak, line 536: Perhaps print what length it was as well.

rust-bins/src/bin/keygen.rs
```diff
@@ -301,6 +472,228 @@ pub fn generate_ed_sk(
...
+        // read the ith word from stdin
+        let word = match verify_bip39 {
+            false => succeed_or_die!(
+                read_word(i),
+                e => "Could not read input from user because {}"
+            ),
+            true => succeed_or_die!(
+                read_bip39_word(i, &bip39_map),
+                e => "Could not read input from user because {}"
+            ),
```
abizjak, line 553: I recommend if statements for readability.

rust-bins/src/bin/keygen.rs
```diff
@@ -301,6 +472,228 @@ pub fn generate_ed_sk(
...
+    // check that word_vec contains allowed number of words
+    match word_vec.len() {
+        12 | 15 | 18 | 21 | 24 => (),
+        _ => return false,
+    };
+
+    // convert word vector to bits
+    let mut bit_vec = BitVec::<Msb0, u8>::new();
+    for word in word_vec {
+        match bip_word_map.get(&**word) {
```
abizjak, line 577: Why is this `&**word` reborrow needed?

rust-bins/src/bin/keygen.rs
```diff
@@ -301,6 +472,228 @@ pub fn generate_ed_sk(
...
+            Some(idx) => {
+                let word_bits = BitVec::<Msb0, u16>::from_slice(&[*idx as u16]).unwrap();
```
abizjak, line 579: Perhaps document why the unwrap is safe (or use .except with a meaningful string).

rust-bins/src/bin/keygen.rs
```diff
@@ -301,6 +472,228 @@ pub fn generate_ed_sk(
...
+                // ignore first 5 bits and add last 11 to bit_vec
+                bit_vec.extend_from_bitslice(&word_bits[5..]);
```
abizjak, line 581: What does this `add last 11 to bit_vec`  mean in relation to the next line?

rust-bins/src/bin/keygen.rs
abizjak, line 579: Also, why is this not using `::from_element`?

rust-bins/src/bin/keygen.rs
```diff
@@ -301,6 +472,228 @@ pub fn generate_ed_sk(
...
+            }
+            None => return false, // not valid if it contains invalid word
+        };
+    }
+
+    // Valid sentence consists of initial entropy of length ent_len plus
+    // checksum of length ent_len/32. Hence, ent_len * 33/32 = bit_vec.len().
+    let ent_len = 32 * bit_vec.len() / 33;
```
abizjak, line 589: I guess this will always be perfectly divisible since word_vec.len() is a multiple of 3 and we append 11 bits per word.

Perhaps this could be documented to avoid spending time understanding this every time.

rust-bins/src/bin/keygen.rs
```diff
@@ -301,6 +472,228 @@ pub fn generate_ed_sk(
...
+
+    // split bits after ent_len off. These correspond to the checksum.
+    let checksum = bit_vec.split_off(ent_len);
+
+    // checksum is supposed to be first cs_len bits of SHA256(entropy)
+    let mut sha = Sha256::new();
+    sha.update(bit_vec.into_vec());
+    let hash = sha.finalize();
```
abizjak, line 597: Thse three lines should be combined into `let hash = Sha256::digest(bit_vec.into_vec())`

rust-bins/src/bin/keygen.rs
```diff
@@ -301,6 +472,228 @@ pub fn generate_ed_sk(
...
+    };
+
+    // checksum length is ent_len / 32
+    let cs_len = ent_len / 32;
+
+    // checksum is first cs_len bits of SHA256(bytes)
+    // first compute hash of bytes
+    let mut sha = Sha256::new();
+    sha.update(bytes);
+    let hash = sha.finalize();
```
abizjak, line 630: Same as above, use the shorter `::digest`.

rust-bins/src/bin/keygen.rs
```diff
@@ -301,6 +472,228 @@ pub fn generate_ed_sk(
...
+    let mut system_randomness = [0u8; 256];
+    rand::thread_rng().fill(&mut system_randomness[..]);
+
+    // Combine both sources of randomness using HKDF extractor.
+    // Using random salt for added security.
+    let salt = b"keygen-rand-wEtIBpTIyzPRpZxNUIherQh14uPlDIdiqngFSo1qrqE1UrXl5DcUfV4xddYNDnOMIumlkqS9HNshATaFxAwqiUtLj5rxeBJIOsav";
+    let mut extract_ctx = HkdfExtract::<Sha256>::new(Some(salt));
+
+    // First add all words separated by " " in input_words to key material.
+    // Separation ensures word boundaries are persevered
```
abizjak, line 678:
```suggestion
    // Separation ensures word boundaries are preserved
```

rust-bins/src/bin/keygen.rs
```diff
@@ -301,6 +472,228 @@ pub fn generate_ed_sk(
...
+    bip_word_list: &[&str],
+) -> Result<Vec<String>, String> {
+    // Get randomness from system.
+    // Fill array with 256 random bytes, corresponding to 2048 bits.
+    let mut system_randomness = [0u8; 256];
+    rand::thread_rng().fill(&mut system_randomness[..]);
+
+    // Combine both sources of randomness using HKDF extractor.
+    // Using random salt for added security.
+    let salt = b"keygen-rand-wEtIBpTIyzPRpZxNUIherQh14uPlDIdiqngFSo1qrqE1UrXl5DcUfV4xddYNDnOMIumlkqS9HNshATaFxAwqiUtLj5rxeBJIOsav";
```
abizjak, line 674: Where does this salt come from? Please document.

rust-bins/src/bin/keygen.rs
```diff
@@ -319,4 +712,136 @@ mod tests {
             hex::decode("171cb88b1b3c1db25add599712e36245d75bc65a1a5c9e18d76f9f2b1eab4012")
         );
     }
+
+    /// Test correct generation of BIP39 sentences.
+    /// Values are taken from https://github.com/trezor/python-mnemonic/blob/master/vectors.json
+    #[test]
+    pub fn test_bip39_generation() {
+        let bip39_vec: Vec<_> = include_str!("data/BIP39English.txt")
```
abizjak, line 720: Use the constant I mentioned in preceding comments to avoid including the same file 3 times.

rust-bins/src/bin/keygen.rs
```diff
@@ -319,4 +712,136 @@ mod tests {
...
+            .join(" "),
+            "void come effort suffer camp survey warrior heavy shoot primary clutch crush open \
+             amazing screen patrol group space point ten exist slush involve unfold",
+        );
+    }
+
+    /// Test BIP39 verification.
+    #[test]
+    pub fn test_bip39_verification() {
+        let bip39_vec: Vec<_> = include_str!("data/BIP39English.txt")
```
abizjak, line 782: And also here.

rust-bins/src/bin/keygen.rs
```diff
@@ -319,4 +712,136 @@ mod tests {
...
+            "abandon".to_string(),
+            "abandon".to_string(),
+            "abandon".to_string(),
+            "abandon".to_string(),
+            "abandon".to_string(),
+            "abandon".to_string(),
+            "abandon".to_string(),
+        ];
+
+        assert_eq!(verify_bib39(&valid_list, &bip39_map), true);
```
abizjak, line 844: This is usually better expressed as just `assert!(verify_bib39(...))`.

rust-bins/src/bin/keygen.rs
```diff
@@ -319,4 +712,136 @@ mod tests {
...
+        assert_eq!(verify_bib39(&invalid_list, &bip39_map), false);
```
abizjak, line 845: And this one as `assert!(!verify_bib39(...))`

rust-bins/docs/keygen.md
```diff
@@ -20,11 +21,14 @@ Generate keys for an anonymity revoker. The following options are supported
...
 - `--global` a filename, containing cryptographic parameters of the chain anonymity revokers will be a part of.
+- `--in-len` an integer determining how many words to user is asked to enter.
 - `--name` a string, a simple identifier of the anonymity revoker. Chosen by themselves.
 - `--url` a URL with the website of the anonymity revoker
-- `--rand-input`, a filename with the randomness input. It must contain at least 64 bytes.
 - `--out`, a filename where the private keys will be emitted
 - `--out-pub`, a filename where the public data will be emitted. This is the data that must go to the chain.
+- `--recover-from-phrase` if set, recover keys from backup phrase. Otherwise, fresh keys are generated.
+- `--no-confirmation` if set, do not ask user to re-enter generated recovery phrase.
+- `--no-verification` if set, do not verify the validity of the input. Otherwise the input is verified to be a valid BIP39 sentence.
```
abizjak, line 31: Why is this flag necessary?

Is it just generality for its own sake or does it have utility? It complicates the logic because there are  now 2 versions of many things so I'd prefer to remove the not-verified option and just require that everything is as in the spec.

abizjak (changes requested):

rust-bins/src/bin/keygen.rs
```diff
@@ -138,10 +202,68 @@ fn output_possibly_encrypted<X: SerdeSerialize>(
...
+            );
+            let input_words =
+                read_words_from_terminal(kgar.in_len, !kgar.no_verification, &bip39_map)?;
+
+            // rerandomize input words using system randomness
+            let randomized_words = rerandomize_bip39(&input_words, &bip39_vec)?;
+
+            // print randomized words and ask user to re-enter
+            // clear screen
+            print!("\x1B[2J");
```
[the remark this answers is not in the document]
chrmatt: Yes, it's meant to clear everything to ensure that the user is not just copying the text from above. I found this online and also don't know about portability.

rust-bins/docs/keygen.md
chrmatt: I found it useful for testing, because you can just enter 24 times nothing with --no-verification. It also allows people to enter their own randomness if they don't have a ledger but something else. But it's probably not needed for AR use case with the ledger.

rust-bins/src/bin/keygen.rs
```diff
@@ -138,10 +202,68 @@ fn output_possibly_encrypted<X: SerdeSerialize>(
...
+
+    // Create hashmap mapping word to its index in the list.
+    // This allows to quickly test membership and convert words to their index
+    let mut bip39_map = HashMap::new();
+    for (i, word) in bip39_vec.iter().enumerate() {
+        bip39_map.insert(*word, i);
     }
+
+    let words_str = match kgar.recover {
+        true => {
```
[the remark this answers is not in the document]
chrmatt: Makes sense. Will change.

rust-bins/src/bin/keygen.rs
```diff
@@ -255,6 +377,55 @@ fn handle_generate_ip_keys(kgip: KeygenIp) -> Result<(), String> {
...
+    if bip39_vec.len() != 2048 {
+        return Err("The BIP39 word list must contain 2048 words.".to_string());
+    }
+
+    // Create hashmap mapping word to its index in the list.
+    // This allows to quickly test membership and convert words to their index
+    let mut bip39_map = HashMap::new();
+    for (i, word) in bip39_vec.iter().enumerate() {
+        bip39_map.insert(*word, i);
+    }
```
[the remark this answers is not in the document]
chrmatt: I tried to have it in a separate function but got didn't get it to work because of differences between &str and String I don't really understand. Please suggest something that works.

rust-bins/src/bin/keygen.rs
```diff
@@ -255,6 +377,55 @@ fn handle_generate_ip_keys(kgip: KeygenIp) -> Result<(), String> {
...
+
+    // get vector of input words from file or stdin
+    let input_words: Vec<_> = match grand.input_path {
+        // if input_path is provided, read file
+        Some(path) => read_words_from_file(path, !grand.no_verification, &bip39_map)?,
+        // if input_path is not provided, read words from stdin
+        None => {
+            println!(
+                "Please generate a seed phrase using a hardware wallet and input the words below."
```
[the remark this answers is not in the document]
chrmatt: Doesn't have to be, but you need a tool that generates a valid BIP39 phrase.

rust-bins/src/bin/keygen.rs
chrmatt: No idea, that was the first thing the compiled accepted. ;)

rust-bins/src/bin/keygen.rs
chrmatt: Doesn't your code also read the entire file? Maybe one should stop after reading 24 lines. But what if the first line is very long?

rust-bins/src/bin/keygen.rs
chrmatt: It should be using `::from_element`. I've changed it.

rust-bins/src/bin/keygen.rs
chrmatt: I've expanded the explanation in the comment.

rust-bins/src/bin/keygen.rs
```diff
@@ -138,10 +203,74 @@ fn output_possibly_encrypted<X: SerdeSerialize>(
...
-    let bytes_from_file = succeed_or_die!(fs::read(kgar.rand_input), e => "Could not read random input from provided file because {}");
-    if bytes_from_file.len() < 64 {
-        return Err("Provided randomness should be of size at least 64 bytes".to_string());
+    // Read word list and make sure it contains 2048 words.
+    let bip39_vec: Vec<_> = include_str!("data/BIP39English.txt")
+        .split_whitespace()
+        .collect();
+    if bip39_vec.len() != 2048 {
+        return Err("The BIP39 word list must contain 2048 words.".to_string());
+    }
```
tschudid, line 212: For a future version:
- Allow the user to select the word list or specify their own.
- Add a hash check to ensure that the correct BIP39English.txt is loaded.

rust-bins/src/bin/keygen.rs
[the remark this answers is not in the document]
tschudid: There is probably a crate for doing this (maybe https://docs.rs/crossterm/0.19.0/crossterm/terminal/struct.Clear.html).

rust-bins/src/bin/keygen.rs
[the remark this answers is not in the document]
tschudid: Ideally, the tool comes with a good entropy source. It would be a bad idea if someone uses their browser to generate the phrase.

rust-bins/src/bin/keygen.rs
chrmatt: If I understand correctly, the list is embedded into the binary. The hash would thus only prevent one from tampering with the source code.

tschudid (commented):
Looks reasonable.

rust-bins/docs/keygen.md
abizjak: Ok.

rust-bins/src/bin/keygen.rs
abizjak: I don't think we want to over-engineer this. It is essentially a one-off tool.

rust-bins/src/bin/keygen.rs
abizjak: It stores its content only once though and in this case we need to store all the words so it must read those.

It doesn't matter so much in this case.

rust-bins/src/bin/keygen.rs
```diff
@@ -301,6 +468,220 @@ pub fn generate_ed_sk(
...
+    bip_word_list: &[&str],
+) -> Result<Vec<String>, String> {
+    // Get randomness from system.
+    // Fill array with 256 random bytes, corresponding to 2048 bits.
+    let mut system_randomness = [0u8; 256];
+    rand::thread_rng().fill(&mut system_randomness[..]);
+
+    // Combine both sources of randomness using HKDF extractor.
+    // Using salt with randomness from random.org for added security.
+    let salt = b"keygen-rand-wEtIBpTIyzPRpZxNUIherQh14uPlDIdiqngFSo1qrqE1UrXl5DcUfV4xddYNDnOMIumlkqS9HNshATaFxAwqiUtLj5rxeBJIOsav";
```
abizjak, line 662: I think this is a bit dodgy since this is unverifiable that you have done so.

Would it not make more sense to have something plain and boring here since the value is public, e.g., 

`concordium-key-generation-tool-version-1`

abizjak (approved):

tschudid (commented):
👍

PR #32, opened 2021-05-03
Make the AR keygen tool interactive.

abizjak:
The same interface that used to exist is still supported, however all arguments
to the tool are now optional. If they are not supplied the relevant values are
queried interactively.

## Purpose

Make the AR keygen tool easier to use.

## Changes

Query for input interactively if it is not supplied on the command-line.

```
Enter the path to the cryptographic parameters file: foo/global.json
Enter AR identity: 1
Enter the name of the AR: Foobar
Enter URL of the AR: foobar@baz
Enter description of the AR: blah
Output file name: ar-data-1.json
Output file for public data: ar-info-1.pub.json
Enter password to encrypt credentials (leave empty for no encryption): 
No password supplied, so output will not be encrypted.
Wrote private keys to ar-data-1.json.
Wrote public keys to ar-info-1.pub.json.
```

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-bins/src/bin/keygen.rs
+++ b/rust-bins/src/bin/keygen.rs
@@ -18,7 +18,6 @@ use rand::Rng;
 use sha2::{Digest, Sha256, Sha512};
 use std::{
     collections::HashMap,
-    convert::TryFrom,
     fs::{self, File},
     io::Write,
     path::PathBuf,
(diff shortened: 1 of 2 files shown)


rust-bins/src/bin/keygen.rs
```diff
@@ -143,7 +138,7 @@ struct GenRand {
 #[structopt(
     about = "Tool for generating keys",
     author = "Concordium",
-    version = "0.2"
+    version = "1"
```
chrmatt, line 141: When I do, e.g., `./keygen keygen-ar --version`, it returns `keygen-keygen-ar 0.1.0` . Is this intended?

chrmatt (commented):
Looks it's doing what it's supposed to do, apart from the inconsistent versioning.

rust-bins/src/bin/keygen.rs
abizjak: Fixed.

chrmatt (approved):
Looks good.

PR #33, opened 2021-05-04
Retry word entry on failure and make it possible to run just with system randomness.

abizjak:
## Purpose

Retry word entry on failure to make it easier to use the tool.

## Changes

- Retry word entry on failure.
- Add an option to just use the system randomness. It really is good enough for all practical purposes so there is no harm in supporting this.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-bins/src/bin/keygen.rs
+++ b/rust-bins/src/bin/keygen.rs
@@ -104,6 +104,12 @@ struct KeygenAr {
         help = "Do not ask user to re-enter generated recovery phrase."
     )]
     no_confirmation: bool,
+    #[structopt(
+        long = "only-system-randomness",
+        help = "Do not ask for a list of words from the user. Generate keys only using the system \
+                randomness."
+    )]
+    only_system_randomness: bool,
 }
 
 #[derive(StructOpt)]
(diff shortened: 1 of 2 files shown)


eb-concordium (approved):

rust-bins/src/bin/keygen.rs
```diff
@@ -259,15 +269,23 @@ fn handle_generate_ar_keys(kgar: KeygenAr) -> Result<(), String> {
...
-            // clear screen
-            execute!(std::io::stdout(), Clear(ClearType::All))
-                .map_err(|_| "Could not clear screen.".to_owned())?;
-            println!("Please enter recovery phrase again to confirm.");
-
-            let confirmation_words = read_words_from_terminal(kgar.in_len, true, &bip39_map)?;
-
-            if confirmation_words != randomized_words {
-                return Err("Recovery phrases do not match.".to_string());
+            let mut first = true;
```
chrmatt, line 272: Variable could be avoided by printing initial prompt before loop and only print retry request if check fails at the end.

chrmatt (commented):
Looks good.

Issue #46, opened 2021-06-11
Move relevant blocksummary types from concordium-node to concordium-base


abhconcordium:
**Task description**
Due to historical reasons, there are types  relating to the block summary defined in concordium-node which would make more sense in concordium-base.

Move the query result types to concordium-base, and move all necessary type definitions for this from node to base.

**Sub-tasks**
- [x] Identify which types should be moved
- [x] Add relevant files and definitions to base, on branch  `bstypes-to-base` in concordium-base
- [x] clean up files moved to base
- [x] delete definitions from node and refer to base instead
- [ ] delete definitions from client and refer to base instead

PR #47, opened 2021-06-11
Move query types to concordium-base

abhconcordium:
## Purpose

This moves various types that are used in JSON GRPC queries to `concordium-base` so that they can be shared with other tooling more conveniently.

Partially addresses #46 together with https://github.com/Concordium/concordium-node/pull/94

## Changes

Added query response types and updates type, as well as minor consistency changes for some primitives

## Checklist

      hard-to-understand areas.


abhconcordium:
Now works with the `refactor-types-to-base` branch on concordium-node.
Still need to confirm that it works well with other repo's which are dependent on Base.

@td202 
Thoughts on this selection of types to move into base, and this structure?

haskell-src/Concordium/Types/Queries.hs
```diff
@@ -0,0 +1,278 @@
...
+        object
+            [ "status" .= String "finalized",
+              "outcomes" .= Map.singleton bh outcome
+            ]
+
+-- |The status of a transaction with respect to a specified block
+data BlockTransactionStatus
+    = -- |Either the transaction is not in that block, or that block is not live
+      BTSNotInBlock
+    | -- |The transaction was received but not known to be in that block
```
abizjak, line 247: The documentation is unclear which of the two statuses above would make sense if a transaction is received, but not in a given block.

abizjak (commented):
The main risk with this PR (and the related one in the node) is that there are typos or minor issues that will break compatibility.

Thus there needs to be comprehensive end-to-end testing of the affected queries with the existing concordium-client before this is merged.

haskell-src/Concordium/Types/Queries.hs
td202: Both are possible. If the transaction is received but not in any block, then the status will be `BTSReceived`. If the transaction is received but in a different block, then the status will be `BTSNotInBlock`. Note that at the RPC level, `BTSNotInBlock` is indistinguishable from the case where the transaction is completely absent.

abizjak: I'm aware of the possibilities, my complaint was that the documentation of the options is ambiguous and should be clarified.

haskell-src/Concordium/Types/Queries.hs
```diff
@@ -0,0 +1,290 @@
...
+      -- |The total GTU in encrypted balances
+      rsTotalEncryptedAmount :: !Amount,
+      -- |The amount in the baking reward account
+      rsBakingRewardAccount :: !Amount,
+      -- |The amount in the finalization reward account
+      rsFinalizationRewardAccount :: !Amount,
+      -- |The amount in the GAS account
+      rsGasAccount :: !Amount
+    }
+    deriving (Show)
```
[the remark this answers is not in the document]
abhconcordium: Good catch, this was one of the changes which was from the protocol-update branch.

I will check if this is actually a breaking change anywhere, and document them properly, and go through the other queries to make sure nothing similar happens in any of the other queries moved from `Getters.`

haskell-src/Concordium/Types/Queries.hs
td202: I don't think I understand what the intended semantics of `getTransactionStatusInBlock` is, then. Perhaps you could document this? Do we rely somewhere on the distinction between these cases?

haskell-src/Concordium/Types/Queries.hs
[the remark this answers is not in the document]
td202: This is a breaking change for the tokenomics reporter.

td202 (approved):

abizjak (approved):

PR #49, opened 2021-06-17
Database Versioning

td202:
These changes are extracted from #1.

## Purpose

Changes to support database versioning for the tree state database.

## Changes

* Add `Eq` and `Show` instances for `ProtocolVersion`
* Add promotion and demotion functions between `ProtocolVersion` and `SProtocolVersion`. These allow conversions between term-level and type-level representations of protocol versions.

## Checklist

      hard-to-understand areas.


concordium-base.cabal
```diff
@@ -473,7 +537,15 @@ benchmark verify-credential-perf
...
-  default-extensions: FlexibleContexts FlexibleInstances FunctionalDependencies GeneralizedNewtypeDeriving LambdaCase MultiParamTypeClasses RecordWildCards TupleSections
+  default-extensions:
+      FlexibleContexts
+      FlexibleInstances
+      FunctionalDependencies
+      GeneralizedNewtypeDeriving
+      LambdaCase
+      MultiParamTypeClasses
+      RecordWildCards
+      TupleSections
```
abizjak, line 548: This reformatting is because you upgraded stack and hpack and this will cause problems for people who have not since stack refuses to regenerate the cabal file, but it is not very clear about it.

Tool upgrades should be planned and, importantly, announced to everybody so that we don't have these failures and noise.

We should also update the static libraries build, and any other place where we pin the stack version, and test that the build still works correctly with the updated version.  This is not a guarantee, it has often failed in the past.

abizjak (approved):

PR #50, opened 2021-06-23
Store commitment randomness

eb-concordium:
## Purpose

To output randomness used in commitments inside credentials so that they can be opened later on.

## Changes

 - The `create_credential` also outputs the randomness from the commitments used 
  in the proofs for credential deployment. The client tool will also output the randomness when
  using it to create credentials. 
- The `create_credential_aux` in the mobile wallet library now outputs the randomness so that it can be stored for later use.

## Checklist

      hard-to-understand areas.


rust-src/id/src/account_holder.rs
```diff
@@ -943,11 +951,18 @@ fn compute_pok_sig<
     Ok((prover, secret))
 }
 
+#[derive(SerdeSerialize, SerdeDeserialize)]
```
abizjak, line 954: I think seeing that this type is now in the public API, would it make sense to move it to `types.rs`?

CHANGELOG.md
```diff
@@ -12,7 +12,8 @@
   changed format, with the addition of the `messageExpiry` field.
 
 ## Mobile wallet
-
+- the randomness output by `create_credential` is now also output by `create_credential_aux`
+  so that the randomness can be stored.
```
abizjak, line 16: Can you please name the exact field that is added to the response of this call in the wallet library.

It would also be good to update the example files in `rust-bins/wallet-notes` with the new response format.

abizjak (approved):

PR #58, opened 2021-07-16
Id library

eb-concordium:
## Purpose

To introduce core functionality for proving and verifying properties about an identity behind an account, such as:
 * reveal an attribute
 * proving ownership of an account
 * proving that an attribute is in a range.
 
More functionality will be added to support more use cases, when these are known.

## Changes

Functionality for proving/verifying the above mentioned properties has been added.

## Checklist

      hard-to-understand areas.


rust-src/id/src/id_prover.rs
```diff
@@ -0,0 +1,91 @@
...
+    let b = upper.to_field_element();
+    let mut scalar1 = delta;
+    let two = C::scalar_from_u64(2);
+    let two_n = two.pow(&[64]);
+    scalar1.add_assign(&two_n);
+    scalar1.sub_assign(&b);
+    let mut scalar2 = delta;
+    scalar2.sub_assign(&a);
+    let rand1 = PedersenRandomness::new(**r);
+    let rand2 = PedersenRandomness::new(**r);
```
abizjak, line 80: Use `r.clone()` to avoid copying secret data. So
```
let rand1 = r.clone();
let rand2 = r.clone();
```

rust-src/id/src/id_verifier.rs
```diff
@@ -0,0 +1,173 @@
...
+    gens: &Generators<C>,
+    lower: &AttributeType,
+    upper: &AttributeType,
+    c: &Commitment<C>,
+    proof: &RangeProof<C>,
+) -> bool {
+    let mut transcript = RandomOracle::domain("attribute_range_proof");
+    let a = lower.to_field_element();
+    let b = upper.to_field_element();
+    let zero_randomness = PedersenRandomness::<C>::new(C::Scalar::zero());
```
abizjak, line 51: Use `PedersenRandoness::zero()`

rust-src/id/src/id_verifier.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    verify_efficient(
+        &mut transcript,
+        64,
+        &[com_delta_minus_b_plus_2n, com_delta_minus_a],
+        proof,
+        gens,
+        keys,
+    )
+    .is_ok()
```
abizjak, line 68: Maybe return the Result instead. I don't think there is harm in that, and it might be useful in some cases to know why the proof failed (even though it is not very informative at the moment).

rust-src/id/src/id_verifier.rs
```diff
@@ -0,0 +1,173 @@
...
+        let attribute = AttributeKind("some attribute value".to_string());
+        let value = Value::<G1>::new(attribute.to_field_element());
+        let (commitment, randomness) = keys.commit(&value, &mut csprng);
+
+        assert!(verify_attribute(
+            &keys,
+            &attribute,
+            &randomness,
+            &commitment
+        ));
```
abizjak, line 150: Recommend adding a short string description to the assertion so that if the test fails it is easier to find and understand what the intention was, e.g.,

rust-src/id/src/id_verifier.rs
```diff
@@ -0,0 +1,173 @@
...
+        let value = Value::<G1>::new(attribute.to_field_element());
+        let (commitment, randomness) = keys.commit(&value, &mut csprng);
+        let maybe_proof =
+            prove_attribute_in_range(&gens, &keys, &attribute, &lower, &upper, &randomness);
+        let result = if let Some(proof) = maybe_proof {
+            verify_attribute_range(&keys, &gens, &lower, &upper, &commitment, &proof)
+        } else {
+            false
+        };
+        assert!(result);
```
abizjak, line 171: Recommend adding a description of what the assertion is doing so that if it fails it is possible to find what went wrong.

rust-src/id/src/id_prover.rs
```diff
@@ -0,0 +1,91 @@
...
+    let delta = attribute.to_field_element();
+    let a = lower.to_field_element();
+    let b = upper.to_field_element();
+    let mut scalar1 = delta;
+    let two = C::scalar_from_u64(2);
+    let two_n = two.pow(&[64]);
+    scalar1.add_assign(&two_n);
+    scalar1.sub_assign(&b);
+    let mut scalar2 = delta;
+    scalar2.sub_assign(&a);
```
abizjak, line 78: I think it would be useful for future reference to document what the above 10 or so lines are doing, and why this proves the `>= a && < b` at the same time.

rust-src/id/src/id_prover.rs
```diff
@@ -0,0 +1,91 @@
...
+/// coincide with the hash of a transaction (assuming that SHA256 is
+/// collision-resistant).
+pub fn prove_ownership_of_account(
+    data: CredentialData,
+    account: AccountAddress,
+    challenge: &[u8],
+) -> AccountOwnershipProof {
+    let mut hasher = Sha256::new();
+    hasher.update(account.0);
+    hasher.update([0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8, 0u8]);
```
abizjak, line 34:
```suggestion
    hasher.update([0u8; 8]);
```

abizjak (commented):
Looks really good, I especially like the range inclusion proof.

I only have some relatively minor comments.

abizjak (approved):

PR #61, opened 2021-07-22
Documentation

abizjak:
## Purpose

Add documentation to exported definitions from the id and common libraries.

## Changes

- add documentation
- move AttributeKind from ffi to constants where it more properly belongs. The reason it was in ffi is a historical accident in that that was the first place it was needed, but now it is used extensively by all instantiations of the id library, so it should be in constants together with other fixed parameters.

## Checklist

      hard-to-understand areas.


abizjak:
@shjortConcordium would you mind going through this PR and checking whether these changes break anything on your end (that is not easily fixable)? I don't think they should, but it is better to be safe. :)

shjortConcordium:
> @shjortConcordium would you mind going through this PR and checking whether these changes break anything on your end (that is not easily fixable)? I don't think they should, but it is better to be safe. :)

I have checked, and it does not break anything (that isn't fixed by adjusting the imports).

abizjak:
> > @shjortConcordium would you mind going through this PR and checking whether these changes break anything on your end (that is not easily fixable)? I don't think they should, but it is better to be safe. :)
> 
> I have checked, and it does not break anything (that isn't fixed by adjusting the imports).

Can you do a proper review then and approve the changes, or insert comments if you have them?

shjortConcordium (commented):
Looked it over, found some small errors.

abizjak:
> Looked it over, found some small errors.

Thanks. I applied your suggestions.

shjortConcordium (approved):

PR #62, opened 2021-07-26
Fail quickly when the initial account transaction goes missing.

abizjak:
## Purpose

Make the identity provider service fail early when the account creation transaction becomes absent. There are some legitimate cases where the transaction could become absent due to configuration errors, e.g., the node the wallet proxy talks to being temporarily out of date. In the past that was really the predominant reason for the transaction being absent, but with the desktop wallet's deterministic key generation it is now quite possible for somebody to get into this state by mistake.

In such a case it is better for a test service like this to fail quickly.

## Changes

Fail early in case initial account creation transaction goes absent.

## Checklist

      hard-to-understand areas.


abizjak:
This will be obsolete with initial account removal.

PR #65, opened 2021-08-17
Use multiexp in commitment_to_share

### LINKED ISSUE #60: Computing commitment to share would benefit from multiexponentiation
**Task description**

The function `id::utils::commitment_to_share` would benefit from multiexponentiation.
This would increase performance in a function that is used a fair amount.

**Sub-tasks**
- [x] Benchmark identity and credential creation before the change
- [x] Write the above mentioned function using multiexp.
- [x] Benchmark identity and credential creation after the change

SimonKamp:
## Purpose

Speed up committing to a share by using multi exponentiation. Close #60 

## Changes

Added benchmark for the old method using Horner's. 
Using multexp instead of Horner's cuts 52% of the execution time.

## Checklist

      hard-to-understand areas.


rust-src/id/src/utils.rs
```diff
@@ -26,15 +26,14 @@ pub fn commitment_to_share<C: Curve>(
...
     coeff_commitments: &[Commitment<C>],
 ) -> Commitment<C> {
-    // TODO: This would benefit from multiexponentiation.
-    let mut cmm_share_point: C = C::zero_point();
-    // Horner's scheme in the exponent
-    for cmm in coeff_commitments.iter().rev() {
-        // FIXME: This would benefit from multiexponentiation.
-        cmm_share_point = cmm_share_point.mul_by_scalar(&share_number);
-        cmm_share_point = cmm_share_point.plus_point(cmm);
+    let n = coeff_commitments.len();
```
abizjak, line 29: I would recommend checking if n == 0 and returning early since `multiexp` will do needless work in that case.

abizjak (approved):

PR #71, opened 2021-09-03
Multiversion query changes

td202:
## Purpose

Introduce absolute block height for queries.

## Changes

* Define `AbsoluteBlockHeight`.
* Change ATI to use `AbsoluteBlockHeight` instead of `BlockHeight`
* Change `ConsensusSummary` to use `AbsoluteBlockHeight`
* Change `BlockInfo` to use `AbsoluteBlockHeight`, while also adding the genesis index and local block height.

## Checklist

      hard-to-understand areas.


haskell-src/Concordium/Types/Queries.hs
```diff
@@ -122,12 +123,18 @@ $(deriveJSON defaultOptions{fieldLabelModifier = firstLower . dropWhile isLower}
...
       -- |The last finalized block when this block was baked
       biBlockLastFinalized :: !BlockHash,
       -- |The height of this block
-      biBlockHeight :: !BlockHeight,
+      biBlockHeight :: !AbsoluteBlockHeight,
+      -- |The genesis index for this block. This counts the number of protocol updates that have
+      -- preceded this block.
+      biGenesisIndex :: !GenesisIndex,
+      -- |The height of this block relative to the (re)genesis block at its genesis index.
+      biLocalBlockHeight :: !BlockHeight,
```
abizjak, line 137: The naming in consensus status is using "current era". I think the same naming should be used in both places, otherwise we have to confuse the users and explain the differences.

haskell-src/Concordium/Types/Queries.hs
```diff
@@ -37,9 +38,9 @@ data ConsensusStatus = ConsensusStatus
       -- |Hash of the last finalized block
       csLastFinalizedBlock :: !BlockHash,
       -- |Height of the best block (since latest regenesis)
```
eb-concordium, line 40:
```suggestion
      -- |Absolute height of the best block
```

haskell-src/Concordium/Types/Queries.hs
```diff
@@ -37,9 +38,9 @@ data ConsensusStatus = ConsensusStatus
...
-      csBestBlockHeight :: !BlockHeight,
+      csBestBlockHeight :: !AbsoluteBlockHeight,
       -- |Height of the last finalized block (since latest regenesis)
```
eb-concordium, line 42:
```suggestion
      -- |Absolute height of the last finalized block
```

haskell-src/Concordium/Types/Block.hs
```diff
@@ -24,6 +24,7 @@ type EpochLength = Slot
 -- |Index of an epoch.
 type Epoch = Word64
 
+-- |Block height relative to the genesis block on the block's chain.
```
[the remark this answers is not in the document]
td202: Done.

haskell-src/Concordium/Types/Queries.hs
td202: I've renamed it to `biEraBlockHeight`. I don't think it should be `biCurrentEraBlockHeight`, since it is relative to the block's own era, not the most recent one.

haskell-src/Concordium/Types/Queries.hs
```diff
@@ -36,10 +37,10 @@ data ConsensusStatus = ConsensusStatus
...
       -- |Hash of the last finalized block
       csLastFinalizedBlock :: !BlockHash,
-      -- |Height of the best block (since latest regenesis)
-      csBestBlockHeight :: !BlockHeight,
-      -- |Height of the last finalized block (since latest regenesis)
-      csLastFinalizedBlockHeight :: !BlockHeight,
+      -- |Absolute height of the best block
+      csBestBlockHeight :: !AbsoluteBlockHeight,
+      -- |Absolute height of the last finalized block
+      csLastFinalizedBlockHeight :: !AbsoluteBlockHeight,
```
abizjak, line 43: I should have mentioned this in the previous round, but it would be nice to include the relative height in consensus status as well, for symmetry with block info.

abizjak (approved):

eb-concordium (approved):

haskell-src/Concordium/Types/Queries.hs
td202: Alternatively, we could include the absolute block height of the genesis block of the current era, and the relative heights can be derived. I will leave it out for now as I don't think it's particularly valuable, though we can add it in future if there's a compelling use case.

PR #72, opened 2021-09-07
Multiexp

SimonKamp:
## Purpose

Using multi exponentiation for better performance.

## Changes

`compute_message` used when signing the pre-identity-object by an identity provider and `commit_point` for aggregate DLog were benchmarked, changed to exploit `multiexp` and found to be 66-70% faster.

`sign_identity_object` (which depends on `compute_message`) is 12% faster, but only constitutes about 1% of `verify_credentials`.

## Checklist

      hard-to-understand areas.


rust-src/id/src/identity_provider.rs
```diff
@@ -443,30 +441,41 @@ fn compute_message<P: Pairing, AttributeType: Attribute<P::ScalarField>>(
         return Err(Reason::TooManyAttributes);
     }
 
+    let mut gs = Vec::new();
+    let mut exps = Vec::new();
```
abizjak, line 445: Performance wise it really won't matter, but I think we should use `with_capacity` because there is no reason not to.

rust-src/id/benches/aggr_dlog_commit_point.rs
```diff
@@ -0,0 +1,25 @@
+use criterion::*;
+use curve_arithmetic::*;
+use id::sigma_protocols::{aggregate_dlog::*, common::*};
+use pairing::bls12_381::G1;
+use rand::*;
+
+fn bench_aggr_dlog_commit_point(c: &mut Criterion) {
```
abizjak, line 7: Please provide a brief comment as to the purpose of this benchmark.

rust-src/id/benches/compute_message.rs
```diff
@@ -0,0 +1,71 @@
...
+        "nationality": "DK",
+        "sex": "1"
+    }"#;
+    type ExampleAttributeList =
+        AttributeList<id::constants::BaseField, id::constants::AttributeKind>;
+    let alist: BTreeMap<AttributeTag, id::constants::AttributeKind> = from_str(aliststr).unwrap();
+
+    // the number of scalars in the pk should be equal to the total number of
+    // attributes, including mandatory attributes: i.e. `valid_to` etc.
+    let public_key: ps_sig::PublicKey<Bls12> =
```
abizjak, line 54: Technically, what is required is that it is at least as much.

abizjak (commented):
It would be useful to know whether this has a meaningful effect on the performance of verify + sign identity object.

rust-src/id/benches/aggr_dlog_commit_point.rs
SimonKamp: Added minimal description. I guess there is no real purpose, as we found out it isn't used. If we need it in the future we now know that it is faster.

SimonKamp:
> It would be useful to know whether this has a meaningful effect on the performance of verify + sign identity object.

To be clear: this means benchmark `verify_credentials`?

abizjak:
> > It would be useful to know whether this has a meaningful effect on the performance of verify + sign identity object.
> 
> To be clear: this means benchmark `verify_credentials`?

Yes. And possibly `sign_identity_object` as well.

SimonKamp:
> > > It would be useful to know whether this has a meaningful effect on the performance of verify + sign identity object.
> > 
> > 
> > To be clear: this means benchmark `verify_credentials`?
> 
> Yes. And possibly `sign_identity_object` as well.

Sorry, should have provided more context: 
`verify_credentials` is basically just a call to `validate_request` and then `sign_identity_object`.
Is it useful to benchmark them in isolation, or do we just want effect on the combined thing.

abizjak:
> > > > It would be useful to know whether this has a meaningful effect on the performance of verify + sign identity object.
> > > 
> > > 
> > > To be clear: this means benchmark `verify_credentials`?
> > 
> > 
> > Yes. And possibly `sign_identity_object` as well.
> 
> Sorry, should have provided more context:
> `verify_credentials` is basically just a call to `validate_request` and then `sign_identity_object`.
> Is it useful to benchmark them in isolation, or do we just want the effect on the combined thing.

It is useful to benchmark in isolation because in use they are used separately, so first `validate_request` is used, then "real world verification (e.g., taking photos)" is done, then `sign_identity_object` is called to return the identity object.

rust-src/id/Cargo.toml
```diff
@@ -100,4 +101,5 @@ harness = false
 [features]
 default = ["ffi"]
 ffi = []
-wasm-browser-test = ["ffi"]
+# allow use of code annotated with `#[cfg(any(test, feature = "test-helpers"))]` in benchmarks using `required-features = ["test-helpers"]` :
```
abizjak, line 100:
```suggestion
# This feature is used to expose some internal testing code so it can be used in integration tests and benchmarks.
```

abizjak (approved):
Looks good to me, let's just wait for Emil's review as well.

.github/workflows/build-test.yaml
```diff
@@ -159,3 +159,4 @@ jobs:
       run: |
         cargo test --manifest-path rust-src/Cargo.toml --all --verbose --release
         cargo test --manifest-path identity-provider-service/Cargo.toml --features=vendored-ssl --all --verbose --release
+        cargo bench --features test-helpers --no-run
```
abizjak, line 162:
```suggestion
        cargo bench --manifest-path rust-src/Cargo.toml --features id/test-helpers --no-run
```

eb-concordium (approved):

PR #73, opened 2021-09-07
Add checks for RO challenge prefix collision in soundness tests

SimonKamp:
## Purpose

Some tests of soundness of various protocols did not test for collision of RO challenge prefixes. These would happen with probability > 1/10^6, and thus make the tests fail in some rare cases.

## Changes

Add tests checking if two randomly selected ROs are identical in case of unexpected verifications, when one is used for generating a proof and the other for verification.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/random_oracle/src/lib.rs
+++ b/rust-src/random_oracle/src/lib.rs
@@ -57,10 +57,10 @@ impl Buffer for RandomOracle {
     fn result(self) -> Self::Result { self.0.finalize() }
 }
 
+impl Eq for RandomOracle {}
+
 impl PartialEq for RandomOracle {
-    fn eq(&self, other: &Self) -> bool {
-        self.split().get_challenge() == other.split().get_challenge()
-    }
+    fn eq(&self, other: &Self) -> bool { self.0.clone().finalize() == other.0.clone().finalize() }
 }
 
 impl RandomOracle {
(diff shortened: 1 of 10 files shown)


rust-src/random_oracle/src/lib.rs
```diff
@@ -56,6 +57,12 @@ impl Buffer for RandomOracle {
     fn result(self) -> Self::Result { self.0.finalize() }
 }
 
+impl PartialEq for RandomOracle {
+    fn eq(&self, other: &Self) -> bool {
+        self.split().get_challenge() == other.split().get_challenge()
+    }
+}
```
abizjak, line 64: I am not a fan of this instance, but I'm not sure we can easily do anything better. Presumably there is no way to check internal state directly?

Additionally, please derive `Eq` a well since this relation is total.

abizjak (commented):
Apart from also adding Eq this looks fine.

rust-src/random_oracle/src/lib.rs
SimonKamp: We could add #[cfg(test)] here or use test().get_challenge everywhere.

SimonKamp: Implemented Eq, Sha3_265 does not export anything useful for implementing it without using finalize()

abizjak (approved):

abizjak:
@SimonKamp Looks good, please merge.

Issue #86, opened 2021-10-12
Provide a C# wrapper for identity issuance libraries.


abizjak:
**Task description**

In order to ease integration we are asked to provide an integration to .NET specifically C#.

We need to first specify how much of the objects should be modeled explicitly and how much should just be JSON blobs.
Some of it will need to be modeled since the identity provider will need to do validation. Ideally all of the objects should be modeled, apart from keys which should probably just stay strings/byte arrays.

At least the following need to be modeled to one degree or another.
- identity provider public and private keys
- identity object request ("PreIdentityObject")
- anonymity revocation record
- initial account credential and initial account creation message
- identity object that is returned to the user.

The interface in C# should be as simple to use as possible.

PR #89, opened 2021-11-02
Export validation and identity creation functions to be used from C#, and create C# Library that uses these.

eb-concordium:
## Purpose

* To export the functions `create_identity_object` and `validate_request` as c functions that can then be called from C#
* To make a C# library that uses the above exported functions. Closes https://github.com/Concordium/concordium-base/issues/86. 

## Changes

* The `idiss` in `concordium-base` has been extended to also export functions `create_identity_object_cs` and  `validate_request_cs` as C functions (to be used from  C#)
* Three C# projects (`IdissLib`, `IdissExample` and `IdissLibTest`) in `idiss-csharp` have been added in `concordium-base`.  The `IdissLib` project is a library that invokes the exported C functions.  The `IdissExample` project is an example that uses the `IdissLib` functions on some input and prints the output. The `IdissLibTest` project tests `IdissLib`.

## Checklist

      hard-to-understand areas.


abizjak (commented):
Please also document the changed response format for the nodejs integration.

We also need to remember to update the typescript integration Notabene is using.

idiss-csharp/IdissLib/Class1.cs
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    public class IpPrivateKeys {
+        public string ip_private_key{get;set;}
+        public string ip_cdi_private_key{get;set;}
+    }
+
+
+
+    public class DictionaryConverter : JsonConverter<Dictionary<AttributeTag, Attribute>>
```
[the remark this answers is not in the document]
eb-concordium: Yes. It is used in tests and in the example.

idiss-csharp/IdissLib/Class1.cs
```diff
@@ -0,0 +1,333 @@
+﻿using System;
```
orhoj, line 1: The class file should have a less generic name than `Class1.cs`.

idiss-csharp/IdissLib/Class1.cs
```diff
@@ -0,0 +1,333 @@
...
+using System.Text.Json;
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+namespace IdissLib
+{
+
+    /// Here we provide classes that match some of the Rust types that involved with identity validation and creation. 
+    /// Most of the field names of these classes are chosen so that they match JSON fields of JSON serialization of the corresponding Rust types.
+    public class YearMonth {
```
orhoj, line 13: Use consistent placement of `{ }`. In C# the usual convention is to always put a newline before, so it would be:
```
public Class YearMonth
{
...
}
```
In this class I see a mix of the styles. I haven't used C# for quite some time, but perhaps there is some standardized linter you could use (VS code probably does it for you too).

idiss-csharp/IdissLib/Class1.cs
```diff
@@ -0,0 +1,333 @@
...
+        public string proofsOfKnowledge{get;set;}
+    }
+
+    public class IdentityObject {
+        public PreIdentityObject preIdentityObject{get;set;}
+        public AttributeList attributeList{get;set;}
+        public string signature{get;set;}
+    }
+
+    public class VersionedPreIdentityObject {
```
orhoj, line 89: As you have multiple `Versioned` classes, it should be possible to simplify it using generics. You should be able to do something like:
```
class Versioned<T> 
{
	public UInt32 v { get; set; }
	public T value { get; set; }
}
```
and then use it like: `Versioned<PreIdentityObject>`.

idiss-csharp/IdissLib/Class1.cs
orhoj: Also, I think it would be an improvement to split the contents of this class into 3 files:
1. All the type definitions.
2. The JsonConverters
3. The class `public static class Idiss`

idiss-csharp/IdissLib/Class1.cs
```diff
@@ -0,0 +1,534 @@
...
+
+        /// A wrapper around the imported C function "validate_request_cs". The arguments are
+        /// - global - the versioned global context
+        /// - ip_info - the versioned identity provider info
+        /// - ars_info - the versioned map of ArInfos
+        /// - request - the identity object request
+        /// The function either
+        /// - returns the address of the initial account, if the request is valid, or
+        /// - throws an exception, if the request is invalid or the input is malformed. 
+        public static AccountAddress validate_request(VersionedGlobalContext global, VersionedIpInfo ip_info, VersionedArInfos ars_infos, IdObjectRequest request)
```
orhoj, line 462: Use PascalCase for C# methods.

idiss-csharp/IdissLibTest/UnitTest1.cs
```diff
@@ -0,0 +1,59 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.IO;
+using System.Text.Json;
+using IdissLib;
+
+namespace IdissLibTest
+{
+    [TestClass]
+    public class UnitTest1
```
orhoj, line 10: UnitTest1 sounds autogenerated by an IDE and is not very descriptive. Preferably it should be updated (together with the filename).

idiss-csharp/IdissExample/Program.cs
```diff
@@ -0,0 +1,69 @@
...
+            };
+
+            UInt64 expiry = 1234567;
+            try
+            {
+                var result = Idiss.validate_request(global, ip_info, ars_infos, request);
+                Console.WriteLine("Validation was successful. Address: ");
+                Console.WriteLine(result);
+            }
+            catch (RequestValidationException e)
```
orhoj, line 47: You should rethrow after the Console output, otherwise it will continue to the identity creation step (and fail there too). I think it's just:
```
throw;
```
in C#.

eb-concordium: I rearranged it so that both calls are under the same `try`.

idiss/Cargo.toml
```diff
@@ -8,13 +9,14 @@ license-file = "../../LICENSE-APACHE"
 description = "Exports of the identity issuance functionality needed by the identity provider. Currently supports use from nodejs."
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
-
+[profile.release.build-override]
+opt-level = 3
```
abizjak, line 13: Can you please add the same comment here that we have in the mobile-wallet/Cargo.toml as to why this is necessary

abizjak (approved):

Issue #90, opened 2021-11-03
Keygen BLS uses wrong endianness


orhoj:
**Bug Description**
The BLS12-381 keygen algorithm (keygen_bls in rust-src) does not follow the specification entirely (https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-bls-signature-04#section-2.3).

1. In the procedure `I2OSP(L, 2)` is calculated incorrectly compared to the specification. We use little endian, but the algorithm specifies big endian.
2. The calls to `G1::scalar_from_bytes` also assume little endian, but the procedure defines it that bytes are interpreted as big-endian. The byte shifts and slices also need to be adjusted accordingly.

**Steps to Reproduce**
Running the following example:
```
let ikm = b"4827ad00de6fb8bf9df7f14bfa2c70665ce0098d3b7959a31840d1095fa58e06";
let key = keygen_bls(ikm, b"").unwrap();
println!("{}", key);
```

**Expected Result**
The expected result (from another bls library with the keygen implemented):
```
Fr(0x57395d570b6471aa7be4f2fab5e1c5014648e7ef5919af287efe68a317fdaa0e)
```

**Actual Result**
The result from the above input is:
```
Fr(0x150347dbd41b4c6df3d06f1ae54fa551582b9be2e44d5e2dba496ba36af93e66)
```

**Versions**
- Software Version 0.1.0

PR #92, opened 2021-11-08
Fix keygen, increase version, add test vectors and changelog.

eb-concordium:
## Purpose

To fix the BLS keygen. Closes https://github.com/Concordium/concordium-base/issues/90. 

## Changes

* The function `keygen_bls` has been corrected according to the standard (cf. https://tools.ietf.org/html/draft-irtf-cfrg-bls-signature-04#section-2.3) so that it now uses the correct endianness. 
* Test vectors have been added.
* A changelog has been added.
* Version increased to 2.0.0. 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-bins/Cargo.toml
+++ b/rust-bins/Cargo.toml
@@ -41,7 +41,7 @@ anyhow = "1.0"
 
 [dependencies.keygen_bls]
 path = "../rust-src/keygen_bls"
-version = "0"
+version = "2.0.0"
 
 [dependencies.crypto_common]
 path = "../rust-src/crypto_common"
--- a/rust-bins/src/bin/keygen.rs
+++ b/rust-bins/src/bin/keygen.rs
@@ -60,7 +60,11 @@ struct KeygenIp {
     out:         PathBuf,
     #[structopt(long = "out-pub", help = "File to output the public keys to.")]
     out_pub:     PathBuf,
-    #[structopt(long = "v1", help = "Use deprecated version 1 of BLS keygen. If keys were generated with version 1, this flag must be used during recovery.")]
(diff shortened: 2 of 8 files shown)


rust-bins/Cargo.toml
```diff
@@ -41,7 +41,7 @@ anyhow = "1.0"
 
 [dependencies.keygen_bls]
 path = "../rust-src/keygen_bls"
-version = "0"
+version = "2.0.0"
```
abizjak, line 44: For backwards compatibility we also need to make sure that the new version of the tool supports key recovery using the old method.

The keygen tool should have a `--v1` or similar flag that supports key recovery with the legacy keygen_bls version.

abizjak (changes requested):

abizjak:
And please update the version of the keygen tool as well, i.e., this
```
#[structopt(
    about = "Keys for identity providers and anonymity revokers for genesis",
    author = "Concordium",
    version = "2"
)]

File: rust-bins/src/bin/keygen.rs [hunk 2]
```diff
@@ -110,6 +112,8 @@ struct KeygenAr {
                 randomness."
     )]
     only_system_randomness: bool,
+    #[structopt(long = "v1", help = "Use deprecated version 1 of BLS keygen.")]
```
abizjak, line 115: I think it would be useful to elaborate that if they have generated keys with that version they **have to** use that flag during key recovery.

abizjak (approved):
Please elaborate on the need for `--v1`, otherwise looks good.

abizjak (approved):
Please fix the format, and then merge away.

PR #99, opened 2021-11-19
Make separate binaries for user and identity provider.

eb-concordium:
## Purpose
To make two binaries for interaction between an enterprise and an identity provider in order to create an identity and an account.

## Changes

* A binary `user_cli` has been provided for the user's side of the process.
* A binary `identity_provider_cli` has been provided for the identity provider's side of the process.
* The existing binary `client` has been polished a bit.
* Two readmes have been added for the newly introduced binaries.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-bins/docs/user-cli.md
+++ b/rust-bins/docs/user-cli.md
@@ -1,15 +1,55 @@
-This page describes how to use the `user_cli` tool to create an identity and an account interacticting with an identity provider using the tool `identity_provider_cli`.
+This page describes how to use the `user_cli` tool to request an identity object from an identity provider and create accounts from the identity returned by the identity provider.
 
-# Steps to obtain an enterprise identity and account
-The entity uses the binary `user_cli` to do the following steps:
-1. Create secret account holder information and generate the data to be sent to the identity provider using the command
-    ```bash
-    ./user_cli start-ip --private private.json --public public.json --ip-info identity_provider.json --ars anonymity_revokers.json --initial-keys initial_account_keys.json
-    ```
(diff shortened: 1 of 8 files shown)


rust-bins/docs/user-cli.md
```diff
@@ -0,0 +1,55 @@
...
+                          --request-out request.json # request to send to the identity provider
+```
+This will ask for some additional input and output the following files
+- `initial-keys.json` data about the initial account, including its address and keys for signing transactions. DO NOT LOSE THIS FILE. It cannot be recovered.
+- `id-use-data.json` contains data that enables the use of the identity object returned by the identity provider. DO NOT LOSE THIS FILE. It cannot be retrieved.
+- `request.json` contains the request that should be sent to the identity provider.
+
+Assuming everything is in order the identity provider should eventually return the identity object. We refer to it as `id-object.json` in the command below.
+
+# Create accounts from an identity object.
```
eb-concordium, line 32:
```suggestion
# Create accounts from an identity object
```

rust-bins/docs/user-cli.md
```diff
@@ -0,0 +1,55 @@
...
+
+To create a create a credential use the following command.
+```console
+user_cli create-credential --id-use-data id-use-data.json \
+                           --id-object id-object.json \
+                           --keys-out account-keys.json
+                           --credential-out credential.json
+```
+this will output two files
+- `account-keys.json` which contains account keys of the account that will be created by the credential. DO NOT LOSE THIS FILE. It canno be recovered.
```
eb-concordium, line 46:
```suggestion
- `account-keys.json` which contains account keys of the account that will be created by the credential. DO NOT LOSE THIS FILE. It cannot be recovered.
```

abizjak (approved):

PR #111, opened 2021-12-08
Describe keys that are generated by user_cli

abizjak:
## Purpose

Describe keys that are generated by the user-cli tool.


abizjak:
@concordium-cl I'll add documentation on how to decrypt the file once Danielle has written it on the developer documentation.

concordium-cl (approved):
It's good.

Maybe add a sentence saying that the key output is meant to be imported in the client's own tools. This is probably known to the clients themselves but you prevent follow-up questions by non-client ppl like me or Mikael. You can of course also choose to answer the question outside of the document when it comes ;-)

concordium-cl (commented):
Perfect

PR #112, opened 2021-12-08
Add documentation link

abizjak:
## Purpose

Add link to documentation of the utils tool for decrypting.

### DIFF
--- a/rust-bins/docs/user-cli.md
+++ b/rust-bins/docs/user-cli.md
@@ -67,7 +67,7 @@ However if the `account-keys.json` file **is not encrypted** it can be imported 
  ```
  where the `--name` option is optional, and if given, will name the account according to the given value, "my-account" in the example above.
  
-If the `account-keys.json` file is encrypted then it must first be decrypted.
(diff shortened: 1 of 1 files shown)


rust-bins/docs/user-cli.md
```diff
@@ -67,7 +67,7 @@ However if the `account-keys.json` file **is not encrypted** it can be imported
  ```
  where the `--name` option is optional, and if given, will name the account according to the given value, "my-account" in the example above.
  
-If the `account-keys.json` file is encrypted then it must first be decrypted.
+If the `account-keys.json` file is encrypted then it must first be decrypted. This can be done with the `utils` tool which is documented in the [developer documentation](https://developer.concordium.software/en/mainnet/net/references/developer-tools.html)
```
abizjak, line 70:
```suggestion
If the `account-keys.json` file is encrypted then it must first be decrypted. This can be done with the `utils` tool which is documented in the [developer documentation](https://developer.concordium.software/en/mainnet/net/references/developer-tools.html).
```

dg-concordium (approved):

PR #115, opened 2021-12-23
Sync calls

abizjak:
## Purpose

Auxiliary types to support V1 version of smart contracts with synchronous contract calls.

## Changes

- Introduce an "Entrypoint" type analogous to contract name and receive name. This is used when invoking a V1 contract since the name of the contract is already known from the address.
- Introduce new events that can result from contract execution.
- P4 boilerplate. This will have to change by the time of release of P4 node, but is needed to get any testing done of the features.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Genesis/Data/P4.hs
+++ b/haskell-src/Concordium/Genesis/Data/P4.hs
@@ -1,0 +1,107 @@
+-- |This module defines the genesis data fromat for the 'P4' protocol version.
+module Concordium.Genesis.Data.P4 where
+
+import Data.Serialize
+
+import Concordium.Common.Version
+import qualified Concordium.Crypto.SHA256 as Hash
+import Concordium.Genesis.Data.Base
+import Concordium.Genesis.Parameters
+import Concordium.Types
+
+-- |Genesis data for the P4 protocol version. The initial variant is here
+-- because it might be used in the future, at present it is not used.
+data GenesisDataP4
+    = GDP4Initial {
+      -- |The immutable genesis parameters.
+      genesisCore :: !CoreGenesisParameters,
+      -- |Serialized initial block state.
+      -- NB: This block state contains some of the same values as 'genesisCore', and they should match.
+      genesisInitialState :: !GenesisState
+    }
+    | GDP4Regenesis { genesisRegenesis :: !RegenesisData }
+    deriving (Eq, Show)
+
+_core :: GenesisDataP4 -> CoreGenesisParameters
+_core GDP4Initial{..} = genesisCore
+_core GDP4Regenesis{genesisRegenesis=RegenesisData{..}} = genesisCore
+
+instance BasicGenesisData GenesisDataP4 where
+    gdGenesisTime = genesisTime . _core
+    {-# INLINE gdGenesisTime #-}
+    gdSlotDuration = genesisSlotDuration . _core
+    {-# INLINE gdSlotDuration #-}
+    gdMaxBlockEnergy = genesisMaxBlockEnergy . _core
+    {-# INLINE gdMaxBlockEnergy #-}
+    gdFinalizationParameters = genesisFinalizationParameters . _core
+    {-# INLINE gdFinalizationParameters #-}
+    gdEpochLength = genesisEpochLength . _core
+    {-# INLINE gdEpochLength #-}
+
+-- |Deserialize genesis data in the V5 format.
+getGenesisDataV6 :: Get GenesisDataP4
+getGenesisDataV6 =
+    getWord8 >>= \case
+        0 -> do
+            genesisCore <- get
+            genesisInitialState <- get
+            return GDP4Initial{..}
+        1 -> do
+            genesisRegenesis <- getRegenesisData
+            return GDP4Regenesis{..}
+        _ -> fail "Unrecognized P4 genesis data type."
+
+-- |Serialize genesis data in the V6 format.
+putGenesisDataV6 :: Putter GenesisDataP4
+putGenesisDataV6 GDP4Initial{..} = do
+  putWord8 0
+  put genesisCore
+  put genesisInitialState
+putGenesisDataV6 GDP4Regenesis{..} = do
+  putWord8 1
+  putRegenesisData genesisRegenesis
+
+-- |Deserialize genesis data with a version tag. The expected version tag is 5
+-- and this must be distinct from version tags of other genesis data formats.
+getVersionedGenesisData :: Get GenesisDataP4
+getVersionedGenesisData =
+    getVersion >>= \case
+        6 -> getGenesisDataV6
+        n -> fail $ "Unsupported genesis data version for P4 genesis: " ++ show n
+
+-- |Serialize genesis data with a version tag.
+-- This will use the V5 format.
+putVersionedGenesisData :: Putter GenesisDataP4
+putVersionedGenesisData gd = do
+    putVersion 6
+    putGenesisDataV6 gd
+
+parametersToGenesisData :: GenesisParameters -> GenesisDataP4
+parametersToGenesisData = uncurry GDP4Initial . parametersToState
+
+-- |Compute the block hash of the genesis block with the given genesis data.
+-- Every block hash is derived from a message that begins with the block slot,
+-- which is 0 for genesis blocks.
+--
+-- NB: For the regenesis variant the serialized state is not included in the
+-- block hash, only the state hash is. This makes it possible to optimize the
+-- format in the future since it does not have protocol defined meaning. In
+-- contrast, for the initial P4 genesis the initial state is hashed as is.
+genesisBlockHash :: GenesisDataP4 -> BlockHash
+genesisBlockHash GDP4Initial{..} = BlockHash . Hash.hashLazy . runPutLazy $ do
+  put genesisSlot
+  put P4
+  putWord8 0 -- initial variant
+  put genesisCore
+  put genesisInitialState
+genesisBlockHash GDP4Regenesis{genesisRegenesis=RegenesisData{..}} = BlockHash . Hash.hashLazy . runPutLazy $ do
+    put genesisSlot
+    put P4
+    putWord8 1 -- regenesis variant
+    -- NB: 'putRegenesisData' is not used since the state serialization does not go into computing the hash.
+    -- Only the state hash is used.
+    put genesisCore
+    put genesisFirstGenesis
+    put genesisPreviousGenesis
+    put genesisTerminalBlock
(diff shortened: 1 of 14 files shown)


EmilLa1 (approved):
A few comments otherwise lgtm.

haskell-src/Concordium/Types/Execution.hs
```diff
@@ -828,6 +841,14 @@ instance S.Serialize Event where
               TransferMemo {..} ->
                 S.putWord8 21 <>
                 S.put tmMemo
+              Interrupted {..} ->
```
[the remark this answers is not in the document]
abizjak: What magic constants are you referring to? Tags for serialization? How would having them as constants help instead of hinder?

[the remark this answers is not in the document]
EmilLa1: Yes the serialization tags. As I understand, they are part of the public API for the node. I think it would be better to treat it more like such instead of having them inline in the ser/deser- functions. 
I think it could be beneficial for developers integrating with the node and possibly also for future modifications to the transactions. 

This could either be via a document or const's with the appropriate documentation attached. 
But as I said, I don't necessarily think this should be part of this PR, also if we want it to stay like it is, then obviously it shouldn't.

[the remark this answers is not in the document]
abizjak: As discussed on slack, I don't believe this would improve maintainability, correctness, or readability.

I do think serialization of this should be documented, but in a separate document. Having constants for tags is not sufficient documentation to be useful.

haskell-src/Concordium/Genesis/Data/P4.hs
```diff
@@ -0,0 +1,107 @@
...
+--
+-- NB: For the regenesis variant the serialized state is not included in the
+-- block hash, only the state hash is. This makes it possible to optimize the
+-- format in the future since it does not have protocol defined meaning. In
+-- contrast, for the initial P4 genesis the initial state is hashed as is.
+genesisBlockHash :: GenesisDataP4 -> BlockHash
+genesisBlockHash GDP4Initial{..} = BlockHash . Hash.hashLazy . runPutLazy $ do
+  put genesisSlot
+  put P4
+  putWord8 0 -- initial variant
```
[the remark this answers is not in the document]
abizjak: This is the naming for P1-P4, so I'm sticking with it.

haskell-src/Concordium/Genesis/Data/P4.hs
```diff
@@ -0,0 +1,107 @@
...
+putGenesisDataV6 :: Putter GenesisDataP4
+putGenesisDataV6 GDP4Initial{..} = do
+  putWord8 0
+  put genesisCore
+  put genesisInitialState
+putGenesisDataV6 GDP4Regenesis{..} = do
+  putWord8 1
+  putRegenesisData genesisRegenesis
+
+-- |Deserialize genesis data with a version tag. The expected version tag is 5
```
Bargsteen, line 64:
```suggestion
-- |Deserialize genesis data with a version tag. The expected version tag is 6
```

haskell-src/Concordium/Types/InvokeContract.hs
```diff
@@ -0,0 +1,115 @@
...
+-- - amount defaults to 0
+-- - parameter defaults to the empty one
+instance AE.FromJSON ContractContext where
+  parseJSON = AE.withObject "ContractContext" $ \obj -> do
+    ccInvoker <- obj AE..:? "invoker"
+    ccContract <- obj AE..: "contract"
+    ccAmount <- obj AE..:? "amount" AE..!= 0
+    ccMethod <- obj AE..: "method"
+    ccParameter <- obj AE..:? "parameter" AE..!= Wasm.emptyParameter
+    ccEnergy <- obj AE..:? "energy" AE..!= 10_000_000
```
td202, line 49: Maybe this number should be a constant.

haskell-src/Concordium/Genesis/Data/P4.hs
```diff
@@ -0,0 +1,107 @@
...
+-- and this must be distinct from version tags of other genesis data formats.
+getVersionedGenesisData :: Get GenesisDataP4
+getVersionedGenesisData =
+    getVersion >>= \case
+        6 -> getGenesisDataV6
+        n -> fail $ "Unsupported genesis data version for P4 genesis: " ++ show n
+
+-- |Serialize genesis data with a version tag.
+-- This will use the V5 format.
```
Bargsteen, line 73:
```suggestion
-- This will use the V6 format.
```

haskell-src/Concordium/Types/Execution.hs
```diff
@@ -715,6 +722,20 @@ data Event =
...
+               -- |Address of the contract that was interrupted.
+               iAddress :: !ContractAddress,
+               -- |Partial event log generated in the execution before the interrupt.
+               iEvents :: ![Wasm.ContractEvent]
+               }
+           -- | Contract execution resumed. This only applies to V1 contracts.
+           | Resumed {
+               -- |Address of the contract that was interrupted.
+               rAddress :: !ContractAddress,
+               -- |Whether the operation succeeded.
```
Bargsteen, line 736: Perhaps something like:
```suggestion
               -- |Whether the operation, which caused the interrupt, succeeded.
```

haskell-src/Concordium/Types/Execution.hs
```diff
@@ -828,6 +849,14 @@ instance S.Serialize Event where
               TransferMemo {..} ->
                 S.putWord8 21 <>
                 S.put tmMemo
+              Interrupted {..} ->
+                S.putWord8 22 <>
+                S.put iAddress <>
+                putListOf S.put iEvents
```
Bargsteen, line 855: We could use fewer than 8 bytes for the length here as we will never have that many events.

haskell-src/Concordium/Types/InvokeContract.hs
```diff
@@ -0,0 +1,115 @@
...
+  -- invoked, by an account with address 0, no credentials and sufficient amount
+  -- of CCD to cover the transfer amount. If given, the relevant address must
+  -- exist in the blockstate.
+  ccInvoker :: !(Maybe Address),
+  -- |Contract to invoke.
+  ccContract :: !ContractAddress,
+  -- |Amount to invoke the contract with.
+  ccAmount :: !Amount,
+  -- |Which entrypoint to invoke.
+  ccMethod :: !Wasm.ReceiveName,
```
Bargsteen, line 31: Could/should this be a `Wasm.EntrypointName` instead?

td202 (approved):

haskell-src/Concordium/Wasm.hs
```diff
@@ -229,6 +335,38 @@ isValidReceiveName proposal =
...
 
+-- |Name of the entrypoint, i.e., the part of the receive name after the dot.
+newtype EntrypointName = EntrypointName { entrypointName :: Text }
+    deriving (Eq, Show, Ord)
+    deriving(AE.ToJSON) via Text
+
+-- |Check whether the given text is a valid entrypoint name.
+-- This is the case if
+--
+-- * length is <= maxFuncNameSize
```
Bargsteen, line 346: We use the same limit for `ReceiveName`, which also has the `<contract_name>.` prefix. So I'm not sure this is correct.

Bargsteen (approved):
Looks good 👍 Just a few minor issues 😊

haskell-src/Concordium/Types/InvokeContract.hs
abizjak: For consistency with "Update" I prefer to use ReceiveName,

haskell-src/Concordium/Types/Execution.hs
abizjak: This would be incorrect. The contract invoked an operation, which failed. It is not the initiator of the operation that failed.

haskell-src/Concordium/Wasm.hs
abizjak: Well it is correct in as much as it is a definition of what a valid entrypoint name is. You could argue whether it is a sensible definition. We could say `< maxFuncNameSize` to account for the `.` but that does not really improve anything.

haskell-src/Concordium/Types/Execution.hs
abizjak: I added a different clarification.

haskell-src/Concordium/Wasm.hs
```diff
@@ -229,6 +335,38 @@ isValidReceiveName proposal =
...
+    bs <- getByteStringWord16
+    case Text.decodeUtf8' bs of
+      Left _ -> fail "Not a valid utf-8 encoding."
+      Right t | isValidEntrypointName t -> return (EntrypointName t)
+              | otherwise -> fail $ "Not a valid entrypoint name: " ++ Text.unpack t
+
+
+-- |Make a receive name from an init name and an entrypoint name.
+makeReceiveName :: InitName -> EntrypointName -> ReceiveName
+makeReceiveName iName EntrypointName{..} = ReceiveName (initContractName iName <> "." <> entrypointName)
```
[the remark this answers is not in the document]
abizjak: I will add some documentation and rename to `unchecked...` to make it clearer.

I don't want to make the result type a Maybe since it just complicates where this is used for no gain.

haskell-src/Concordium/Types.hs
```diff
@@ -805,15 +806,16 @@ getIncomingAmountsList AccountEncryptedAmount{..} =
...
 -- |Serialization format as specified
 --
 -- * @SPEC: <$DOCS/Transactions#transaction-header>
 instance S.Serialize PayloadSize where
   put (PayloadSize n) = S.putWord32be n
-  get = do
-    thePayloadSize <- S.getWord32be
-    when (thePayloadSize > maxPayloadSize) $ fail "Payload size exceeds maximum transaction payload size."
-    return PayloadSize{..}
+  get = PayloadSize <$> S.getWord32be
```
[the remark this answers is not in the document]
abizjak: I can define those, but I cannot meaningfully version them so they would check validity. This is used in serialization of transaction headers, which are in turn used in serialization of account transactions. There is no protocol version available in those definitions.

I do not have time to embark on adding these versions now.

haskell-src/Concordium/Types/Execution.hs
abizjak: I use 4 bytes now to be consistent with the remaining lists of events (even though 4 bytes is also too much, it must be that for backwards compatibility).

PR #118, opened 2022-01-10
Remove the use of deprecated functions.

### LINKED ISSUE #116: Deprecated use of ::new in ed25519 signatures
**Task description**

The ed25519 crate deprecated the use of the https://docs.rs/ed25519/latest/ed25519/struct.Signature.html#method.new method in 1.3.
We still use it in a few places in crypto_common which causes warnings with 1.3 version.

We should move away from it when convenient, and replace it with from_bytes or something else when appropriate.

abizjak:
## Purpose

Fixes #116 

## Changes

Remove the uses of ::new with ::from

This is purely an internal change. No externally visible change is made.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/crypto_common/src/impls.rs
+++ b/rust-src/crypto_common/src/impls.rs
@@ -194,7 +194,7 @@ impl Deserial for Signature {
     fn deserial<R: ReadBytesExt>(source: &mut R) -> ParseResult<Self> {
         let mut buf = [0u8; SIGNATURE_LENGTH];
         source.read_exact(&mut buf)?;
-        Ok(Signature::new(buf))
+        Ok(Signature::from(buf))
     }
 }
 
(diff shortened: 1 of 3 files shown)


rust-src/crypto_common/src/impls.rs
```diff
@@ -194,7 +194,7 @@ impl Deserial for Signature {
     fn deserial<R: ReadBytesExt>(source: &mut R) -> ParseResult<Self> {
         let mut buf = [0u8; SIGNATURE_LENGTH];
         source.read_exact(&mut buf)?;
-        Ok(Signature::new(buf))
+        Ok(Signature::from(buf))
```
EmilLa1, line 197: What about using either `from_bytes` or `try_from`. `from` will panic if the signature bytes are invalid.

EmilLa1 (changes requested):

rust-src/crypto_common/src/impls.rs
abizjak: No, it won't panic. It is the identity function since Signature is [u8; 64].

abizjak: So in 1.0 it is the identity function, but they changed the semantics in 1.3 apparently. Awesome.

abizjak: I'm using from_bytes now.

I believe this does not change the semantics anywhere else in the project.

PR #128, opened 2022-02-13
Manual instances of Event JSON serialization.

abizjak:
## Purpose

Allow new code to successfully parse old JSON responses. In particular this applies to Init and Update contract. They now have an added contract version field which would break compatibility (in the specific sense mentioned) for the automatically generated parsers.

## Changes

Replace automatically generated instance with manually written ones, defaulting some values during parsing.

I have also added a test that serialization is invertible.

Additionally I have run a test that checks that the new instance is compatible with the automatically generated one. However this is not part of this PR since that property is only interesting once, and is not part of the semantics of the type.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Types/Execution.hs
+++ b/haskell-src/Concordium/Types/Execution.hs
@@ -11,7 +11,7 @@ import Prelude hiding(fail)
 
 import Control.Monad.Reader
 
-import Data.Char
+import qualified Data.Text as Text
 import qualified Data.Aeson as AE
 import Data.Aeson((.=), (.:))
 import Data.Aeson.TH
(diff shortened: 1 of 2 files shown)


EmilLa1 (approved):
lgtm.

I think it could be useful to introduce some test vectors of the serialized types. 
They could both serve as regression tests and could also be useful for integrators.

abizjak:
> lgtm.
> 
> I think it could be useful to introduce some test vectors of the serialized types. They could both serve as regression tests and could also be useful for integrators.

While in the ideal world yes, I simply do not have time for this. There are many events.

Bargsteen (approved):
LGTM 👍

PR #134, opened 2022-03-17
Fix a race condition in the identity provider service.

abizjak:
`remove_file` returns an error if the file does not exist. This, in combination with not checking whether a submission is still pending, can lead to a panic.

## Purpose

Fix a panic due to concurrent requests.

## Changes

Check whether the submission has already been marked finalized/deleted before attempting to delete it again.
Do this in the scope of an acquired lock.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/identity-provider-service/src/bin/main.rs
+++ b/identity-provider-service/src/bin/main.rs
@@ -378,25 +378,45 @@ impl DB {
 
     pub fn mark_finalized(&self, key: &str) {
         let mut lock = self.pending.lock().unwrap();
-        lock.remove(key);
-        let pending_path = self.root.join("pending").join(key);
-        std::fs::remove_file(pending_path).unwrap();
+        if lock.remove(key).is_some() {
+            let pending_path = self.root.join("pending").join(key);
+            if let Err(e) = std::fs::remove_file(pending_path) {
+                error!("Could not delete pending file: {}", e);
+            }
+        } else {
+            log::debug!(
+                "{} could not be marked as finalized, as it has already been removed from the \
+                 pending list.",
+                key
+            );
+        }
     }
 
     pub fn delete_all(&self, key: &str) {
         let mut lock = self.pending.lock().unwrap();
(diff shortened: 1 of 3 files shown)


identity-provider-service/src/bin/main.rs
```diff
@@ -378,25 +378,38 @@ impl DB {
...
-        lock.remove(key);
-        let pending_path = self.root.join("pending").join(key);
-        std::fs::remove_file(pending_path).unwrap();
+        if lock.remove(key).is_some() {
+            let pending_path = self.root.join("pending").join(key);
+            if let Err(e) = std::fs::remove_file(pending_path) {
+                error!("Could not delete pending file: {}", e);
+            }
+        } else {
+            log::debug!("{} already marked as finalized.", key)
```
shjortConcordium, line 387:
```suggestion
            log::debug!("{} could not be marked as finalized, as it has already been removed from pending list.", key)
```

identity-provider-service/src/bin/main.rs
```diff
@@ -378,25 +378,38 @@ impl DB {
...
+                error!("Could not back up the revocation record: {}", e);
+            }
+            if let Err(e) = std::fs::rename(id_path, self.backup_root.join("identity").join(key)) {
+                error!("Could not back up the identity object: {}", e);
+            }
+            if let Err(e) = std::fs::remove_file(pending_path) {
+                error!("Could not delete pending file: {}.", e)
+            }
+        } else {
+            log::debug!("{} already deleted.", key);
```
shjortConcordium, line 411:
```suggestion
            log::debug!("{} could not be deleted, as it has already been removed from pending list.", key);
```

identity-provider-service/src/bin/main.rs
shjortConcordium: Unless we know that it must have been finalized, and not been deleted.

shjortConcordium (approved):

PR #135, opened 2022-04-05
Updates to the mobile wallet library

eb-concordium:
## Purpose

* To expose the newly introduced functions for generating baker keys and creating the transactions `configure baker` and `configure delegation` when building for Android.
* To update the JSON representations of the input to the library so that it becomes consistent with the wallet-proxy.

## Changes

* The JSON serialization of the Rust type `DelegationTarget` has been updated to be consistent with the JSON serialization of the corresponding Haskell type. This implies that the JSON representation of the DelegationTarget in the input to `create_configure_delegation_transaction` in the mobile wallet library is now consistent with the wallet-proxy when looking up `/v0/accBalance/{address}` on a delegator account.
* The JSON serialization of the commission rates in the input to `create_configure_baker_transaction` in the mobile wallet library are now consistent with the wallet-proxy when looking up either the allowed commission ranges at `/v0/chainParameters` or the commission rates of a baker pool at `/v0/bakerPool/{n}`.
* Version increased from `0.10.0` to `0.11.0`.
* The functions `create_configure_delegation_transaction`, `create_configure_baker_transaction` and `generate_baker_keys` are now also exposed when building for Android.

## Checklist

      hard-to-understand areas.


rust-src/crypto_common/src/serialize.rs
```diff
@@ -484,35 +484,6 @@ impl<T: Deserial> Deserial for [T; 8] {
...
-        let x_10 = T::deserial(source)?;
-        let x_11 = T::deserial(source)?;
-        let x_12 = T::deserial(source)?;
-        let x_13 = T::deserial(source)?;
-        let x_14 = T::deserial(source)?;
-        Ok([
-            x_1, x_2, x_3, x_4, x_5, x_6, x_7, x_8, x_9, x_10, x_11, x_12, x_13, x_14,
-        ])
-    }
-}
```
abizjak, line 515: I'm not sure why this specific instance was deleted.

We should now (that const-generics are stable) just implement serial for all N, but that should be in a separate PR.

abizjak (approved):

PR #147, opened 2022-05-06
Upgrades the resolver version to lts-19.6.

rimbi:
Mostly applies the changes in the APIs of the dependencies.

## Purpose

Upgrade the toolchain and the dependencies to the latest version.

## Changes

Upgrades the resolver version to `lts-19.6` and ghc to `9.0.2`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Crypto/ByteStringHelpers.hs
+++ b/haskell-src/Concordium/Crypto/ByteStringHelpers.hs
@@ -10,7 +10,6 @@ import qualified Data.ByteString.Base16 as BS16
 import Data.Text.Encoding as Text
 
 import qualified Data.Aeson as AE
-import qualified Data.Aeson.Encoding as AE
 import qualified Data.Aeson.Types as AE
 import qualified Data.Text as Text
 import Prelude hiding (fail)
@@ -78,10 +77,12 @@ instance FBS.FixedLength a => AE.ToJSON (FBSHex a) where
 
 fbsHexFromText :: forall a . FBS.FixedLength a => Text.Text -> AE.Parser (FBSHex a)
 fbsHexFromText t =
-  let (bs, rest) = BS16.decode (Text.encodeUtf8 t)
-    in if BS.null rest && BS.length bs == FBS.fixedLength (undefined :: a) then return (FBSHex (FBS.fromByteString bs))
-       else if not (BS.null rest) then AE.typeMismatch "Not a valid Base16 encoding." (AE.String t)
-            else AE.typeMismatch "Decoded string not of correct length" (AE.String t)
+  case BS16.decode (Text.encodeUtf8 t) of
+    Right bs ->
+      if BS.length bs == FBS.fixedLength (undefined :: a)
+        then return (FBSHex (FBS.fromByteString bs))
+        else AE.typeMismatch "Decoded string not of correct length" (AE.String t)
+    Left _ -> AE.typeMismatch "Not a valid Base16 encoding." (AE.String t)
 
 instance FBS.FixedLength a => AE.FromJSON (FBSHex a) where
   parseJSON = AE.withText "FixedByteStringHex" $ fbsHexFromText
@@ -90,7 +91,7 @@ instance FBS.FixedLength a => AE.FromJSONKey (FBSHex a) where
   fromJSONKey = AE.FromJSONKeyTextParser fbsHexFromText
 
 instance FBS.FixedLength a => AE.ToJSONKey (FBSHex a) where
-  toJSONKey = AE.ToJSONKeyText serializeBase16 (AE.text . serializeBase16)
+  toJSONKey = AE.toJSONKeyText serializeBase16
 
 -- |Type whose only purpose is to enable derivation of serialization instances.
 newtype Short65K = Short65K ShortByteString
@@ -113,9 +114,9 @@ instance AE.ToJSON Short65K where
 -- |JSON instances based on base16 encoding.
 instance AE.FromJSON Short65K where
   parseJSON = AE.withText "Short65K" $ \t ->
-    let (bs, rest) = BS16.decode (Text.encodeUtf8 t)
-    in if BS.null rest then return (Short65K (BSS.toShort bs))
-       else AE.typeMismatch "Not a valid Base16 encoding." (AE.String t)
+    case BS16.decode (Text.encodeUtf8 t) of
+      Right bs -> return (Short65K (BSS.toShort bs))
+      Left _ -> AE.typeMismatch "Not a valid Base16 encoding." (AE.String t)
 
 
 -- |JSON instances based on base16 encoding.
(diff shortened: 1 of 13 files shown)


rimbi:
This is a work in progress but also mature enough to get your feedback.

haskell-src/Concordium/Crypto/ByteStringHelpers.hs
```diff
@@ -113,9 +114,9 @@ instance AE.ToJSON Short65K where
 -- |JSON instances based on base16 encoding.
 instance AE.FromJSON Short65K where
   parseJSON = AE.withText "Short65K" $ \t ->
-    let (bs, rest) = BS16.decode (Text.encodeUtf8 t)
-    in if BS.null rest then return (Short65K (BSS.toShort bs))
-       else AE.typeMismatch "Not a valid Base16 encoding." (AE.String t)
+    case BS16.decode (Text.encodeUtf8 t) of
+      Right bs -> return (Short65K (BSS.toShort bs))
+      Left _ -> AE.typeMismatch "Not a valid Base16 encoding." (AE.String t)
```
abizjak, line 119: For debugging it would be useful to keep the error.

haskell-src/Concordium/Crypto/ByteStringHelpers.hs
```diff
@@ -125,9 +126,9 @@ instance AE.ToJSON ByteStringHex where
 -- |JSON instances based on base16 encoding.
 instance AE.FromJSON ByteStringHex where
   parseJSON = AE.withText "ByteStringHex" $ \t ->
-    let (bs, rest) = BS16.decode (Text.encodeUtf8 t)
-    in if BS.null rest then return (ByteStringHex (BSS.toShort bs))
-       else AE.typeMismatch "Not a valid Base16 encoding." (AE.String t)
+    case BS16.decode (Text.encodeUtf8 t) of
+      Right bs -> return (ByteStringHex (BSS.toShort bs))
+      Left _ -> AE.typeMismatch "Not a valid Base16 encoding." (AE.String t)
```
abizjak, line 131: For debugging it would be useful to keep the error.

haskell-src/Concordium/Crypto/ByteStringHelpers.hs
rimbi: when needed it can be changed

haskell-src/Concordium/Crypto/ByteStringHelpers.hs
rimbi: when needed it can be changed

haskell-src/Concordium/Types/Transactions.hs
```diff
@@ -36,6 +36,56 @@ import Concordium.Types.Updates
...
+      <> S.put thExpiry
+
+  get = do
+    thSender <- S.get
+    thNonce <- S.get
+    thEnergyAmount <- S.get
+    thPayloadSize <- S.get
+    thExpiry <- S.get
+    return $! TransactionHeader {..}
+
```
abizjak, line 88: I do not understand this change. 

This PR is meant to update the toolchains. We cannot at the same time be moving things around. That is a recipe for problems and needless conflicts.

Is this moved because of a staging restriction, or why is it necessary?

stack.yaml
```diff
@@ -59,7 +59,8 @@ extra-lib-dirs:
 # arch: x86_64
 #
 # Extra directories used by stack for building
-# extra-include-dirs: [/path/to/dir]
+extra-include-dirs:
+    - /Library/Developer/CommandLineTools/SDKs/MacOSX12.3.sdk/usr/include/ffi/
```
abizjak, line 63: We cannot have this in, it is a local directory on your machine.

haskell-src/Concordium/Types/Transactions.hs
rimbi: I am not in favor of unnecessary changes either. Do you remember our discussion about the `deriveJSON` macro not working? It turned out to be a problem with the order of the declarations. I don't know the exact reason but with the `9.0.2` the current order of the declarations has to be changed to get rid of the compile problems.

stack.yaml
rimbi: I don't have a better solution for that atm. Do you?

haskell-src/Concordium/Utils.hs
```diff
@@ -23,11 +23,11 @@ class (Ixed m) => At' m where
   at' :: Index m -> Lens' m (Maybe (IxValue m))
 
 instance (Hashable k, Eq k) => At' (H.HashMap k v) where
-  at' = flip H.alterF
+  at' k f = H.alterF f k
```
abizjak, line 26: Why was this necessary or useful?

rimbi: It was necessary, did not dig into the details, might be the same/similar issue to the one Thomas pointed. (eta-expansion)

td202, line 26:
```suggestion
  at' = \k f -> H.alterF f k
```
While the eta-expansion is necessary, I think expressing it as a lambda may give the optimiser more chance of inlining. Lenses in particular are often composed rather than directly applied.

td202 (approved):

haskell-src/Concordium/Utils.hs
rimbi: Thanks for the suggestion Thomas. On the other I will not use it this time as the linter complains about it.

rimbi: And here is the error message with the original format:

```console
/Users/cemeliguzel/scratch/concordium-client/deps/concordium-base/haskell-src/Concordium/Utils.hs:26:9: error:
    • Couldn't match type: k0
                           -> H.HashMap k0 v0 -> f0 (H.HashMap k0 v0)
                     with: forall (f :: * -> *).
                           Functor f =>
                           (Maybe (IxValue (H.HashMap k v))
                            -> f (Maybe (IxValue (H.HashMap k v))))
                           -> H.HashMap k v -> f (H.HashMap k v)
      Expected: Index (H.HashMap k v)
                -> Lens' (H.HashMap k v) (Maybe (IxValue (H.HashMap k v)))
        Actual: (Maybe v0 -> f0 (Maybe v0))
                -> k0 -> H.HashMap k0 v0 -> f0 (H.HashMap k0 v0)
    • In the expression: H.alterF
      In an equation for ‘at'’: at' = H.alterF
      In the instance declaration for ‘At' (H.HashMap k v)’
    • Relevant bindings include
        at' :: Index (H.HashMap k v)
               -> Lens' (H.HashMap k v) (Maybe (IxValue (H.HashMap k v)))
          (bound at haskell-src/Concordium/Utils.hs:26:3)
   |
26 |   at' = H.alterF
   |         ^^^^^^^^
```

Issue #55, opened 2021-07-09
Issues with MintRate JSON instance/tests


abizjak:
Specifically on numbers where the mantissa is of the form `n00`, `n000`, ... with the exponent respectively `-1`, `-2`, ... fail to parse.

While these should not appear in practice, since they would lead to more than 100% increase of total GTU per slot, and it would make sense to have the type only for values <= 1, that is not currently the case, and it does lead to globalstate-types tests occasionally failing.

PR #149, opened 2022-05-18
Added several Arbitrary instances

mkmks:
## Purpose

Tests added in https://github.com/Concordium/concordium-node/pull/335 require several types to have instances of `Arbitrary`.

## Changes

`AmountFraction`, `Slot`, `MintDistribution`, `PartsPerHundredThousand` got instances of Arbitrary.

## Checklist

      hard-to-understand areas.

## CLA acceptance

### DIFF
--- a/haskell-src/Concordium/Types.hs
+++ b/haskell-src/Concordium/Types.hs
@@ -225,7 +225,7 @@ import Lens.Micro.Platform
 
 import Text.Read (readMaybe)
 
-import Test.QuickCheck ( Arbitrary )
+import Test.QuickCheck ( Arbitrary, choose )
 import Test.QuickCheck.Arbitrary (Arbitrary(arbitrary))
 
 -- |A value equipped with its hash.
@@ -279,7 +279,7 @@ type LotteryPower = Ratio Amount
 -- This wrapper will be used by both @AmountFraction@ and @ElectionDifficulty@.
 -- It was agreed in tokenomics discussions to be sufficient.
 newtype PartsPerHundredThousands = PartsPerHundredThousands { partsPerHundredThousand :: Word32 }
-  deriving newtype (Eq, Ord, Num, Real, Enum, Integral, Arbitrary)
+  deriving newtype (Eq, Ord, Num, Real, Enum, Integral)
 
 hundredThousand :: Word32
 hundredThousand = 100000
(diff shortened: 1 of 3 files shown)


haskell-src/Concordium/Types.hs
```diff
@@ -529,6 +535,12 @@ instance HashableTo Hash.Hash MintRate where
 
 instance Monad m => MHashableTo m Hash.Hash MintRate
 
+instance Arbitrary MintRate where
+  arbitrary = do
+    mrMantissa <- arbitrary
+    mrExponent <- arbitrary
+    return MintRate{..}
```
abizjak, line 542: There are some unstated assumptions related to this type. The intention seems to be that the values are reasonable, and some functionality depends on this, see e.g. https://github.com/Concordium/concordium-base/issues/55

But here you are generating arbitrary values that are in general completely nonsensical for the mint rate.

mkmks: The idea behind (random) testing is exactly that: to uncover hidden assumptions and codify them. What does it mean for a `MintRate` value to be "reasonable"?

abizjak: I do not need generic lessons about randomized testing. I am well aware of it and its value and limitations. And in fact that is **exactly** how that issue that is linked was discovered.

I simply pointed out that what you are doing here contradicts some of the code that already exists. In particular the JSON parsing does not handle values as described in the linked issue.

In practice there would be no loss of generality if values were always less than 1.

mkmks: It would be much more constructive and to the point to state exactly that in your initial comment:

"The values of `MintRate` are expected to be less than 1."

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -101,6 +103,11 @@ instance IsChainParametersVersion cpv => Serialize (MintDistribution cpv) where
 instance IsChainParametersVersion cpv => HashableTo Hash.Hash (MintDistribution cpv) where
   getHash = Hash.hash . encode
 
+instance Arbitrary (MintDistribution 'ChainParametersV1) where
+  arbitrary = do
+    mps <- vector 2 `suchThat` (\[x, y] -> isJust $ addAmountFraction x y)
+    return $ MintDistribution MintPerSlotForCPV0None (mps !! 0) (mps !! 1)
```
td202, line 109:
```suggestion
    (x, y) <- arbitrary `suchThat` (\(x, y) -> isJust $ addAmountFraction x y)
    return $ MintDistribution MintPerSlotForCPV0None x y
```

td202 (approved):
I am a bit concerned that this is duplicating functionality from Generators, so I suggest we make a task to rationalise them.

mkmks:
I think if we create an issue for that, we'll never come back to it. Can we simply export Generators from concordium-base?

haskell-src/Concordium/Types/Parameters.hs
td202: I think it might be better to choose the first value from the entire range, and then the second from the rest, avoiding using `suchThat`.

td202: On consideration, probably this approach is fine given the fix to the PartsPerHundredThousands instance, since each choice of the two values has a 50% chance of being acceptable.

PR #157, opened 2022-05-25
Create libraries for deterministic key derivation

orhoj:
## Purpose
Add library support for the mobile wallets to derive identity and account keys from a seed phrase, as well as blinding  randomness and attribute commitment randomness.

## Changes
- Added a library (_ed25519_hd_key_derivation_) that implements [SLIP0010](https://github.com/satoshilabs/slips/blob/master/slip-0010.md) for ed25519 specifically.
- Added a library (_key_derivation_) that implements the Concordium key derivation as specified in the cryptographic blue paper.
    - Note that the tests for this library check keys against keys I have derived using a TypeScript implementation of the same key derivation scheme. I will be testing by deriving keys for the same paths in the Ledger repository afterwards as extra verification.
- Added Android and iOS wrappers for getting identity keys, blinding randomness, accounts keys and attribute randomness.
   - `get_identity_keys_and_randomness` returns `id_cred_sec`, `prf_key` and `blinding_randomness`
   - `get_account_keys_and_randomness` returns `signing_key`, `verify_key` and `attribute_randomness`. Note that attribute randomness for _all_ attributes will be returned.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/mobile_wallet/mobile_wallet.h
+++ b/mobile_wallet/mobile_wallet.h
@@ -166,6 +166,36 @@ char *generate_baker_keys(uint8_t *success);
 uint64_t decrypt_encrypted_amount(const char *input_ptr, uint8_t *success);
 
 /**
+ * Take a pointer to a NUL-terminated UTF8-string and return a NUL-terminated
+ * UTF8-encoded string. The returned string must be freed by the caller by
+ * calling the function 'free_response_string'. In case of failure the function
+ * returns an error message as the response, and sets the 'success' flag to 0.
+ *
+ * See rust-bins/wallet-notes/README.md for the description of input and output
+ * formats.
+ *
+ * # Safety
+ * The input pointer must point to a null-terminated buffer, otherwise this
+ * function will fail in unspecified ways.
+ */
+char *get_identity_keys_and_randomness(const char *input_ptr, uint8_t *success);
+
+/**
+ * Take a pointer to a NUL-terminated UTF8-string and return a NUL-terminated
+ * UTF8-encoded string. The returned string must be freed by the caller by
+ * calling the function 'free_response_string'. In case of failure the function
+ * returns an error message as the response, and sets the 'success' flag to 0.
+ *
+ * See rust-bins/wallet-notes/README.md for the description of input and output
+ * formats.
+ *
+ * # Safety
+ * The input pointer must point to a null-terminated buffer, otherwise this
+ * function will fail in unspecified ways.
+ */
+char *get_account_keys_and_randomness(const char *input_ptr, uint8_t *success);
+
+/**
  * # Safety
  * This function is unsafe in the sense that if the argument pointer was not
  * Constructed via CString::into_raw its behaviour is undefined.
--- a/mobile_wallet/src/android.rs
+++ b/mobile_wallet/src/android.rs
@@ -6,6 +6,7 @@ use crate::{
     create_configure_delegation_transaction, create_credential, create_encrypted_transfer,
     create_id_request_and_private_data, create_pub_to_sec_transfer, create_sec_to_pub_transfer,
     create_transfer, decrypt_encrypted_amount, generate_accounts, generate_baker_keys,
+    get_account_keys_and_randomness, get_identity_keys_and_randomness,
 };
 use jni::{
     objects::{JClass, JString, JValue},
@@ -125,6 +126,84 @@ pub extern "system" fn Java_com_concordium_mobile_1wallet_1lib_WalletKt_generate
     let mut success: u8 = 127;
     let cstr_res = unsafe {
         let unsafe_res_ptr = generate_accounts(input_str.as_ptr(), &mut success);
+        if unsafe_res_ptr.is_null() {
+            return wrap_return_tuple(&env, 127, "Pointer returned from crypto library was NULL");
+        }
+        CString::from_raw(unsafe_res_ptr)
+    };
+
+    match cstr_res.to_str() {
+        Ok(str_ref) => wrap_return_tuple(&env, success, str_ref),
+        Err(e) => wrap_return_tuple(
+            &env,
+            127,
+            &format!("Could not read CString from crypto library {:?}", e),
+        ),
+    }
+}
+
+#[no_mangle]
+pub extern "system" fn Java_com_concordium_mobile_1wallet_1lib_WalletKt_get_1identity_1keys_1and_1randomness(
+    env: JNIEnv,
+    _: JClass,
+    input: JString,
+) -> jobject {
+    let input_str = match env.get_string(input) {
+        Ok(res_str) => res_str,
+        Err(e) => {
+            return wrap_return_tuple(
+                &env,
+                127,
+                &format!(
+                    "Could not read java.lang.String given as input due to {:?}",
+                    e
+                ),
+            )
+        }
+    };
+
+    let mut success: u8 = 127;
+    let cstr_res = unsafe {
+        let unsafe_res_ptr = get_identity_keys_and_randomness(input_str.as_ptr(), &mut success);
+        if unsafe_res_ptr.is_null() {
+            return wrap_return_tuple(&env, 127, "Pointer returned from crypto library was NULL");
+        }
+        CString::from_raw(unsafe_res_ptr)
+    };
+
+    match cstr_res.to_str() {
+        Ok(str_ref) => wrap_return_tuple(&env, success, str_ref),
+        Err(e) => wrap_return_tuple(
+            &env,
+            127,
+            &format!("Could not read CString from crypto library {:?}", e),
+        ),
+    }
+}
+
+#[no_mangle]
+pub extern "system" fn Java_com_concordium_mobile_1wallet_1lib_WalletKt_get_1account_1keys_1and_1randomness(
+    env: JNIEnv,
+    _: JClass,
+    input: JString,
+) -> jobject {
+    let input_str = match env.get_string(input) {
+        Ok(res_str) => res_str,
+        Err(e) => {
+            return wrap_return_tuple(
+                &env,
+                127,
+                &format!(
+                    "Could not read java.lang.String given as input due to {:?}",
+                    e
+                ),
+            )
+        }
+    };
+
+    let mut success: u8 = 127;
+    let cstr_res = unsafe {
+        let unsafe_res_ptr = get_account_keys_and_randomness(input_str.as_ptr(), &mut success);
         if unsafe_res_ptr.is_null() {
             return wrap_return_tuple(&env, 127, "Pointer returned from crypto library was NULL");
         }
--- a/mobile_wallet/src/lib.rs
+++ b/mobile_wallet/src/lib.rs
@@ -824,7 +824,7 @@ fn get_account_keys_and_randomness_aux(input: &str) -> anyhow::Result<String> {
             attribute_tag.0.into(),
         );
         let commitment_randomness_hex = hex::encode(commitment_randomness);
-        attribute_commitment_randomness.insert(attribute_tag, commitment_randomness_hex);
+        attribute_commitment_randomness.insert(attribute_tag.0, commitment_randomness_hex);
     }
 
     let response = json!({
--- a/mobile_wallet/src/lib.rs
+++ b/mobile_wallet/src/lib.rs
@@ -831,7 +831,7 @@ fn get_account_keys_and_randomness_aux(input: &str) -> anyhow::Result<String> {
     }
 
     let response = json!({
-        "signingKey": account_signing_key_hex,
+        "signKey": account_signing_key_hex,
         "verifyKey": account_verify_key_hex,
         "attributeCommitmentRandomness": attribute_commitment_randomness
     });
--- a/mobile_wallet/src/lib.rs
+++ b/mobile_wallet/src/lib.rs
@@ -787,18 +787,15 @@ fn get_identity_keys_and_randomness_aux(input: &str) -> anyhow::Result<String> {
     let identity_index = try_get(&v, "identityIndex")?;
 
     let id_cred_sec = wallet.get_id_cred_sec(identity_index)?;
-    let id_cred_sec_hex = hex::encode(id_cred_sec);
 
     let prf_key = wallet.get_prf_key(identity_index)?;
-    let prf_key_hex = hex::encode(prf_key);
 
     let blinding_randomness = wallet.get_blinding_randomness(identity_index)?;
-    let blinding_randomness_hex = hex::encode(blinding_randomness);
 
     let response = json!({
-        "idCredSec": id_cred_sec_hex,
-        "prfKey": prf_key_hex,
-        "blindingRandomness": blinding_randomness_hex
+        "idCredSec": base16_encode_string(&id_cred_sec),
+        "prfKey": base16_encode_string(&prf_key),
+        "blindingRandomness": base16_encode_string(&blinding_randomness)
     });
     Ok(to_string(&response)?)
 }
@@ -826,7 +823,7 @@ fn get_account_keys_and_randomness_aux(input: &str) -> anyhow::Result<String> {
             account_credential_index,
             attribute_tag,
         )?;
-        let commitment_randomness_hex = hex::encode(commitment_randomness);
+        let commitment_randomness_hex = base16_encode_string(&commitment_randomness);
         attribute_commitment_randomness.insert(attribute_tag.0, commitment_randomness_hex);
     }
 
--- a/mobile_wallet/src/lib.rs
+++ b/mobile_wallet/src/lib.rs
@@ -770,11 +770,14 @@ fn decrypt_encrypted_amount_aux(input: &str) -> anyhow::Result<Amount> {
 
 fn parse_wallet_input(v: &Value) -> anyhow::Result<ConcordiumHdWallet> {
     let seed_hex: String = try_get(v, "seed")?;
-    let net: Net = try_get(v, "net")?;
-    let wallet = ConcordiumHdWallet {
-        seed: hex::decode(&seed_hex)?,
-        net,
+    let seed_decoded = hex::decode(&seed_hex)?;
+    let seed: [u8; 64] = match seed_decoded.try_into() {
+        Ok(s) => s,
+        Err(_) => bail!("The provided seed {} was not 64 bytes", seed_hex),
     };
+
+    let net: Net = try_get(v, "net")?;
+    let wallet = ConcordiumHdWallet { seed, net };
     Ok(wallet)
 }
 
@@ -783,13 +786,13 @@ fn get_identity_keys_and_randomness_aux(input: &str) -> anyhow::Result<String> {
     let wallet = parse_wallet_input(&v)?;
     let identity_index = try_get(&v, "identityIndex")?;
 
-    let id_cred_sec = wallet.get_id_cred_sec(identity_index);
+    let id_cred_sec = wallet.get_id_cred_sec(identity_index)?;
     let id_cred_sec_hex = hex::encode(id_cred_sec);
 
-    let prf_key = wallet.get_prf_key(identity_index);
+    let prf_key = wallet.get_prf_key(identity_index)?;
     let prf_key_hex = hex::encode(prf_key);
 
-    let blinding_randomness = wallet.get_blinding_randomness(identity_index);
+    let blinding_randomness = wallet.get_blinding_randomness(identity_index)?;
     let blinding_randomness_hex = hex::encode(blinding_randomness);
 
     let response = json!({
@@ -807,11 +810,11 @@ fn get_account_keys_and_randomness_aux(input: &str) -> anyhow::Result<String> {
     let account_credential_index = try_get(&v, "accountCredentialIndex")?;
 
     let account_signing_key =
-        wallet.get_account_signing_key(identity_index, account_credential_index);
+        wallet.get_account_signing_key(identity_index, account_credential_index)?;
     let account_signing_key_hex = hex::encode(account_signing_key);
 
     let account_verify_key =
-        wallet.get_account_public_key(identity_index, account_credential_index);
+        wallet.get_account_public_key(identity_index, account_credential_index)?;
     let account_verify_key_hex = hex::encode(account_verify_key);
 
     let mut attribute_commitment_randomness = HashMap::new();
@@ -821,8 +824,8 @@ fn get_account_keys_and_randomness_aux(input: &str) -> anyhow::Result<String> {
         let commitment_randomness = wallet.get_attribute_commitment_randomness(
             identity_index,
             account_credential_index,
-            attribute_tag.0.into(),
-        );
+            attribute_tag,
+        )?;
         let commitment_randomness_hex = hex::encode(commitment_randomness);
         attribute_commitment_randomness.insert(attribute_tag.0, commitment_randomness_hex);
     }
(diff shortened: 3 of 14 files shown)


rust-src/key_derivation/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+        .rev()
+        .flat_map(|val| val.to_be_bytes())
+        .collect();
+    let mut bls_key = [0u8; 32];
+    bls_key.clone_from_slice(&key_bytes[0..32]);
+    bls_key
+}
+
+pub struct ConcordiumHdWallet {
+    pub seed: Vec<u8>,
```
orhoj, line 37: We may want to impose a restriction that this only allows for a 24 word seed phrase, i.e. for 64 bytes instead of it being adjustable.

rust-src/ed25519_hd_key_derivation/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+    segments
+        .iter()
+        .skip(1)
+        .map(|seg| seg.replace("'", ""))
+        .all(|seg| seg.parse::<u32>().is_ok())
+}
+
+/// Derives hierarchical deterministic keys for ed25519 according to the SLIP0010 (https://github.com/satoshilabs/slips/blob/master/slip-0010.md)
+/// specification.
+pub fn derive(path: &str, seed: &[u8]) -> Result<HdKeys, String> {
```
orhoj, line 67: The `seed` should be validated to be between `16` and `64` bytes as this is required by the SLIP0010 specification.

rust-src/ed25519_hd_key_derivation/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+    let mut ir = [0u8; 32];
+    ir.clone_from_slice(&i[32..64]);
+
+    HdKeys {
+        key:        il,
+        chain_code: ir,
+    }
+}
+
+fn is_valid_path(path: &str) -> bool {
```
abizjak, line 51: Please add a doc comment specifying what a valid path is.

rust-src/ed25519_hd_key_derivation/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+    let path_regex = Regex::new("^m(/[0-9]+')+$").unwrap();
```
abizjak, line 52: I'd prefer not to bring in regex just for this, but I see that it is not that straightforward to implement this check.
I suppose this is not used in any context with malicious input where we'd have to worry about the performance of regex matching and its pathologies.

rust-src/ed25519_hd_key_derivation/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+    if !path_regex.is_match(path) {
+        return false;
+    }
+
+    let segments: Vec<&str> = path.split('/').collect();
```
abizjak, line 57: I don't think you need a collect there since you're immediately iterating over the collection.

rust-src/ed25519_hd_key_derivation/src/lib.rs
abizjak: Is this format of the path something we came up with or is it standard. The use of `'` is a bit weird.

rust-src/ed25519_hd_key_derivation/src/lib.rs
abizjak: Shouldn't this be added both to the documentation of this function, as well as checked at the start.

abizjak, line 67: It works better if errors are structured. Some time ago we started using `thiserror` for this. In this case it would look like
```rust
use thiserror::Error;
#[derive(Debug, Error)]
pub enum DeriveError {
    #[error("Invalid derivation path.")]
    InvalidPath,
    #[error("Seed must be between 128 and 512 bits.")]
    InvalidSeed,
}
```

rust-src/ed25519_hd_key_derivation/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+
+    let master_key = get_master_key_from_seed(&seed);
+
+    let segments: Vec<&str> = path.split('/').collect();
+    let segments = segments
+        .iter()
+        .skip(1)
+        .map(|seg| seg.replace("'", ""))
+        .map(|seg| seg.parse::<u32>().unwrap())
+        .collect::<Vec<_>>();
```
abizjak, line 80: It might be useful to avoid this duplication. Instead you'd have 
```
fn is_valid_path(path: &str) -> Result<Vec<u32>, DeriveError> {
...
}
```
to avoid duplicating this rather complicated path transformation.

rust-src/ed25519_hd_key_derivation/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    let master_key = get_master_key_from_seed(&seed);
+
+    let segments: Vec<&str> = path.split('/').collect();
+    let segments = segments
+        .iter()
+        .skip(1)
+        .map(|seg| seg.replace("'", ""))
+        .map(|seg| seg.parse::<u32>().unwrap())
```
abizjak, line 79: If I understand correctly here (and below on lines 82/83), if I have paths `m/01'` and `m/1'` will lead to the same `HdKeys`. Is that the intention?

rust-src/ed25519_hd_key_derivation/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+        assert_eq!(
+            public_key_derived.to_vec(),
+            hex::decode(public_key).unwrap()
+        )
+    }
+
+    #[test]
+    pub fn test_valid_path() {
+        let valid_path = "m/44'/919'";
+        assert!(is_valid_path(&valid_path));
```
abizjak, line 111: I usually find it is helpful to provide a message to be printed when the assertion fails.
```rust
        assert!(is_valid_path(&valid_path), "Path {} should be valid.", valid_path);
```
I find it easier to locate the errors when tests fail.

rust-src/key_derivation/src/lib.rs
abizjak: Is there a reason to be more general at this point? Otherwise I would prefer that this is more strict.

Additionally, this is part of the public API so should be well-documented. It is also useful to have Debug implementation however in this case it should be manual to not print the seed I imagine?

rust-src/ed25519_hd_key_derivation/src/lib.rs
```diff
@@ -0,0 +1,274 @@
+use hmac::{Hmac, Mac};
+use regex::Regex;
+use sha2::Sha512;
+
+const ED25519_CURVE: &[u8; 12] = b"ed25519 seed";
+const HARDENED_OFFSET: u32 = 0x80000000;
```
abizjak, line 6: If I understood correctly in the actual derivations for the ConcordiumHdWallet below, this offset limits the range of allowed indices to < 2^31. Is that correct?

rust-src/key_derivation/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+
+    pub fn get_prf_key(&self, identity_index: u32) -> [u8; 32] {
+        let path = self.make_path(&format!("{}'/3'", identity_index));
+        let prf_key_seed = derive(&path, &self.seed).unwrap().key;
+        bls_key_bytes_from_seed(prf_key_seed)
+    }
+
+    pub fn get_blinding_randomness(&self, identity_index: u32) -> [u8; 32] {
+        let path = self.make_path(&format!("{}'/4'", identity_index));
+        let blinding_randomness_seed = derive(&path, &self.seed).unwrap().key;
```
abizjak, line 74: I think the same comment applies to this as to `get_account_signing_key`.

abizjak (commented):
I think it looks good overall. I like the test vectors.

I think there are some issues in error handling and the precise semantics of the functions. And documentation of the rust library is lacking.

rust-src/ed25519_hd_key_derivation/src/lib.rs
orhoj: Yes. It was meant as a note for myself for today. I have added a check and added it to the function documentation.

mobile_wallet/src/lib.rs
```diff
@@ -766,6 +768,73 @@ fn decrypt_encrypted_amount_aux(input: &str) -> anyhow::Result<Amount> {
...
+    for attribute_name in ATTRIBUTE_NAMES {
+        let attribute_tag = AttributeTag::from_str(attribute_name)?;
+        let commitment_randomness = wallet.get_attribute_commitment_randomness(
+            identity_index,
+            account_credential_index,
+            attribute_tag.0.into(),
+        );
+        let commitment_randomness_hex = hex::encode(commitment_randomness);
+        attribute_commitment_randomness.insert(attribute_tag.0, commitment_randomness_hex);
+    }
```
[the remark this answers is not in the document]
orhoj: I was debating the two options: return all the randomness, or let the caller send a list of attributes to get the randomness for. I went with the first because that was simpler on our end, but either way works for me. Let me know if you prefer them providing the list to get the attributes for.

In that regard, should I use the attribute_tag (i.e. the index 0, 1, 2, ...) or the text representation of the attributes?

mobile_wallet/src/lib.rs
```diff
@@ -766,6 +768,73 @@ fn decrypt_encrypted_amount_aux(input: &str) -> anyhow::Result<Amount> {
...
+
+    let response = json!({
+        "signingKey": account_signing_key_hex,
+        "verifyKey": account_verify_key_hex,
+        "attributeCommitmentRandomness": json!(attribute_commitment_randomness)
```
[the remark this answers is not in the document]
orhoj: You are right. I have removed it.

rust-bins/wallet-notes/README.md
```diff
@@ -444,6 +444,57 @@ The return value is a a JSON array with JSON objects as entries. Each object has
...
+## get_identity_keys_and_randomness
+
+Semantics: Deterministically derives id_cred_sec, prf_key and blinding randomness for an identity.
+
+This function takes as input a NUL-terminated UTF8-encoded string. The string
+must be a valid JSON object with fields
+
+- `"seed"` ... the seed used to derive keys from, as a hex string.
+
+- `"net"` ... determines whether to derive keys for Mainnet or a testnet. Has to be "Mainnet" or "Testnet", all other values will fail.
```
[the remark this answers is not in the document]
orhoj: Casing is indeed relevant as it is parsed into the corresponding `enum`. I have adjusted the documentation accordingly.

rust-src/ed25519_hd_key_derivation/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+
+pub struct HdKeys {
+    pub key:    [u8; 32],
+    chain_code: [u8; 32],
```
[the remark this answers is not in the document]
orhoj: I had no use for it for this exact use case, but for the library to be complete it really should be public. So I have changed that.

rust-src/ed25519_hd_key_derivation/src/lib.rs
orhoj: For our own use cases the `path` will always be valid. I think checking it "manually" will be quite messy.

The format of the path is the standard way of denoting a hardened path, so we cannot get rid of that. I assume it is because all the numbers become quite unmanageable to read if they are all greater than `2^31`.

rust-src/ed25519_hd_key_derivation/src/lib.rs
orhoj: That's way better. Good to know.

chrmatt (approved):
The paths and how keys are generated seem to match the bluepaper. Looks good overall.

rust-src/key_derivation/src/lib.rs
chrmatt: I think in principle, people could make their own wallet with 12 word seed phrases. It's less secure, so we don't do it in our wallets, but it should still work...

rust-src/ed25519_hd_key_derivation/src/lib.rs
chrmatt: BIP-0032 defines "hardened" and "non-hardened" keys, where for non-hardened ones, you can derive child keys from parent keys, hardened keys don't allow that. Adding a `'` (and setting the highest order bit) indicates that you want a hardened key. We only use hardened keys.

rust-src/ed25519_hd_key_derivation/src/lib.rs
chrmatt: That's a good point. The indices are meant to be integers, so 01 = 1 and I guess it's intentional. Maybe we don't want to allow 01 in valid paths to avoid this confusion. Maybe it would actually be better to have the type of the path be a list of integers instead of a string.

rust-src/ed25519_hd_key_derivation/src/lib.rs
chrmatt: Yes.

rust-src/ed25519_hd_key_derivation/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+    HdKeys {
+        key:        il,
+        chain_code: ir,
+    }
+}
+
+fn ckd_priv(keys: HdKeys, index: u32) -> HdKeys {
+    if index & HARDENED_OFFSET == 0 {
+        panic!("Received a non-hardened index!");
+    }
```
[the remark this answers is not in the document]
orhoj: I changed it to return a `Result` and updated `derive` accordingly.

mobile_wallet/src/lib.rs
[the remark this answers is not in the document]
abizjak: I think this option is fine. In any case it's us who'll have to handle it in a followup call to create_credential. This is fine.

I prefer string tags. Those are meant for JSON.

rust-src/ed25519_hd_key_derivation/src/lib.rs
abizjak: Ok. It's a weird format, but if it is standard I have no further comments or objections.

rust-src/ed25519_hd_key_derivation/src/lib.rs
abizjak: Yes, that would be useful and cleaner (see also my comment below about constructing a string and then decomposing it just a bit later).

It sounds to me like paths are standard, so they should still be supported. But it would be nice to have an API based on Iter<Item=u32> or maybe just &[u32].

rust-src/ed25519_hd_key_derivation/src/lib.rs
orhoj: That is correct. I have updated the validation to ensure a failure occurs if an index is greater than `2^31`.

rust-src/ed25519_hd_key_derivation/src/lib.rs
orhoj: I refactored it like this to avoid parsing through the segments multiple times.

mobile_wallet/src/android.rs
```diff
@@ -141,6 +142,84 @@ pub extern "system" fn Java_com_concordium_mobile_1wallet_1lib_WalletKt_generate
...
+    env: JNIEnv,
+    _: JClass,
+    input: JString,
+) -> jobject {
+    let input_str = match env.get_string(input) {
+        Ok(res_str) => res_str,
+        Err(e) => {
+            return wrap_return_tuple(
+                &env,
+                127,
```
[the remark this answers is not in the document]
abizjak: This is just JNI code mapping. I believe this ultimately translates to "Unknown error".

rust-src/ed25519_hd_key_derivation/src/lib.rs
orhoj: That is correctly understood, and I guess just a bit too lenient to have it implicitly mean the same. I have updated the regex to disallow indices with prefixed `0`'s.

rust-src/ed25519_hd_key_derivation/src/lib.rs
```diff
@@ -0,0 +1,274 @@
...
+    let mut ir = [0u8; 32];
+    ir.clone_from_slice(&i[32..64]);
+
+    HdKeys {
+        key:        il,
+        chain_code: ir,
+    }
+}
+
+fn ckd_priv(keys: HdKeys, index: u32) -> HdKeys {
```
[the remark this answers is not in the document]
orhoj: I added a bit of documentation stating that it is the child key derivation.

mobile_wallet/mobile_wallet.h
```diff
@@ -165,6 +165,36 @@ char *generate_baker_keys(uint8_t *success);
...
+ * Take a pointer to a NUL-terminated UTF8-string and return a NUL-terminated
+ * UTF8-encoded string. The returned string must be freed by the caller by
+ * calling the function 'free_response_string'. In case of failure the function
+ * returns an error message as the response, and sets the 'success' flag to 0.
+ *
+ * See rust-bins/wallet-notes/README.md for the description of input and output
+ * formats.
+ *
+ * # Safety
+ * The input pointer must point to a null-terminated buffer, otherwise this
```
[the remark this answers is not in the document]
orhoj: All the other methods write it like this in the "Safety" section so I will keep it as is.

mobile_wallet/src/android.rs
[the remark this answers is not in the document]
orhoj: It is a return code, but this was just copy-pasted. This whole file could use a refactoring as they all have the same boilerplate code around a Rust method. We will keep it outside of the scope of this PR to fix that.

rust-bins/wallet-notes/README.md
```diff
@@ -444,6 +444,57 @@ The return value is a a JSON array with JSON objects as entries. Each object has
...
+
+- `"identityIndex"` ... the index of the identity to derive keys and randomness for, a u32 value
+
+- `"accountCredentialIndex"` ... the index of the account credential to derive keys and randomness for, a u32 value
+
+The returned value is a JSON object with the following fields:
+
+- `"signingKey"` ... the account signing key as a hex encoded string.
+
+- `"verifyKey"` ... the account verification key as a hex encoded string.
```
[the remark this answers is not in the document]
orhoj: Well... I was thinking that they probably got both the private and public key generated by our libraries currently, but I didn't actually check. My assumption was also that they just got `IdCredSec` generated and that our libraries automatically calculates `IdCredPub` from this when needed, but I didn't check this either. We should verify exactly what keys they store today so we can give them the same from this library.

I may not be able to find out today.

rust-src/key_derivation/src/lib.rs
orhoj: I have made the restriction that is must be 64 bytes.

orhoj: I think it would be fine to have a Debug implementation that also prints the seed. I suppose being in Debug mode is a controlled environment so I think it would be okay.

rust-src/key_derivation/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+    fn make_path(&self, path: &str) -> String {
+        let root_path: &str = "m/44'";
+        let derivation_path = format!("{}/{}'/{}", root_path, &self.net, path);
+        derivation_path
+    }
+
+    pub fn get_account_signing_key(&self, identity_index: u32, credential_index: u32) -> [u8; 32] {
+        let path = self.make_path(&format!("{}'/0'/{}'", identity_index, credential_index));
+        derive(&path, &self.seed).unwrap().key
+    }
```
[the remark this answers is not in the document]
orhoj: Haha, I get what you are saying. I haven't even considered that as most of these libraries take the path as a string as the input. The string format is just the readable version of a path, which also quickly tells you what indices are hardened and not. But for ed25519 only hardened indices are allowed, so we don't benefit that much from it.

We could change it so that the library takes a `Vec<u32>` instead of a string. This would also make it possible to get rid of the regex checking, as we wouldn't need it. Someone using the library who wants to use paths in that form would then write the converter instead. I won't be able to redo this today, but I wouldn't mind it being changed.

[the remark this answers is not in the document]
orhoj: You are correct that `unwrap` is not safe there. I am changing it all to `Result`.

rust-src/key_derivation/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+        let path = self.make_path(&format!("{}'/4'", identity_index));
+        let blinding_randomness_seed = derive(&path, &self.seed).unwrap().key;
+        bls_key_bytes_from_seed(blinding_randomness_seed)
+    }
+
+    pub fn get_attribute_commitment_randomness(
+        &self,
+        identity_index: u32,
+        credential_index: u32,
+        attribute_index: u32,
```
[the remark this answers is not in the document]
orhoj: It is just the `attribute_tag`. I have updated the type to be `AttributeTag`.

rust-src/key_derivation/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+        let derivation_path = format!("{}/{}'/{}", root_path, &self.net, path);
+        derivation_path
+    }
+
+    pub fn get_account_signing_key(&self, identity_index: u32, credential_index: u32) -> [u8; 32] {
+        let path = self.make_path(&format!("{}'/0'/{}'", identity_index, credential_index));
+        derive(&path, &self.seed).unwrap().key
+    }
+
+    pub fn get_account_public_key(&self, identity_index: u32, credential_index: u32) -> [u8; 32] {
```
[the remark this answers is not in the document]
orhoj: I have renamed it to `credential_counter`. In the desktop wallet we also use `credential_index` for what you state here, I just misremembered it.

rust-bins/wallet-notes/README.md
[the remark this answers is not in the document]
abizjak: Currently they have two calls
- the first `create_id_request_and_private_data` which returns both idCredSec and idCredPub and everything else (e.g., randomness for retrieving the signature)
- `create_credential` which returns the credential to send to the chain, as well as verifyKey and signKey (as well as commitments randomness for all the commitments that were made to the attributes in the credential.

rust-src/key_derivation/src/lib.rs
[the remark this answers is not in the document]
abizjak: I have added a function that allows you to supply a list of u32 values instead of a string, and am using that now in the library to derive prf, idcredsec, account keys, etc.

rust-src/key_derivation/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+        attribute_index: u32,
+    ) -> [u8; 32] {
+        let path = self.make_path(&format!(
+            "{}'/5'/{}'/{}'",
+            identity_index, credential_index, attribute_index
+        ));
+        let attribute_commitment_randomness_seed = derive(&path, &self.seed).unwrap().key;
+        bls_key_bytes_from_seed(attribute_commitment_randomness_seed)
+    }
+}
```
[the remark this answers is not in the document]
abizjak: I have added these comments now.

abizjak (approved):

rust-bins/wallet-notes/README.md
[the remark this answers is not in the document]
chrmatt: But then shouldn't `get_identity_keys_and_randomness` here also return `idCredPub` to be consistent?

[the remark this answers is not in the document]
abizjak: It's a bit complicated there, because `idCredPub = g^idCredSec` where `g` is the first part of the on-chain commitment key.
So that'd have to be passed into this function as well.

As I wrote above (see also #160 ) we need to rework a little bit the wallet API anyhow. I think for this function returning just idCredSec is OK.

PR #164, opened 2022-06-04
Speed up credential loading

abizjak:
## Purpose

Auxiliary changes to speed up node startup.

## Changes

The main change is to make the types for account credentials and account encryption keys less strict when these are used in trusted contexts (e.g., when the value has already been verified). These are the "Raw" account encryption keys and "Raw" credential registration IDs.

The reason this is beneficial is that a raw value is substantially (orders of magnitude) cheaper to deserialize, as well as taking up around 1/3 of the space (for these two cases).

See https://github.com/Concordium/concordium-node/pull/377 for more details as to why this makes sense.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/ID/Types.hs
+++ b/haskell-src/Concordium/ID/Types.hs
@@ -650,6 +650,9 @@ instance Serialize CredentialAccount where
         return $! NewAccount keys threshold
       _ -> fail "Input must be either a new account with a list of keys and threshold."
 
+-- |Values, as opposed to proofs, contained in a credential deployment. The type
+-- parameter @credTy@ determines the representation of credential registration
+-- IDs.
 data CredentialDeploymentValues' credTy = CredentialDeploymentValues {
   -- |Either an address of an existing account, or the list of keys the newly
   -- created account should have, together with a threshold for how many are needed
--- a/haskell-src/Concordium/ID/Types.hs
+++ b/haskell-src/Concordium/ID/Types.hs
@@ -287,6 +287,8 @@ newtype AccountEncryptionKey = AccountEncryptionKey {_elgamalPublicKey :: Elgama
 data AccountEncryptionKeySize
 
 instance FBS.FixedLength AccountEncryptionKeySize where
+  -- a group element is encoded as 48 bytes (BLS-G1 element), and an encryption
+  -- key is a pair of two
   fixedLength _ = 2 * 48
 
 -- |A byte representation of the account encryption key. This is both smaller
@@ -301,7 +303,8 @@ newtype RawAccountEncryptionKey = RawAccountEncryptionKey (FBS.FixedByteString A
 
 -- |Reconstruct the account encryption key from the raw representation. This
 -- function is unsafe and will raise an exception if the input is not a valid
--- account encryption key.
+-- account encryption key. Hence it should only be used on inputs that are known
+-- to be well-formed because of their provenance.
 unsafeEncryptionKeyFromRaw :: RawAccountEncryptionKey -> AccountEncryptionKey
 unsafeEncryptionKeyFromRaw (RawAccountEncryptionKey fbs) =
   case decode (FBS.toByteString fbs) of
@@ -802,14 +805,15 @@ instance Serialize InitialCredentialAccount where
     return InitialCredentialAccount{..}
 
 -- |The data for the initial account creation. This is submitted by the identity
--- provider on behalf of the account holder.
-data InitialCredentialDeploymentValues' ty = InitialCredentialDeploymentValues {
+-- provider on behalf of the account holder. The type parameter @credTy@
+-- determines the representation of credential registration IDs.
+data InitialCredentialDeploymentValues' credTy = InitialCredentialDeploymentValues {
   -- |List of keys the new account should have, together with a threshold
   -- for how many are needed. Its address is derived from the registration
   -- id of this credential.
   icdvAccount :: !CredentialPublicKeys,
   -- |Registration id of __this__ credential.
-  icdvRegId :: !ty,
+  icdvRegId :: !credTy,
   -- |Identity of the identity provider who signed this account creation
   icdvIpId :: !IdentityProviderIdentity,
   -- |Policy.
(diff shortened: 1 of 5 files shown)


haskell-src/Concordium/Wasm.hs
```diff
@@ -568,7 +568,7 @@ data SenderPolicy = SenderPolicy {
   spItems :: ![(AttributeTag, AttributeValue)]
   }
 
-mkSenderPolicy :: AccountCredential -> SenderPolicy
+mkSenderPolicy :: AccountCredential' credTy -> SenderPolicy
```
EmilLa1, line 571: Document `credTy`.
```suggestion
-- The type parameter @credTy@ determines the representation
-- of credential registration IDs.
mkSenderPolicy :: AccountCredential' credTy -> SenderPolicy
```

haskell-src/Concordium/ID/Types.hs
```diff
@@ -168,11 +171,11 @@ instance HashableTo SHA256.Hash AccountInformation where
 matchesAccountInformation :: AccountInformation -> SHA256.Hash -> Bool
 matchesAccountInformation ai h = getHash ai == h
 
-getCredentialPublicKeys :: AccountCredential -> CredentialPublicKeys
+getCredentialPublicKeys :: AccountCredential' credTy -> CredentialPublicKeys
```
EmilLa1, line 174: doc

haskell-src/Concordium/ID/Types.hs
```diff
@@ -610,13 +647,13 @@ instance Serialize CredentialAccount where
...
 
-data CredentialDeploymentValues = CredentialDeploymentValues {
+data CredentialDeploymentValues' credTy = CredentialDeploymentValues {
   -- |Either an address of an existing account, or the list of keys the newly
   -- created account should have, together with a threshold for how many are needed
   -- Its address is derived from the registration id of this credential.
   cdvPublicKeys :: !CredentialPublicKeys,
   -- |Registration id of __this__ credential.
-  cdvCredId     :: !CredentialRegistrationID,
+  cdvCredId     :: !credTy,
```
EmilLa1, line 656: update doc such that it now reflects the `credTy` parameter.

EmilLa1 (approved):

haskell-src/Concordium/ID/Types.hs
```diff
@@ -271,10 +274,45 @@ instance ToJSONKey IdentityProviderIdentity where
...
+-- to save some space.
 newtype AccountEncryptionKey = AccountEncryptionKey {_elgamalPublicKey :: ElgamalPublicKey}
     deriving (Eq, Show, Serialize, FromJSON, ToJSON) via ElgamalPublicKey
 
+-- |A marker type that has the 'FBS.FixedLength' instance to determine the size
+-- of the account encryption key in bytes.
+data AccountEncryptionKeySize
+
+instance FBS.FixedLength AccountEncryptionKeySize where
+  fixedLength _ = 2 * 48
```
mkmks, line 290: If this 48 number is the same 48 number as the credential ID size, make it explicitly so: 

```
fixedLength _ = 2 * fixedLength (undefined :: RegIdSize)
```

If it's not, please document where does it come from.

mkmks (approved):

haskell-src/Concordium/ID/Types.hs
```diff
@@ -271,10 +274,45 @@ instance ToJSONKey IdentityProviderIdentity where
...
+    deriving(Eq, Ord)
+    deriving (Show, Serialize, FromJSON, ToJSON) via FBSHex AccountEncryptionKeySize
+
+-- |Reconstruct the account encryption key from the raw representation. This
+-- function is unsafe and will raise an exception if the input is not a valid
+-- account encryption key.
+unsafeEncryptionKeyFromRaw :: RawAccountEncryptionKey -> AccountEncryptionKey
+unsafeEncryptionKeyFromRaw (RawAccountEncryptionKey fbs) =
+  case decode (FBS.toByteString fbs) of
+    Left _ -> error "Precondition violation. Invalid encryption key."
```
mkmks, line 308: It's not clear from this error message what to do to fix the underlying problem.

Perhaps, at this time this error message should at least be prepended with a caveat "not supposed to happen", so whoever sees it after the node crashes knows that it's not them who did something wrong but us.

haskell-src/Concordium/Wasm.hs
abizjak: THis is documented as part of the `AccountCredential` type already. I do not think it is necessary or practical to also document it at every occurrence.

haskell-src/Concordium/ID/Types.hs
```diff
@@ -168,11 +171,11 @@ instance HashableTo SHA256.Hash AccountInformation where
 matchesAccountInformation :: AccountInformation -> SHA256.Hash -> Bool
 matchesAccountInformation ai h = getHash ai == h
 
-getCredentialPublicKeys :: AccountCredential -> CredentialPublicKeys
+getCredentialPublicKeys :: AccountCredential' credTy -> CredentialPublicKeys
 getCredentialPublicKeys (InitialAC icdv) = icdvAccount icdv
 getCredentialPublicKeys (NormalAC cdv _) = cdvPublicKeys cdv
 
-getAccountInformation :: AccountThreshold -> Map.Map CredentialIndex AccountCredential -> AccountInformation
+getAccountInformation :: AccountThreshold -> Map.Map CredentialIndex (AccountCredential' credTy) -> AccountInformation
```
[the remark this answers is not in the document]
abizjak: As above, I do not believe it is practical or useful to document the meaning of the type variable at each function. It should be documented at the type definition, which it is.

haskell-src/Concordium/ID/Types.hs
```diff
@@ -763,25 +803,27 @@ instance Serialize InitialCredentialAccount where
...
 -- provider on behalf of the account holder.
-data InitialCredentialDeploymentValues = InitialCredentialDeploymentValues {
+data InitialCredentialDeploymentValues' ty = InitialCredentialDeploymentValues {
   -- |List of keys the new account should have, together with a threshold
   -- for how many are needed. Its address is derived from the registration
   -- id of this credential.
   icdvAccount :: !CredentialPublicKeys,
   -- |Registration id of __this__ credential.
-  icdvRegId :: !CredentialRegistrationID,
+  icdvRegId :: !ty,
```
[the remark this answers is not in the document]
abizjak: The meaning of the field is the same. I have added a comment explaining the meaning of the type variable to the type.

haskell-src/Concordium/ID/Types.hs
abizjak: It is not. I mean it is, but only because the same group is used for both. But conceptually they are not to be conflated

PR #167, opened 2022-06-10
Merge 4.1 bugfixes

### LINKED ISSUE #162: Possible overflow in leverage bound calculations
**Bug Description**

It is possible for leverage bound calculations to have an overflow if the baker's equity capital is high and/or the leverage factor has a large numerator.  This could cause the leverage bound to be calculated as lower than it should be.

**Steps to Reproduce**

Set the leverage bound to `(n*a + 1) / a`, where `a` is very large.
Create a baker with an open pool and significant capital.
Attempt to delegate to the pool with `n - 1` times the baker's capital.

**Expected Result**

Should succeed.

**Actual Result**

Will probably fail.

**Versions**
- Software Version 4.0.11

abizjak:
## Purpose

Merge release 4.1 bugfixes into main.

The changes here are those in #163 

Closes #162

### DIFF
--- a/haskell-src/Concordium/Types/Parameters.hs
+++ b/haskell-src/Concordium/Types/Parameters.hs
@@ -534,8 +534,13 @@ instance FromJSON LeverageFactor where
     return $ LeverageFactor r
 
 -- |Apply a leverage factor to a capital amount.
+-- If the computed amount would be larger than the maximum amount, this returns 'maxBound'.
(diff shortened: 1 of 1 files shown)


Bargsteen (approved):
I agree with @EmilLa1 regarding the tests.

abizjak:
I agree about tests in general, but I'm not so sure about this specific function which is very declarative and a test would involve essentially the same implementation of it.

PR #169, opened 2022-06-17
Remove transaction logging

abizjak:
## Purpose

Part of https://github.com/Concordium/concordium-client/pull/160

## Changes

Remove the postgres and database specific parts from the base. This removes the postgres dependency from base.
These have been moved to the wallet-proxy where they are needed.

## Checklist

      hard-to-understand areas.


haskell-src/Concordium/ID/Types.hs
```diff
@@ -1058,7 +1058,7 @@ values (NormalACWP cdi) = NormalAC (cdiValues cdi) <$> proofCommitments (cdiProo
 class HasCredentialType a where
   credentialType :: a -> CredentialType
 
-instance HasCredentialType AccountCredential where
```
limemloh, line 1061: I don't see why this change is relevant for the transaction logging, but I see no problem in it

limemloh (approved):

td202 (approved):

haskell-src/Concordium/ID/Types.hs
abizjak: It is not relevant.

But I needed it when I was updating the dependency in concordium-client (so I could update it in the wallet-proxy).

It is a leftover from the "improve startup speed" PR and since it's so trivial I did not want to make a separate PR with it.

PR #207, opened 2022-08-19
Contract error schemas

Bargsteen:
## Purpose

Part of https://github.com/Concordium/concordium-rust-smart-contracts

## Changes

- Added function for generating schemas to use v2 schemas

## Checklist

      hard-to-understand areas.

### DIFF
--- a/smart-contracts/wasm-chain-integration/src/utils.rs
+++ b/smart-contracts/wasm-chain-integration/src/utils.rs
@@ -252,7 +252,7 @@ pub fn generate_contract_schema_v0(
 }
 
 /// Tries to generate schemas for parameters and return values of methods for a
-/// V1 contract.
+/// contract with a V1 schema.
 pub fn generate_contract_schema_v1(
     module_bytes: &[u8],
 ) -> ExecResult<schema::VersionedModuleSchema> {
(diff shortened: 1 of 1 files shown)


abizjak (commented):
This branch was out of date with `main`, so it had some reformatting changes Emil introduced but were reverted. Not sure how you got this history.

I reverted that commit now, but it would probably be good to clean up the history of this branch to just have the commit  that changes `generate_contract_schema_v1` -> `v2`.

smart-contracts/wasm-chain-integration/src/utils.rs
```diff
@@ -252,8 +252,8 @@ pub fn generate_contract_schema_v0(
 }
 
 /// Tries to generate schemas for parameters and return values of methods for a
-/// V1 contract.
-pub fn generate_contract_schema_v1(
+/// V1 contract with a v2 schema.
+pub fn generate_contract_schema_v2(
```
abizjak, line 256: I suppose we could have both `V1` and `V2` functions here, since they are both in principle usable.

I guess concordium-std version ultimately determines which version is the correct one. Maybe we should document some compatibilities between cargo-concordium and concordium-std.

Bargsteen: Where do you propose we document the compatibility? Developer docs?

abizjak: I was thinking just for us in the README to cargo-concordium.

abizjak (approved):

PR #208, opened 2022-08-29
Add identity provider index to key derivation paths

orhoj:
## Purpose
Update the `key_derivation` library to pre-fix paths with the `identity provider index`. The purpose of doing this is to ensure that keys will not be used multiple times across different identity providers, i.e. the keys are isolated per identity provider.

## Changes
- Updated `key_derivation` methods to require the index of an identity provider.
- Added official test vectors for `words_to_seed` function in `key_derivation`.
- Updated rust-bins to work with the new library.
- Updated mobile_wallet to work with the new library.

## Checklist

      hard-to-understand areas.


abizjak (commented):
I guess the test vectors are compared to what the ledger/web wallet does?

We need to update the version of both `user_cli` and the `client` tool. The `user_cli` is versioned in `user_cli.rs` using `structopt`/`clap` versioning support. Can you please do that?

orhoj:
> I guess the test vectors are compared to what the ledger/web wallet does?
> 
> We need to update the version of both `user_cli` and the `client` tool. The `user_cli` is versioned in `user_cli.rs` using `structopt`/`clap` versioning support. Can you please do that?

The test vectors are generated from a Typescript implementation I have that does the same thing. The plan is to have the same tests for the Ledger when we implement it there, but we haven't yet. I should probably add those tests before we release the mobile wallets / browser wallet to ensure that everything is consistent.

I will bump the versions a minor.

eb-concordium (approved):
Looks good.

chrmatt (approved):
Looks good.

abizjak (approved):

PR #211, opened 2022-09-06
Document recovery endpoint.

eb-concordium:
## Purpose

To document the recovery endpoint of the internal identity provider.

## Changes

* README updated.

### DIFF
--- a/identity-provider-service/README.md
+++ b/identity-provider-service/README.md
@@ -283,7 +283,7 @@ containing valid JSON of the form
 }
 ```
 
-In case of success (if the request is valid), the response will be of the form
+In case of success (if the request is valid), the response is HTTP 200 with a body of the form
 ```json
 {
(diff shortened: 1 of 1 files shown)


identity-provider-service/README.md
```diff
@@ -265,3 +266,43 @@ being processed, then an error can be returned:
...
+    "message": "Invalid timestamp."
+}
+```
+
+Otherwise, if the timestamp is OK, but the request is invalid due to an invalid proof, the response will be
+```json
+{
+    "code": 400,
+    "message": "Invalid ID recovery proof."
+}
```
abizjak, line 307: Can you specify what the HTTP status code will be as well? It should be 400 - Bad Request.

abizjak (approved):
Please elaborate the status code. Otherwise looks good.

PR #221, opened 2022-09-19
Add ability to construct contract state from iterators and streams.

abizjak:
## Purpose

Make it easier to use the contract state in the rust sdk.

## Changes

Expose contract state construction methods, from iterators, and from streams.
The latter is useful in combination with the grpc API since the return value there is naturally a stream. This allows us to not construct an intermediate representation in memory.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/smart-contracts/wasm-chain-integration/src/v1/trie/api.rs
+++ b/smart-contracts/wasm-chain-integration/src/v1/trie/api.rs
@@ -250,6 +250,9 @@ impl PersistentState {
     /// If the stream yields duplicate keys then values at later keys
     /// override earlier ones. The resulting state lies entirely in memory
     /// and so can be used with any [`Loader`].
+    ///
+    /// This function should be preferred over
+    /// [`try_from_stream`](Self::try_from_stream) when it is applicable.
     pub async fn from_stream<S>(s: S) -> Self
     where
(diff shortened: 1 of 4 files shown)


smart-contracts/wasm-chain-integration/src/v1/trie/api.rs
```diff
@@ -199,6 +199,68 @@ impl PersistentState {
...
+            None => Self::Empty,
+        })
+    }
+
+    #[cfg(feature = "async")]
+    /// Construct a [PersistentState] from a stream of key-value pairs.
+    /// If the stream yields duplicate keys then values at later keys
+    /// override earlier ones. The resulting state lies entirely in memory
+    /// and so can be used with any [`Loader`].
+    pub async fn from_stream<S>(s: S) -> Self
```
Bargsteen, line 253: I think we should refer to `try_from_stream` in the doc comment and explain when `from_stream` should be preferred.

Bargsteen (approved):
LGTM :+1:

smart-contracts/wasm-chain-integration/src/v1/trie/api.rs
abizjak: I have added this.

PR #223, opened 2022-09-24
Add documentation to `resume_receive`, and fix a bug.

abizjak:
## Purpose

Fix bug in `resume_receive`, and clarify the usage.
This bug is not an issue on the node since the value that is supplied is the same for both state_changed flags. But if this function is used independently then it is not correct, and in any case it does not conform to the intent.

## Changes

- Add documentation
- Use correct `state_changed` flag when informing the instance.

## Checklist

      hard-to-understand areas.


EmilLa1 (approved):
Great doc!

limemloh (approved):

PR #224, opened 2022-09-26
Add ability to provide custom JSON parser when decrypting.

abizjak:
## Purpose

In support of https://github.com/Concordium/concordium-client/issues/170

## Changes

- Add a JSON parser for encrypted data that takes a custom parser, not relying on the FromJSON instance directly.
- Add a helper "isLikelyEncrypted" method.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Utils/Encryption.hs
+++ b/haskell-src/Concordium/Utils/Encryption.hs
@@ -9,6 +9,10 @@ import Control.Monad.Except
 import Control.Exception
 
 import qualified Data.Aeson as AE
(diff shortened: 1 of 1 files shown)


EmilLa1 (approved):
So I suspect the custom parser will be put in concordium client?  Perhaps it should also be part of base and the parser will be rather "standard" wrt. the supported export formats we use.

abizjak:
Yes, the parser will be put in concordium-client, for a type defined in the client. It's not that standard due to legacy issues.

eb-concordium (approved):

PR #233, opened 2022-10-10
Introduce AccountV2 and protocol version P5

td202:
## Purpose

This introduces a new account version, `AccountV2`, and a new protocol version `P5` that uses this account version. In other respects, `P5` is the same as `P4`, and `AccountV2` is the same as `AccountV1`: this is just setting the groundwork for the change.

## Changes

- Protocol version `P5` is introduced with genesis data version tag 7. The genesis data is the same as for `P4`.
- Introduce a new account version `AccountV2`.
- Introduce a new constraint `AVSupportsDelegation` for account versions that support delegation, which is satisfied by both `AccountV1` and `AccountV2`.
  - `AVSupportsDelegation av` is implemented as the constraint `AVSupportsDelegationB av ~ 'DelegationSupported av`, where `AVSupportsDelegationB` is a type family (i.e. type-level function) defined as:
    ```haskell
    -- |Type-level predicate that determines if an account version supports delegation.
    type family AVSupportsDelegationB (av :: AccountVersion) :: DelegationSupport where
        AVSupportsDelegationB 'AccountV0 = 'DelegationNotSupported 'AccountV0
        AVSupportsDelegationB av = 'DelegationSupported av
    ```
  - As a consequence, in contexts where `AVSupportsDelegation av` is required but not inferred, the type checker will generate errors that `AVSupportsDelegationB av` does not match `'DelegationSupported av`. Or, in the concrete case of `AccountV0`, that `'DelegationNotSupported 'AccountV0` does not match `'DelegationSupported 'AccountV0`.
  - Definitions which previously used `AccountV1` are generalized to any account version that satisfies `AVSupportsDelegation`. (Hashing is an exception to this rule.)
  - The constraint `SupportsDelegation pv` is (re)defined as a synonym for `AVSupportsDelegation (AccountVersionFor pv)`.
- Introduce a GADT `SAVDelegationSupport` that either witnesses that the account version is `AccountV0` (for `SAVDelegationNotSupported`) or that the account version satisfies `AVSupportsDelegation` (for `SAVDelegationSupported`).
  ```haskell
  data SAVDelegationSupport (av :: AccountVersion) where
      SAVDelegationNotSupported :: SAVDelegationSupport 'AccountV0
      SAVDelegationSupported :: AVSupportsDelegation av => SAVDelegationSupport av
  ```
  - The function `delegationSupport :: forall av. (IsAccountVersion av) => SAVDelegationSupport av` provides a way to obtain this for an account version.  Code that branches on whether delegation is supported should typically be implemented by pattern matching on `delegationSupport @av`, where previously pattern matching on `accountVersion @av` was used.
- As it is sometimes useful to know the chain parameters version that is implied by knowing the protocol version satisfies `SupportsDelegation`, we introduce `delegationChainParameters` (and the associated GADT `DelegationChainParameters`) that can be used to constrain the chain parameters version by pattern matching on the result.
- `[get|put]Account[Baker|Stake]` renamed to `[de|]serializeAccount[Baker|Stake]`. This is to reduce ambiguity with similarly-named functions in the node.
- In some modules redundant constraint warnings have been suppressed. This is because the compiler flags the warning if the constraint is not necessary for type checking, but may still be used for coverage checking in pattern matches. (For instance, in the context `AVSupportsDelegation av`, pattern matching on `accountVersion @av` can omit the case for `SAccountV0` without warning of an incomplete pattern match.)

## Checklist

      hard-to-understand areas.


haskell-src/Concordium/Genesis/Data/P5.hs
```diff
@@ -0,0 +1,189 @@
...
+
+-- |The regenesis represents a reset of the protocol with a new genesis block.
+--  This does not include the full new state, but only its hash.
+--
+-- The relationship between the new state and the state of the
+-- terminal block of the old chain should be defined by the
+-- chain update mechanism used.
+--
+-- There are two variants, one when migrating from P3, and another one for an
+-- update from P5 to P5.
```
abizjak, line 35: This does not seem accurate.

haskell-src/Concordium/Types/Accounts.hs
```diff
@@ -262,6 +279,10 @@ instance Serialize effectiveTime => Serialize (StakePendingChange' effectiveTime
 
 type StakePendingChange (av :: AccountVersion) = StakePendingChange' (PendingChangeEffective av)
 
+-- |Coerce a 'StakePendingChange' to use 'Timestamp' for the effective time.
+stakePendingChangeTimestamp :: forall av. (AVSupportsDelegation av, IsAccountVersion av) => StakePendingChange av -> StakePendingChange' Timestamp
```
limemloh, line 283: Consider another name for this function, currently it is just the return type.

I had to look up the function a few times to remind me of what it actually does, maybe add `coerce` or something.

EmilLa1 (approved):
LGTM

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -200,6 +221,78 @@ instance IsAccountVersion 'AccountV0 where
 instance IsAccountVersion 'AccountV1 where
     accountVersion = SAccountV1
 
+instance IsAccountVersion 'AccountV2 where
+    accountVersion = SAccountV2
+
+type family AVNat (av :: AccountVersion) where
```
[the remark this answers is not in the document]
td202: Will remove this and also `PVNat` which are unused.

haskell-src/Concordium/Types/Accounts.hs
td202: I cannot think of a very good name for this function, so I will replace it with the definition, since it is so simple.

PR #235, opened 2022-10-10
Add schema version 3

DOBEN:
## Purpose

Schemas for events

## Changes

Add support for schema version 3

## Checklist

      hard-to-understand areas.

### DIFF
--- a/smart-contracts/wasm-chain-integration/src/utils.rs
+++ b/smart-contracts/wasm-chain-integration/src/utils.rs
@@ -356,9 +356,7 @@ pub fn generate_contract_schema_v3(
                     .or_insert_with(schema::ContractV3::default);
                 contract_schema.event = function_schema_event;
             }
-        }
-
-        if let Some(rest) = name.as_ref().strip_prefix("concordium_schema_function_") {
+        } else if let Some(rest) = name.as_ref().strip_prefix("concordium_schema_function_") {
             if let Some(contract_name) = rest.strip_prefix("init_") {
                 let function_schema = generate_schema_run(&artifact, name.as_ref())?;
 
@@ -380,8 +378,8 @@ pub fn generate_contract_schema_v3(
 
                 contract_schema.receive.insert(function_name.to_owned(), function_schema);
             } else {
-                // do nothing, some other function that is neither init nor
-                // receive.
+                // do nothing: no event schema and not a schema that was attached to an init/
+                // receive function
             }
         }
     }
--- a/smart-contracts/wasm-chain-integration/src/utils.rs
+++ b/smart-contracts/wasm-chain-integration/src/utils.rs
@@ -378,8 +378,8 @@ pub fn generate_contract_schema_v3(
 
                 contract_schema.receive.insert(function_name.to_owned(), function_schema);
             } else {
-                // do nothing: no event schema and not a schema that was attached to an init/
-                // receive function
+                // do nothing: no event schema and not a schema that was
+                // attached to an init/ receive function
             }
         }
     }
--- a/smart-contracts/wasm-chain-integration/src/utils.rs
+++ b/smart-contracts/wasm-chain-integration/src/utils.rs
@@ -355,7 +355,7 @@ pub fn generate_contract_schema_v3(
                 let contract_schema = contract_schemas
                     .entry(contract_name.to_owned())
                     .or_insert_with(schema::ContractV3::default);
-                contract_schema.event = function_schema_event;
+                contract_schema.event = Some(function_schema_event);
             }
         } else if let Some(rest) = name.as_ref().strip_prefix("concordium_schema_function_") {
             if let Some(contract_name) = rest.strip_prefix("init_") {
--- a/smart-contracts/wasm-chain-integration/src/utils.rs
+++ b/smart-contracts/wasm-chain-integration/src/utils.rs
@@ -347,6 +347,17 @@ pub fn generate_contract_schema_v3(
     let mut contract_schemas = BTreeMap::new();
 
     for name in artifact.export.keys() {
+        if let Some(rest) = name.as_ref().strip_prefix("concordium_event_schema_") {
+            if let Some(contract_name) = rest.strip_prefix("init_") {
+                let function_schema_event = generate_schema_run(&artifact, name.as_ref())?;
+
+                let contract_schema = contract_schemas
+                    .entry(contract_name.to_owned())
+                    .or_insert_with(schema::ContractV3::default);
+                contract_schema.event = function_schema_event;
+            }
+        }
+
         if let Some(rest) = name.as_ref().strip_prefix("concordium_schema_function_") {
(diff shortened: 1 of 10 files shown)


abizjak (commented):
This looks good, but the dependencies might change so I am holding off approval until then.

abizjak (approved):
Good job.

abizjak:
Actually, @DOBEN can you add "support for V3 schema" to the changelog of the mobile wallet library? And bump the version?

smart-contracts/wasm-chain-integration/src/utils.rs
```diff
@@ -337,6 +337,60 @@ pub fn generate_contract_schema_v2(
...
+
+    for name in artifact.export.keys() {
+        if let Some(rest) = name.as_ref().strip_prefix("concordium_event_schema_") {
+            if let Some(contract_name) = rest.strip_prefix("init_") {
+                let function_schema_event = generate_schema_run(&artifact, name.as_ref())?;
+
+                let contract_schema = contract_schemas
+                    .entry(contract_name.to_owned())
+                    .or_insert_with(schema::ContractV3::default);
+                contract_schema.event = function_schema_event;
```
abizjak, line 357: I'm a bit confused why this typechecks. Isn't `contract_schema.event: Option<Type>`?

So I would have expected you need to wrap this in a Some, unless your `concordium_event_schema` function serializes a `Some` which I do not see that it does.

So I kind of expect
```rust
contract_schema.event = Some(function_schema_event);
```

Am I misunderstanding something?

smart-contracts/wasm-chain-integration/src/utils.rs
```diff
@@ -337,6 +337,60 @@ pub fn generate_contract_schema_v2(
...
+            }
+        }
+
+        if let Some(rest) = name.as_ref().strip_prefix("concordium_schema_function_") {
```
abizjak, line 361:
```suggestion
        } else if let Some(rest) = name.as_ref().strip_prefix("concordium_schema_function_") {
```

smart-contracts/wasm-chain-integration/src/utils.rs
```diff
@@ -337,6 +337,60 @@ pub fn generate_contract_schema_v2(
...
+                let function_name = split_name[1];
+
+                let contract_schema = contract_schemas
+                    .entry(contract_name.to_owned())
+                    .or_insert_with(schema::ContractV3::default);
+
+                contract_schema.receive.insert(function_name.to_owned(), function_schema);
+            } else {
+                // do nothing, some other function that is neither init nor
+                // receive.
```
abizjak, line 384: Suggest updating this comment to reflect the additional choice.

smart-contracts/wasm-chain-integration/src/utils.rs
DOBEN: I think that event is already in the format `Some(event_schema)` from this part here:

https://github.com/Concordium/concordium-rust-smart-contracts/blob/0e1331af82938f02d489db87659f782f87fb7235/concordium-std-derive/src/lib.rs#L1075

abizjak: I see, you are serializing an Option<Type> there already. This is a bit different from init I think. An alternative would be
```rust

#[cfg(feature = "build-schema")]
fn contract_function_schema_tokens2(
    event_option: Option<syn::LitStr>,
    contract_name: String,
) -> syn::Result<proc_macro2::TokenStream> {
    let mut embed = false;

    let event_schema = format_ident!("event");
    let event_embed = if let Some(event_ty) = event_option {
        let ty = event_ty.parse::<syn::Type>()?;
        let schema_name = format!("concordium_event_schema_{}", contract_name);
        let schema_ident = format_ident!("concordium_event_schema_{}", contract_name);
        quote! {
             let #event_schema = Some(<#ty as schema::SchemaType>::get_type());
             let schema_bytes = concordium_std::to_bytes(&#event_schema);
             concordium_std::put_in_memory(&schema_bytes)
        }
    } else {
        Ok(proc_macro2::TokenStream::new())
    }
```

and then here you know that if the function exists, then the event schema will be there, so you would  wrap it into Some.

DOBEN: updated this pull request and the `concordium-rust-smart-contracts` one.

abizjak: Great, thank you.

smart-contracts/wasm-chain-integration/src/utils.rs
```diff
@@ -337,6 +337,60 @@ pub fn generate_contract_schema_v2(
...
+pub fn generate_contract_schema_v3(
+    module_bytes: &[u8],
+) -> ExecResult<schema::VersionedModuleSchema> {
+    let artifact = utils::instantiate::<ArtifactNamedImport, _>(&TestHost, module_bytes)?;
+
+    let mut contract_schemas = BTreeMap::new();
+
+    for name in artifact.export.keys() {
+        if let Some(rest) = name.as_ref().strip_prefix("concordium_event_schema_") {
+            if let Some(contract_name) = rest.strip_prefix("init_") {
```
abizjak, line 351: Am I understanding correctly that for a contract `foo`, the exported function will be
```
concordium_event_schema_init_foo
```

I am guessing that the `init` is not necessary. I don't think it is necessary to change this, but I would like to know whether it is necessary for some reason or it was just a choice.

smart-contracts/wasm-chain-integration/src/utils.rs
```diff
@@ -337,6 +337,60 @@ pub fn generate_contract_schema_v2(
...
+                // Generate event schema
+                let function_schema_event = generate_schema_run(&artifact, name.as_ref())?;
+
+                let contract_schema = contract_schemas
+                    .entry(contract_name.to_owned())
+                    .or_insert_with(schema::ContractV3::default);
+                contract_schema.event = Some(function_schema_event);
+            }
```
abizjak, line 359: Perhaps add an explicit comment here as to why we do nothing in the else branch. Like we do below for the other part of the schema.

abizjak (approved):

smart-contracts/wasm-chain-integration/src/utils.rs
DOBEN: The `init` part in the name is technically not necessary but since the event schema is stated as part of the init function definition, it seems nice to include `init` here so it is consistent with how the other schemas are named. I kept `init` in the name for now.

abizjak: Ok. I think technically the reason for this for the parameter and return value is that this is the way to distinguish them from the receive ones.

But since you already have gteh `concordium_event_schema` as a prefix, as opposed to `concordium_function_schema`, it would be unnecerssary.

But this is fine. Good job.

PR #237, opened 2022-10-12
Host function for upgrading a contract

EmilLa1:
## Purpose

Support for smart contract upgrading. 

This PR contains the following:

Haskell base:
- 'Upgraded' event.

wasm-chain-integration: 
- A new host function `upgrade`. 
Upon invocation the execution of the smart contract will signal an `Interrupt` (just like for `call` and `transfer`) with a new code: 2. This indicates to the `Scheduler` that the contract wishes to perform an `upgrade`.  
- `validate_and_process_v1` now takes in an additional parameter to specify whether we should support the upgrade or not.
This is used to deduce whether the module to validate and process are allowed to import `upgrade`.

## Checklist

      hard-to-understand areas.


haskell-src/Concordium/Types/Execution.hs
```diff
@@ -1967,6 +1967,13 @@ data RejectReason = ModuleNotWF -- ^Error raised when validating the Wasm module
                   | PoolWouldBecomeOverDelegated
                   -- |The pool is not open to delegators.
                   | PoolClosed
+                  -- |The module reference provided for the contract upgrade did not
+                  -- point to a deployed module.
+                  | UpgradeInvalidModuleReference !ModuleRef
```
limemloh, line 1972: I don't think these will be used, at least not until we have stack traces for failed executions.
The reason being that when a contract call `upgrade` it will be told whether it failed and should decide how to deal with the failure. In the case it rejects the transaction, it will be a `RejectedInit` or `RejectedReceive`.

smart-contracts/wasm-chain-integration/src/v1/mod.rs
```diff
@@ -25,6 +25,11 @@ use wasm_transform::{
     utils,
 };
 
+// TOOD: Move these to concordium-contracts-common?
```
limemloh, line 28: I have added a module reference in concordium-contracts-common since I also needs it.
Will make a PR soon

limemloh: https://github.com/Concordium/concordium-contracts-common/pull/55

haskell-src/Concordium/Types/Execution.hs
abizjak: That is correct. We will need this in the internals, i.e., signalling from the scheduler back to the contract, but we won't record it in the final transaction event.

smart-contracts/wasm-chain-integration/src/v1/mod.rs
```diff
@@ -309,6 +325,20 @@ mod host {
...
+                    length
+                );
+                ensure!(start + length <= memory.len(), "Illegal memory access.");
+                let mut module_ref_bytes = [0u8; MODULE_REF_SIZE];
+                module_ref_bytes.copy_from_slice(&memory[start..start + MODULE_REF_SIZE]);
+                let module_ref = ModuleRef(module_ref_bytes);
+                Ok(Some(Interrupt::Upgrade {
+                    module_ref,
+                }))
+            }
```
abizjak, line 341: The `upgrade` should not be exposed as an operation in this way. The reason for this is that this way there is no practical way of checking "is it upgradable". We need to add a new host function.

However internally we can handle it as an operation like you did (internally being in the scheduler and FFI).

smart-contracts/wasm-chain-integration/src/v1/mod.rs
```diff
@@ -760,6 +774,29 @@ mod host {
...
+        memory: &mut Vec<u8>,
+        stack: &mut machine::RuntimeStack,
+        energy: &mut InterpreterEnergy,
+    ) -> machine::RunResult<Option<Interrupt>> {
+        let module_ref_start = unsafe { stack.pop_u32() } as usize;
+        let module_ref_end = module_ref_start + MODULE_REFERENCE_SIZE as usize;
+        ensure!(module_ref_end <= memory.len(), "Illegal memory access.");
+        let mut module_reference_bytes = [0u8; MODULE_REFERENCE_SIZE];
+        module_reference_bytes
+            .copy_from_slice(&memory[module_ref_start..module_ref_start + MODULE_REFERENCE_SIZE]);
```
abizjak, line 790:
```suggestion
            .copy_from_slice(&memory[module_ref_start..module_ref_end]);
```

smart-contracts/wasm-chain-integration/src/v1/types.rs
```diff
@@ -445,6 +446,7 @@ impl Output for ImportFunc {
                 ReceiveOnlyFunc::GetReceiveEntrypointSize => 29,
                 ReceiveOnlyFunc::GetReceiveEntryPoint => 30,
                 ReceiveOnlyFunc::Invoke => 31,
+                ReceiveOnlyFunc::Upgrade => 32,
```
abizjak, line 449: I don't think you can use 32, it is already used for 
```
ImportFunc::Common(CommonFunc::VerifyEd25519)
```

Also you need to add this tag to the
```
    fn parse(
        ctx: Ctx,
        cursor: &mut std::io::Cursor<&'a [u8]>,
    ) -> wasm_transform::parse::ParseResult<Self> {
```
method.

smart-contracts/wasm-transform/src/utils.rs
```diff
@@ -19,6 +19,8 @@ pub fn instantiate<I: TryFromImport, VI: ValidateImportExport>(
 }
 
 /// Parse, validate, inject metering, and compile to a runnable artifact.
+/// Returning the runnable artifact and a bool indicating whether the
```
EmilLa1, line 22: Remove false comment.

smart-contracts/wasm-chain-integration/src/v1/ffi.rs
```diff
@@ -14,6 +14,8 @@
 //! In addition to pointers to structured objects, the remaining data passed
 //! between foreign code and Rust is mainly byte-arrays. The main reason for
 //! this is that this is cheap and relatively easy to do.
+use std::convert::TryInto;
+
```
abizjak, line 18: remove empty line

haskell-src/Concordium/Types/Execution.hs
```diff
@@ -1967,6 +1991,13 @@ data RejectReason = ModuleNotWF -- ^Error raised when validating the Wasm module
...
+                  -- |The module for upgrading to did not contain a contract name of the one upgrading from.
+                  | UpgradeInvalidContractName !ModuleRef !Wasm.InitName
+                  -- |The module for upgrading to is not a supported smart contract version.
+                  | UpgradeInvalidVersion !ModuleRef !Wasm.WasmVersion
```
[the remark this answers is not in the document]
EmilLa1: They are used internally by the `Scheduler` for signaling failing scenarios during transaction execution. But yes, these will not be part of the emitted transaction event as `upgrades` are always occurs within a nested execution (i.e. upon an `Interrupt` signal.) 

The reason is that the `RejectReason` type does not distinguish whether the reject happens within a 'nested environment' or from a top level transaction (transfers, stake changes, baker transactions etc.) 

I don't see any value adding this distinction now. 
So no, why they are part of the API then they are never actually occurring in it. But that might change in the future if we want to expose stacktraces etc.

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -302,4 +302,14 @@ supportsV1Contracts SP1 = False
 supportsV1Contracts SP2 = False
 supportsV1Contracts SP3 = False
 supportsV1Contracts SP4 = True
-supportsV1Contracts SP5 = True
\ No newline at end of file
+supportsV1Contracts SP5 = True
+
+-- |Whether the protocol version supports upgradable smart contracts.
+-- (Supported in 'P5' onwards)
```
Bargsteen, line 308:
```suggestion
-- (Supported in 'P5' and onwards)
```

smart-contracts/wasm-chain-integration/src/v1/mod.rs
```diff
@@ -39,8 +40,14 @@ pub enum Interrupt {
         name:      OwnedEntrypointName,
         amount:    Amount,
     },
+    Upgrade {
+        module_ref: ModuleReference,
+    },
 }
 
+// TODO: Move to contracts common
+pub const MODULE_REFERENCE_SIZE: usize = 32;
```
abizjak, line 49: Figure out.

haskell-src/Concordium/Types/Execution.hs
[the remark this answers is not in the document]
abizjak: I have remove these from the external API now.

abizjak (approved):

haskell-src/Concordium/Types/Execution.hs
```diff
@@ -1605,6 +1624,11 @@ instance AE.ToJSON Event where
         "delegatorId" .= edrDelegatorId,
         "account" .= edrAccount
       ]
+    Upgraded {..} -> AE.object [
+        "tag" .= AE.String "Upgraded",
```
limemloh, line 1628: This does not match the `FromJson` instance.
The `euAddress` seems to be missing

limemloh (changes requested):

Bargsteen (approved):

PR #238, opened 2022-10-13
Extend mobile library

orhoj:
## Purpose
Extend the mobile library for the WalletConnect use cases:
- Sign account transactions
- Sign messages
- Transform a serialized transaction into JSON, which can be used to display the transaction

## Changes
- Added sign transaction function
- Added sign message function
- Added transaction to JSON function
- Added missing documentation example files

## Checklist

      hard-to-understand areas.


mobile_wallet/src/lib.rs
```diff
@@ -88,6 +97,52 @@ fn make_signatures(
...
+fn sign_message_with_keys(
+    keys: &AccountKeys,
+    msg: String,
+    account_address: [u8; 32],
+) -> BTreeMap<CredentialIndex, BTreeMap<KeyIndex, Signature>> {
+    let message_digest = get_message_digest(account_address, msg);
+
+    // TODO: The code below has been copied from sign_transaction_hash(), where to
+    // put a shared function to avoid duplication?
+    let iter = keys
```
orhoj, line 124: This is copy-pasted, but I was unsure where to put the code to share it. Any pointers?

abizjak: Can you not use that function directly?

changing
```
fn get_message_digest(account_address: [u8; 32], message: String) -> hashes::TransactionSignHash {
```
and then 
```
let hash_to_sign = get_message_digest(...);
keys.sign(&hash_to_sign)
```
Maybe I am missing something that is blocking this?

mobile_wallet/src/lib.rs
```diff
@@ -163,6 +218,182 @@ fn create_encrypted_transfer_aux(input: &str) -> anyhow::Result<String> {
     Ok(to_string(&response)?)
 }
 
+pub fn replace_json_value(v: &Value, replacement_key: &String, replacement: &Value) -> Value {
```
abizjak, line 221: This is rather an odd function. Can you provide context for why it is needed?

mobile_wallet/src/lib.rs
```diff
@@ -163,6 +218,182 @@ fn create_encrypted_transfer_aux(input: &str) -> anyhow::Result<String> {
...
+        Payload::Update { payload } => {
+            let schema: String = try_get(&v, "schema")?;
+            let schema_version: Option<u8> = maybe_get(&v, "schemaVersion")?;
+            let parameters_as_json = get_parameters_as_json(payload, &schema, &schema_version)?;
+            let transaction_as_json = serde_json::json!(pre_tx);
+            replace_json_value(
+                &transaction_as_json,
+                &String::from("message"),
+                &parameters_as_json,
+            )
```
abizjak, line 359: This seems a bit hacky and not in the spirit of the other code we have. Is it possible to instead return 
```
json!({
    "hashToSign": pre_tx.hash_to_sign,
    "header: pre_tx.header,
    "payload": {
        "update": {
        "amount": payload.amount,
        "address": payload.address,
        "receiveName": payload.receive_name,
        "message": parameters_as_json
       }
)
```
instead?

mobile_wallet/src/lib.rs
orhoj: You are right. It does become:
```
keys.sign_transaction_hash(&hash_to_sign)
```
which is a little awkward as it is not a transaction hash, but I think it is fine.

mobile_wallet/src/lib.rs
orhoj: It was just to replace a JSON element in a JSON object. I will make the explicit JSON object as per your suggestion.

mobile_wallet/src/lib.rs
orhoj: I have changed it to follow your suggestion.

mobile_wallet/src/lib.rs
```diff
@@ -163,6 +194,166 @@ fn create_encrypted_transfer_aux(input: &str) -> anyhow::Result<String> {
...
+        Ok(versioned) => get_receive_schema(versioned, contract_name, entrypoint_name)?,
+        Err(_) => match schema_version {
+            Some(version) => {
+                let mut versioned_schema_bytes: Vec<u8> = vec![u8::MAX, u8::MAX, *version];
+                versioned_schema_bytes.append(&mut schema_bytes);
+                let versioned_schema: VersionedModuleSchema =
+                    match from_bytes(&versioned_schema_bytes) {
+                        Ok(v) => v,
+                        Err(_) => return Err(anyhow::anyhow!("Invalid schema or schema version")),
+                    };
```
abizjak, line 279: Instead of this we could also do
```
match schema_version {
    0 => Ok(VersionedModuleSchema::V0(from_bytes(&schema_bytes)?),
    1 => Ok(...)
    2 => Ok(...)
}
```

Which at least looks a bit less hacky. If you don't think this is better, then please add documentation to the current approach of prepending these bits.

mobile_wallet/src/lib.rs
```diff
@@ -163,6 +194,166 @@ fn create_encrypted_transfer_aux(input: &str) -> anyhow::Result<String> {
...
+                get_receive_schema(versioned_schema, contract_name, entrypoint_name)?
+            }
+            None => return Err(anyhow::anyhow!("Missing schema version")),
+        },
+    };
+
+    let parameter_bytes: Vec<u8> = payload.message.clone().into();
+    let mut parameters_cursor = Cursor::new(&parameter_bytes[..]);
```
abizjak, line 287:
```suggestion
    let parameter_bytes = payload.message.as_ref();
    let mut parameters_cursor = Cursor::new(&parameter_bytes[..]);
```

rust-bins/wallet-notes/README.md
```diff
@@ -607,6 +607,63 @@ The returned value is a JSON object with the following fields:
...
+Semantics: Converts a serialized transaction (as bytes) into JSON.
+
+This function takes as input a NUL-terminated UTF8-encoded string. The string
+must be a valid JSON object with fields
+
+- `"transaction"` ... a hex encoding of the serialized transaction without signatures, i.e. the serialized header, type and payload.
+
+- `"schema"` ... optional, required for update smart contract transactions, hex encoded schema for the corresponding smart contract.
+
+- `"schemaVersion"` ... optional, required for contracts without an embedded version to declare the version of the provided schema. The value is ignored if the version is embedded in the schema.
```
abizjak, line 654: Perhaps add that this kind of schema is only in older, legacy contracts.

abizjak (approved):

PR #240, opened 2022-10-20
WIP: Towards an update of BLS and pairing deps.

eb-concordium:
## Purpose

To update to the newest versions of the https://crates.io/crates/pairing and https://crates.io/crates/bls12_381 crates. 
Our current version (0.15.0) of the pairing crate has the bls12_381 module inside, but in newer versions it is in the separate crate https://crates.io/crates/bls12_381. 

STATUS: We need to be able to serialize `Curve::TargetField` elements, but the `Fp12` inside the target group `Gt` has been made private in https://crates.io/crates/bls12_381, so we are currently not able to serialize `Gt` elements. 
Concretely, this is needed for the Sigma protocol `ComEqSig` in order to produce a proof knowledge of a Pointcheval-Sanders Signature. The `id` crate does therefore not compile.  

## Changes

The crates
- bulletproofs/
- crypto_common
- curve_arithmetic
- dodis_yampolskiy_prf
- elgamal
- id
- keygen_bls
- pedersen_scheme
- ps_sig

have so far been touched in order to use the https://crates.io/crates/bls12_381 (the `id` crate does not compile).

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/bulletproofs/src/inner_product_proof.rs
+++ b/rust-src/bulletproofs/src/inner_product_proof.rs
@@ -3,6 +3,8 @@ use crypto_common_derive::*;
 use curve_arithmetic::{multiexp, Curve};
 use ff::Field;
 use random_oracle::RandomOracle;
+use std::ops::{Neg, AddAssign, SubAssign, MulAssign, Mul};
+
 
 #[derive(Clone, Serialize, Debug)]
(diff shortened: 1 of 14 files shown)


abizjak:
This is no longer relevant. We are instead going to use ark works dependency that is being implemented in a followup PR

PR #241, opened 2022-10-20
BP Set Membership Proof

tschudid:
## Purpose

This adds set membership proofs to the bulletproof crate.

## Changes

Add set membership prover and verifier.

## Checklist

      hard-to-understand areas.


rust-src/bulletproofs/benches/smp_bench.rs
```diff
@@ -0,0 +1,330 @@
...
+    /// Could not compute the inner product scalars
+    IpScalarCheckFailed,
+    /// Could not invert the given y
+    DivisionError,
+    /// Verification sum non-zero
+    VerificationSumNonZero,
+}
+
+#[allow(non_snake_case)]
+pub fn verify_ultra_efficient<C: Curve, R: Rng>(
```
abizjak, line 31: Can you please provide context for why we are doing this, why this is in the bench file, and what the conclusions are.

Essentially putting in this file what was discussed at the meeting last week.

rust-src/bulletproofs/src/inner_product_proof.rs
```diff
@@ -331,6 +332,118 @@ pub fn verify_inner_product<C: Curve>(
     P_prime.minus_point(&RHS).is_zero_point()
 }
 
+/// This function is an optimized variant of the above.
+/// It is verified whether P'=<a,G>+<b,H'>+<a,b>Q for P' =
```
abizjak, line 336: It seems like this function is verifying the same as the above one, is that correct?

In that case, why do we retain the old one.

It would also be useful to include some data as to what "optimized" means. What are the gains? That can be put into the PR description.

rust-src/bulletproofs/src/range_proof.rs
```diff
@@ -160,93 +123,108 @@ pub fn prove<C: Curve, T: Rng>(
     v_keys: &CommitmentKey<C>,
     randomness: &[Randomness<C>],
 ) -> Option<RangeProof<C>> {
+    // Part 1: Setup and generation of vector commitments
+    // V (for the values),
+    // A (their binary representation),
+    // S (the blinding factors)
     let nm = usize::from(n) * usize::from(m);
+    // TODO: Check that v_vec has the same length as randomness
+    // Check that we have enough generators for vector commitments
```
abizjak, line 132: Just add the check then.

rust-src/bulletproofs/src/range_proof.rs
```diff
@@ -160,93 +123,108 @@ pub fn prove<C: Curve, T: Rng>(
...
+    // TODO: Add proper error handling
```
abizjak, line 133: Remove the todo. There is nothing improper about the current error handling.

rust-src/bulletproofs/src/range_proof.rs
```diff
@@ -57,53 +67,6 @@ fn two_n_vec<F: Field>(n: u8) -> Vec<F> {
...
-    z_n
-}
-
-/// Struct containing generators G and H needed for range proofs
-#[allow(non_snake_case)]
-#[derive(Debug, Clone, Serialize, SerdeBase16Serialize)]
-pub struct Generators<C: Curve> {
-    #[size_length = 4]
-    pub G_H: Vec<(C, C)>,
-}
```
abizjak, line 86: This library is exported as part of the rust SDK. Removing this type from it is a needless breaking API change.

Please re-export this type from the module.

rust-src/bulletproofs/src/utils.rs
```diff
@@ -0,0 +1,132 @@
...
+#[allow(non_snake_case)]
+#[derive(Debug, Clone, Serialize, SerdeBase16Serialize)]
+pub struct Generators<C: Curve> {
+    #[size_length = 4]
+    pub G_H: Vec<(C, C)>,
+}
+
+impl<C: Curve> Generators<C> {
+    /// **Warning** do not use in production!
+    /// This **unsafely** generates a list of generators of a given size for
```
abizjak, line 17: THere is nothing inherently unsafe about this method. It is perfectly safe as long as the Rng is secure enough.

This comment is needless scaremongering.

Why does it even have to be pub if you've put it behind a test feature?

rust-src/bulletproofs/src/utils.rs
```diff
@@ -0,0 +1,132 @@
...
+            gh.push((x, y));
+        }
+        Self { G_H: gh }
+    }
+
+    pub fn take(&self, nm: usize) -> Self {
+        Self {
+            G_H: self.G_H[0..nm].to_vec(),
+        }
+    }
```
abizjak, line 35: Document, including precondition on `nm` and length of self.

rust-src/bulletproofs/src/utils.rs
```diff
@@ -0,0 +1,132 @@
...
+    let n = the_set.len();
+    let mut s_vec = Vec::with_capacity(n);
+    for elem_i in the_set {
+        let s_i = C::scalar_from_u64(*elem_i);
+        s_vec.push(s_i);
+    }
+    s_vec
+}
+
+/// Pads a field vector two a power of two length by repeating the last element
```
abizjak, line 71: That's not entirely what it does though, right. If the vector is empty nothing is done. Document this behaviour.
```suggestion
/// Pads a field vector to a power of two length by repeating the last element
```

rust-src/bulletproofs/src/utils.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use ff::Field;
+    use rand::thread_rng;
+
+    use super::{pad_vector_to_power_of_two, z_vec};
+
+    // type SomeCurve = pairing::bls12_381::G1;
```
abizjak, line 90: Remove.

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
+use crate::{inner_product_proof::*, utils::*};
```
abizjak, line 1: Please provide a module level comment, i.e., `//! ....` describing what is being done here, briefly.

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
...
+use ff::Field;
+use pedersen_scheme::*;
+use rand::*;
+use random_oracle::RandomOracle;
+use std::iter::once;
+
+#[derive(Clone, Serialize, SerdeBase16Serialize, Debug)]
+#[allow(non_snake_case)]
+pub struct SetMembershipProof<C: Curve> {
+    /// Commitment to the evaluation of the indicator function I_{v} on the_set
```
abizjak, line 14:
```suggestion
    /// Commitment to the evaluation of the indicator function `I_{v}` on the_set
```
and similarly for the comments below. Make use of markdown.

In general I would also recommend that all these fields are either private or at most pub(crate). They are an implementation detail .

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
...
+}
+
+/// This function takes a set (as a vector) and a value v as input.
+/// If v in S the function computes bit vectors aL and aR where
+/// aL_i = 1 <=> s_i = v
+/// and a_R is the bit-wise negation of a_L
+/// Note: For multi sets this function only sets the first hit to one.
+#[allow(non_snake_case)]
+fn a_L_a_R<F: Field>(v: &F, set_vec: &Vec<F>) -> Option<(Vec<F>, Vec<F>)> {
+    // TODO: Add proper error types
```
abizjak, line 54: Either resolve or remove the TODO.

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
...
+    let n = set_vec.len();
+    let mut a_L = Vec::with_capacity(n);
+    let mut a_R = Vec::with_capacity(n);
+    let mut found_element = false;
+    for i in 0..n {
+        let mut bit = F::zero();
+        let s_i = set_vec.get(i);
+        s_i?;
```
abizjak, line 62:
```suggestion
        let s_i = set_vec.get(i)?;
```

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
...
+        if (!found_element) && (*v == *s_i.unwrap()) {
```
abizjak, line 63:
```suggestion
        if (!found_element) && (v == s_i) {
```

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
...
+    InnerProductProofFailure,
+    /// Could not invert y
+    DivisionError,
+}
+
+/// This function takes a set (as a vector) and a value v as input.
+/// If v in S the function computes bit vectors aL and aR where
+/// aL_i = 1 <=> s_i = v
+/// and a_R is the bit-wise negation of a_L
+/// Note: For multi sets this function only sets the first hit to one.
```
abizjak, line 51: Can you document why this is the desired behaviour?

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
...
+        }
+        a_L.push(bit);
+        bit.sub_assign(&F::one());
+        a_R.push(bit);
+    }
+    // The set does not contain v
+    if !found_element {
+        return None;
+    }
+    Some((a_L, a_R))
```
abizjak, line 75:
```suggestion
    if found_element {
        Some((a_L, a_R))
    } else {
        None
    }
```

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
...
+}
+
+/// This function produces a set membership proof, i.e. a proof of knowledge
+/// of a value v that is in a given set the_set  and that is consistent with the
```
abizjak, line 79:
```suggestion
/// of a value v that is in a given set `the_set` and that is consistent with the
```

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
...
+/// commitment V to v. The arguments are
+/// - transcript - the random oracle for Fiat Shamir
+/// - csprng - cryptographic safe randomness generator
+/// - the_set - the set as a vector
+/// - v the value
+/// - gens - generators containing vectors G and H both of length nm
+/// - v_keys - commitment keys B and B_tilde
+/// - v_rand - the randomness used to commit to v using v_keys
```
abizjak, line 87:
```suggestion
/// - `transcript` - the random oracle for Fiat Shamir
/// - `csprng` - cryptographic safe randomness generator
/// - `the_set` - the set as a vector
/// - `v` the value
/// - `gens` - generators containing vectors G and H both of length nm
/// - `v_keys` - commitment keys B and B_tilde
/// - `v_rand` - the randomness used to commit to v using v_keys
```

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
...
+            S,
+            T_1,
+            T_2,
+            tx,
+            tx_tilde,
+            e_tilde,
+            ip_proof,
+        });
+    }
+    Err(ProverError::InnerProductProofFailure)
```
abizjak, line 334:
```suggestion
    if let Some(ip_proof) = proof {
        Ok(SetMembershipProof {
            A,
            S,
            T_1,
            T_2,
            tx,
            tx_tilde,
            e_tilde,
            ip_proof,
        })
    } else {
        Err(ProverError::InnerProductProofFailure)
    }
```

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
...
+    if !n.is_power_of_two() {
+        return Err(VerificationError::SetSizeNotPowerOfTwo);
+    }
+    if gens.G_H.len() < n {
+        return Err(VerificationError::NotEnoughGenerators);
+    }
+
+    // TODO: Check whether n fits into u64
+    // TODO: Check whether u64 integers fit into the field (note this should be the
+    // case assuming reasonable fields)
```
abizjak, line 381: THere is nothing to check here. That is assumed by C being a Curve, i.e., implementing that trait. Remove this TODO.

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
...
+    // Part 1: Setup
+    let n = the_set.len();
+    if !n.is_power_of_two() {
+        return Err(VerificationError::SetSizeNotPowerOfTwo);
+    }
+    if gens.G_H.len() < n {
+        return Err(VerificationError::NotEnoughGenerators);
+    }
+
+    // TODO: Check whether n fits into u64
```
abizjak, line 379: Do.

```
let n: u64 = the_set.len().try_into().map_err(|_| VerificationError::SetTooLarge)?;
```

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
...
+    /// of value v that is in a set S and that is consistent
+    /// with a commitment V to v. The arguments are
+    /// - S - the set as a vector
+    /// - V - commitment to v
+    /// - proof - the set membership proof to verify
+    /// - gens - generators containing vectors G and H both of length at least
+    ///   |S| (bold g,h in bluepaper)
+    /// - v_keys - commitment keys B and B_tilde (g,h in bluepaper)
+    #[allow(non_snake_case)]
+    pub fn verify_naive<C: Curve>(
```
abizjak, line 557: Please describe also why this is here. What is "naive" about it.

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,960 @@
...
+        let n = the_set.len();
+        if !n.is_power_of_two() {
+            return Err(VerificationError::SetSizeNotPowerOfTwo);
+        }
+        if gens.G_H.len() < n {
+            return Err(VerificationError::NotEnoughGenerators);
+        }
+
+        // TODO: Check whether n fits into u64
+        let the_set_vec = get_set_vector::<C>(the_set);
```
abizjak, line 575: Handle the todo.

abizjak (commented):
I did not go through the computations in the proofs, I do not have time for that at the moment and I hope @hamiidreza and @eb-concordium can do that.

Overall the code looks fine, there are minor issues here and there that should be addressed though.

rust-src/bulletproofs/src/inner_product_proof.rs
chrmatt: The old function is used in the range proof file, so we'd need to adjust that if we want to remove the old function.

rust-src/bulletproofs/src/range_proof.rs
chrmatt: What is meant here is that other functions return an error saying what went wrong, while this return none. It's in principle fine as is, but if we add the other checks, it seems to make sense to have different return values for different errors.

rust-src/bulletproofs/src/utils.rs
chrmatt: Even if RNG is completely random, it is unsafe in the sense that it (may) generate a random exponent x and then compute g^x. In that sense, the discrete log is "known" to the computer generating it and it must be deleted safely etc. So I think it is correct to say that this function should not be used in production.

rust-src/bulletproofs/src/set_membership_proof.rs
chrmatt: I suggest to remove the function. The reason we had is that it calls the existing inner product verification black box, while the other ones requires the optimized inner product verification with scalars. If you are OK with adding the new inner product verification, there is not reason to have this "naive" function here.

rust-src/bulletproofs/src/range_proof.rs
chrmatt: We need the struct also for the set-membership proof. We could import it there from the range-proof, but that seems the wrong place to have it. What is the proper way to resolve this?

rust-src/bulletproofs/benches/smp_bench.rs
chrmatt: Removed because the normal verify is equally efficient.

rust-src/bulletproofs/src/range_proof.rs
abizjak: It is what I suggested, to re-export.

```
pub use utils::Generators;
```

rust-src/bulletproofs/src/range_proof.rs
abizjak: The todo should not stay in the code, and there is nothing improper as is. Either resolve or remove.

TODOs left in code never get resolved in practice.

rust-src/bulletproofs/src/inner_product_proof.rs
abizjak: So why not if the behaviour is the same?

chrmatt: It takes different inputs, so the naive range proof verification won't work. Note that the range proof also has a naive and ultra efficient variant etc., and that could also be streamlined with the set-membership proof. But seemed out of scope for this. I think ideally we'd kill the naive range proof verification and adjust the regular one to use the new inner product with scalars function.

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,738 @@
...
+/// This function produces a set membership proof, i.e. a proof of knowledge
+/// of a value v that is in a given set `the_set` and that is consistent with
+/// the commitment `V` to `v`. The arguments are
+/// - `transcript` - the random oracle for Fiat Shamir
+/// - `csprng` - cryptographic safe randomness generator
+/// - `the_set` - the set as a vector
+/// - `v` the value
+/// - `gens` - generators containing vectors `G` and `H` both of at least length
+///   `n`
+/// - `v_keys` - commitment keys `B` and `B_tilde`
```
hamiidreza, line 89:
```suggestion
/// - `v_keys` - commitment keys `B` and `B_tilde` (`g,h` in bluepaper)
```

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,738 @@
...
+    let y_inv = match y.inverse() {
+        Some(inv) => inv,
+        None => return Err(ProverError::DivisionError),
+    };
+    let H_prime_scalars = z_vec(y_inv, 0, n);
+    // compute inner product proof
+    let proof =
+        prove_inner_product_with_scalars(transcript, &G, &H, &H_prime_scalars, &Q, &lx, &rx);
+
+    // return range proof
```
hamiidreza, line 324:
```suggestion
    // return set membership proof
```

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,738 @@
...
+    let S = proof.S;
+    // append commitments A and S to transcript
+    transcript.append_message(b"A", &A);
+    transcript.append_message(b"S", &S);
+
+    // get challenges y,z from transcript
+    let y: C::Scalar = transcript.challenge_scalar::<C, _>(b"y");
+    let z: C::Scalar = transcript.challenge_scalar::<C, _>(b"z");
+
+    // define the commitments A,S
```
hamiidreza, line 404:
```suggestion
    // define the commitments T1, T2
```

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,738 @@
...
+    // append T1, T2 commitments to transcript
+    transcript.append_message(b"T1", &T_1);
+    transcript.append_message(b"T2", &T_2);
+
+    // get challenge x (evaluation point) from transcript
+    let x: C::Scalar = transcript.challenge_scalar::<C, _>(b"x");
+
+    // define polynomial evaluation value
+    let tx = proof.tx;
+    // define blinding factors for tx and i.p. proof
```
hamiidreza, line 416:
```suggestion
    // define blinding factors for tx and IP proof
```

hamiidreza (changes requested):
The code looks correct and consistent with the bluepaper to me. 
I just suggested some minor changes.

rust-src/bulletproofs/src/range_proof.rs
tschudid: Added check.

rust-src/bulletproofs/src/range_proof.rs
tschudid: Removed todos

rust-src/bulletproofs/src/range_proof.rs
tschudid: Done

rust-src/bulletproofs/src/set_membership_proof.rs
tschudid: Done.

rust-src/bulletproofs/src/inner_product_proof.rs
```diff
@@ -331,6 +335,118 @@ pub fn verify_inner_product<C: Curve>(
...
+        rhs_bases.push(gens.G_H[i].0);
+    }
+    for i in 0..n {
+        rhs_bases.push(gens.G_H[i].1);
+    }
+
+    // add further elements to rhs_bases
+    rhs_bases.push(*Q);
+    rhs_bases.append(&mut nsum_bases);
+    rhs_bases.extend(P_prime_bases);
```
eb-concordium, line 418: It seems that Q = g_hat is already in `P_prime_bases` (at least where we are calling `verify_inner_product_with_scalars` from). Is it intentional that we are doing `Q^(ab) Q^(-tx)` and not `Q^(ab-tx)`? It probably does not matter much efficiency wise, just wondering.

chrmatt: The only reason is that we may want to call this function also if `P_prime_bases` is different and assuming it contains a particular element in a specific position would make the function less general.

hamiidreza (approved):

rust-src/bulletproofs/src/inner_product_proof.rs
eb-concordium: I see.

abizjak: Do you need this in the set non-membership proof? Or why is the generality useful @chrmatt ?

Generally one should wait with introducing generality until it is needed.

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,738 @@
...
+    // Generators for single commitments and blinding
+    let B = v_keys.g;
+    let B_tilde = v_keys.h;
+    // Compute aL (indicator vector) and aR
+    let maybe_aLaR = a_L_a_R(&v_scalar, &set_vec);
+    if maybe_aLaR.is_none() {
+        return Err(ProverError::CouldNotFindValueInSet);
+    }
+    let (a_L, a_R) = maybe_aLaR.unwrap();
+    // Setup blinding factors for a_L and a_R
```
abizjak, line 135:
```suggestion
    let (a_L, a_R) = a_L_a_R(&v_scalar, &set_vec).ok_or(ProverError::CouldNotFindValueInSet)?;
    // Setup blinding factors for a_L and a_R
```

rust-src/bulletproofs/src/utils.rs
```diff
@@ -0,0 +1,139 @@
...
+    if n == 0 {
+        return;
+    }
+    let k = n.next_power_of_two();
+    if let Some(last) = vec.last().cloned() {
+        let d = k - n;
+        for _ in 0..d {
+            vec.push(last)
+        }
+    }
```
abizjak, line 88: This function is not used apart from in the test that it works correctly.

Thus it should generally not be part of the PR. Why is it here?

chrmatt: I is probably needed at some higher level. We may want to prove statements about sets of size not a power of 2. And there it is crucial to extend the set properly, i.e., changing the multiset without changing the set.

rust-src/bulletproofs/src/inner_product_proof.rs
chrmatt: The generality is not needed for the non-membership. But it's not really introducing generality, but rather skipping an optimization: We currently already say that "It is assumed that the first base points are `G | H`, which are implicit and omitted from `P_prime_bases`", which allows a substantial speedup, but does not provide the nicest interface. We would now need to additional assume that `Q` is also implicitly there. We can do that if you think it makes sense, but it seems to overcomplicate the usage of the function.

rust-src/bulletproofs/src/inner_product_proof.rs
chrmatt: Is this resolved for now? We can make a separate PR for revising the range proof and killing the then unused functions.

rust-src/bulletproofs/src/utils.rs
chrmatt: I revised the comment and added an explanation why it should not be used in production. Is this OK?

rust-src/bulletproofs/src/utils.rs
tschudid: We could remove it here, but it will be needed if we want to prove statements about sets with arbitrary not a power of 2. Should we create adapter functions that allow to prove statements for sets of arbitrary size?

rust-src/bulletproofs/src/inner_product_proof.rs
abizjak: This is (or rather should be) an internal function in this crate.

In that case a complicated interface is fine. I don't particularly mind if it stays like this, but I was replying to the statement you made about "more general". 

But we should make this function `pub(crate)`. I don't think we need it outside the bulletproofs.

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,734 @@
...
+    DivisionError,
+    /// Inner product proof verification failed
+    IPVerificationError,
+}
+
+/// This function verifies a set membership proof, i.e. a proof of knowledge
+/// of value v that is in a set S and that is consistent
+/// with a commitment V to v. The arguments are
+/// - `transcript` - the random oracle for Fiat Shamir
+/// - `the_set` - the set as a vector
```
abizjak, line 358: Please make it explicit here that this only works for sets of size power of two. In bold :) 

Alternatively just pad in this function (and in the prover). Nobody outside of this crate should use this verifier and prover directly.

Or add another function that pads + verifies and pads + proves.

abizjak (approved):

rust-src/bulletproofs/src/utils.rs
abizjak: See my comment below. You have a good point that this function should stay, and we should probably just use it in the prover and verifier (at least the ones we expose).

rust-src/bulletproofs/src/set_membership_proof.rs
tschudid: Added padding directly into prove and verify.

rust-src/bulletproofs/src/inner_product_proof.rs
chrmatt: Made this function `pub(crate)` and added the Q to the implicit bases.

rust-src/bulletproofs/src/utils.rs
chrmatt: Now used in prove and verify.

rust-src/bulletproofs/src/set_membership_proof.rs
```diff
@@ -0,0 +1,702 @@
...
+}
+
+/// This function takes a set (as a vector) and a value v as input.
+/// If v in S the function computes bit vectors aL and aR where
+/// aL_i = 1 <=> s_i = v
+/// and a_R is the bit-wise negation of a_L
+/// Note: For multi sets this function only sets the first hit to one, to allow
+/// set membership proofs in multi sets.
+#[allow(non_snake_case)]
+fn a_L_a_R<F: Field>(v: &F, set_vec: &Vec<F>) -> Option<(Vec<F>, Vec<F>)> {
```
eb-concordium, line 55: I think it would be better to take a slice, i.e. `&[F]`,  and then iterate over the slice instead.

chrmatt: Why?

abizjak: There is no benefit in passing a &Vec<A> instead of &[A] unless for some reason you care about capacity. The former is a pointer to a pointer, and the latter is just a pointer to a slice. 

See https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg

chrmatt: Thanks!

PR #242, opened 2022-10-20
Account Stake Hashing Changes

td202:
## Purpose

Small changes to the hashing of account stakes.

## Changes

- For `AccountV2`, the stake hash does not include the staked amount, but does include the "delegator ID" and pending change. (Note, the delegator ID is probably not really necessary, but doesn't do harm to include. The pending change should be included, but was wrongly omitted from `AccountV1` stake hashing.)
- The prefixes for hashing are changed to distinguish them from previous account versions.

## Checklist

      hard-to-understand areas.


haskell-src/Concordium/Types/Accounts.hs
```diff
@@ -471,31 +487,31 @@ instance HashableTo (AccountStakeHash 'AccountV1) (AccountStake 'AccountV1) wher
 -- |Hash of 'AccountStakeNone' in 'AccountV2'.
 accountStakeNoneHashV2 :: AccountStakeHash 'AccountV2
 {-# NOINLINE accountStakeNoneHashV2 #-}
-accountStakeNoneHashV2 = AccountStakeHash $ Hash.hash "NoStake"
+accountStakeNoneHashV2 = AccountStakeHash $ Hash.hash "A2NoStake"
 
--- |The 'AccountV2' hashing of 'AccountStake' uses tags to enforce distinction between the cases.
+-- |The 'AccountV2' hashing of 'AccountStake' DOES NOT INCLUDE the staked amount.
```
abizjak, line 492: It would be useful to elaborate as to why that is the case.

abizjak (approved):
Please document the reason why the staked amount is not included. Otherwise looks good.

PR #247, opened 2022-10-26
Base changes supporting a new hashing scheme for transaction outcomes.

EmilLa1:
## Purpose

Introduce  multiple versions of `TransactionOutcomes` which can be used for supporting different kinds of hashing schemes.
In particular from P5 and onwards we do not want to hash the exact `RejectReason` anymore. (See https://github.com/Concordium/concordium-node/issues/566)

There are two versions of `TransactionOutcomes` and these are determined by the `ProtocolVersion`. 
PV1 to PV4 makes use of `TransactionOutcomesVersion 0` (`TOV0`) while P5 uses `TOV1`. 

Further there is introduced a `TransactionSummaryV1` type which is just a `newtype` wrapper around the existing `TransactionSummary` type. The purpose of this is to allow for a different `HashableTo` instance which omits the exact reject reason in the computed hash.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Types.hs
+++ b/haskell-src/Concordium/Types.hs
@@ -116,8 +116,6 @@ module Concordium.Types (
   BlockNonce,
   BlockSignature,
   BlockProof,
-  TransactionOutcomesHashV0(..),
-  TransactionOutcomesHash,
   StateHashV0(..),
   StateHash,
   BlockHash(..),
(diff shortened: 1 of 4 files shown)


abizjak (approved):
In a better world we would have revised `transactionOutcomes` to be parameterized by a transactionoutcomesversion, and then we'd have the hash type also parameterized by that.

But that is too much change at the moment and that refactoring can be done after since it does not change the protocol.

So let's merge this now.

PR #249, opened 2022-10-27
Relax smart contract restrictions

Bargsteen:
## Purpose

Relax restrictions for parameter and return value size and number of logs. See https://github.com/Concordium/concordium-node/pull/573 for more details.

Solves part of https://github.com/Concordium/concordium-node/issues/563.

## Changes

- Add functions for getting the limits, which depend on the PV.
- Remove `Serialize` instance for `Parameter` as we now need the PV for determining the max size.
   - Added put/get functions for it instead.
- Pass in the limits from the scheduler and save it in the init/receive host, to be used when relevant.
- Change the cost function for parameters larger than 1kb.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Constants.hs
+++ b/haskell-src/Concordium/Constants.hs
@@ -25,8 +25,22 @@ maxPayloadSize SP5 = maxPayloadSize SP4
 -- * Web assembly related constants
 
 -- |Maximum length of the parameter to init and receive methods.
-maxParameterLen :: Word16
-maxParameterLen = 1024
+-- The limit was changed from `1024` to `65535` in P5.
+maxParameterLen :: SProtocolVersion pv -> Word16
+maxParameterLen SP1 = 1024
+maxParameterLen SP2 = 1024
+maxParameterLen SP3 = 1024
+maxParameterLen SP4 = 1024
+maxParameterLen _ = 65535
+
+-- |Whether the number of logs and size of return values should be limited.
+-- The limits have been removed in P5 and onward.
+limitLogsAndReturnValues :: SProtocolVersion pv -> Bool
+limitLogsAndReturnValues SP1 = True
+limitLogsAndReturnValues SP2 = True
+limitLogsAndReturnValues SP3 = True
+limitLogsAndReturnValues SP4 = True
+limitLogsAndReturnValues _ = False
 
 -- |Maximum module size of a V0 module.
 maxWasmModuleSizeV0 :: Word32
--- a/haskell-src/Concordium/Constants.hs
+++ b/haskell-src/Concordium/Constants.hs
@@ -31,7 +31,7 @@ maxParameterLen SP1 = 1024
 maxParameterLen SP2 = 1024
 maxParameterLen SP3 = 1024
 maxParameterLen SP4 = 1024
-maxParameterLen _ = 65535
+maxParameterLen SP5 = 65535
 
 -- |Whether the number of logs and size of return values should be limited.
 -- The limits have been removed in P5 and onward.
@@ -40,7 +40,7 @@ limitLogsAndReturnValues SP1 = True
 limitLogsAndReturnValues SP2 = True
 limitLogsAndReturnValues SP3 = True
 limitLogsAndReturnValues SP4 = True
-limitLogsAndReturnValues _ = False
+limitLogsAndReturnValues SP5 = False
 
 -- |Maximum module size of a V0 module.
 maxWasmModuleSizeV0 :: Word32
--- a/haskell-src/Concordium/Types/Execution.hs
+++ b/haskell-src/Concordium/Types/Execution.hs
@@ -2088,8 +2088,9 @@ instance S.Serialize RejectReason where
       rejectReason <- S.getInt32be
       contractAddress <- S.get
       receiveName <- S.get
-      -- Get the parameter without validating its length. This is OK here, because the length has been checked before reaching this point.
-      -- Checking requires the protocol version as input, which is why it is skipped here.
+      -- We only ever deserialize valid transactions, which means that the parameter size is known to be valid.
+      -- This allows us to skip the size check, which is useful because checking requires the protocol version,
+      -- which we do not currently have access to here.
       parameter <- Wasm.getParameterUnchecked
       return RejectedReceive {..}
     14 -> return InvalidProof
--- a/haskell-src/Concordium/Types/Execution.hs
+++ b/haskell-src/Concordium/Types/Execution.hs
@@ -410,13 +410,13 @@ putPayload InitContract{..} =
       S.put icAmount <>
       S.put icModRef <>
       S.put icInitName <>
-      S.put icParam
+      Wasm.putParameter icParam
 putPayload Update{..} =
     P.putWord8 2 <>
     S.put uAmount <>
     S.put uAddress <>
     S.put uReceiveName <>
-    S.put uMessage
+    Wasm.putParameter uMessage
 putPayload Transfer{..} =
     P.putWord8 3 <>
     S.put tToAddress <>
@@ -560,13 +560,13 @@ getPayload spv size = S.isolate (fromIntegral size) (S.bytesRead >>= go)
               icAmount <- S.get
               icModRef <- S.get
               icInitName <- S.get
-              icParam <- S.get
+              icParam <- Wasm.getParameter spv
               return InitContract{..}
             2 -> do
               uAmount <- S.get
               uAddress <- S.get
               uReceiveName <- S.get
-              uMessage <- S.get
+              uMessage <- Wasm.getParameter spv
               return Update{..}
             3 -> do
               tToAddress <- S.get
@@ -1062,7 +1062,7 @@ putEvent = \case ModuleDeployed mref ->
                    S.put euAddress <>
                    S.put euInstigator <>
                    S.put euAmount <>
-                   S.put euMessage <>
+                   Wasm.putParameter euMessage <>
                    S.put euReceiveName <>
                    S.put euContractVersion <>
                    S.putWord32be (fromIntegral (length euEvents)) <>
@@ -1245,7 +1245,7 @@ getEvent spv =
       euAddress <- S.get
       euInstigator <- S.get
       euAmount <- S.get
-      euMessage <- S.get
+      euMessage <- Wasm.getParameter spv
       euReceiveName <- S.get
       euContractVersion <- S.get
       euEventsLength <- fromIntegral <$> S.getWord32be
@@ -2026,7 +2026,7 @@ instance S.Serialize RejectReason where
       S.putInt32be rejectReason <>
       S.put contractAddress <>
       S.put receiveName <>
-      S.put parameter
+      Wasm.putParameter parameter
     InvalidProof -> S.putWord8 14
     AlreadyABaker bid -> S.putWord8 15 <> S.put bid
     NotABaker addr -> S.putWord8 16 <> S.put addr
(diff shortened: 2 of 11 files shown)


haskell-src/Concordium/Wasm.hs
```diff
@@ -453,11 +457,22 @@ newtype Parameter = Parameter { parameter :: ShortByteString }
...
-  get = do
+-- |Put (serialize) a @Parameter@.
+putParameter :: Putter Parameter
+putParameter = putShortByteStringWord16 . parameter
+
+-- |Get (deserialize) a @Parameter@ and ensure that its size is valid. The size limit depends on the protocol version.
+getParameter :: SProtocolVersion pv -> Get Parameter
+getParameter spv = do
+    len <- getWord16be
+    unless (len <= maxParameterLen spv) $ fail "Parameter size exceeds limits."
```
EmilLa1, line 468:
```suggestion
    unless (len <= maxParameterLen spv) $ fail "Parameter size exceeds the limit."
```

haskell-src/Concordium/Constants.hs
```diff
@@ -25,8 +25,22 @@ maxPayloadSize SP5 = maxPayloadSize SP4
...
 -- |Maximum length of the parameter to init and receive methods.
-maxParameterLen :: Word16
-maxParameterLen = 1024
+-- The limit was changed from `1024` to `65535` in P5.
+maxParameterLen :: SProtocolVersion pv -> Word16
+maxParameterLen SP1 = 1024
+maxParameterLen SP2 = 1024
+maxParameterLen SP3 = 1024
+maxParameterLen SP4 = 1024
+maxParameterLen _ = 65535
```
EmilLa1, line 34: Why not qualify the SP5 case also ?

haskell-src/Concordium/Wasm.hs
EmilLa1: I guess there is only one limit?

smart-contracts/wasm-chain-integration/src/constants.rs
```diff
@@ -4,15 +4,12 @@ pub const MAX_CONTRACT_STATE: u32 = 16384; // 16kB
...
-
 /// Maximum size of a log message.
 pub const MAX_LOG_SIZE: u32 = 512;
 
-/// Maximum number of log messages per execution.
-/// This together with the previous constant limits the amount of data that can
-/// be logged to 16kB.
+/// Maximum number of log messages per execution *in protocol version < 5* (no
+/// limit from 5 and onward). This together with the previous constant limits
+/// the amount of data that can be logged to 16kB.
```
EmilLa1, line 12: Isn´t it bound by the block now (p5 and onwards) as opposed to the 16kb specified here?

haskell-src/Concordium/Constants.hs
abizjak, line 34:
```suggestion
maxParameterLen SP5 = 65535
```
It is more maintainable this way.

haskell-src/Concordium/Constants.hs
```diff
@@ -25,8 +25,22 @@ maxPayloadSize SP5 = maxPayloadSize SP4
...
+
+-- |Whether the number of logs and size of return values should be limited.
+-- The limits have been removed in P5 and onward.
+limitLogsAndReturnValues :: SProtocolVersion pv -> Bool
+limitLogsAndReturnValues SP1 = True
+limitLogsAndReturnValues SP2 = True
+limitLogsAndReturnValues SP3 = True
+limitLogsAndReturnValues SP4 = True
+limitLogsAndReturnValues _ = False
```
abizjak, line 43: Same here.

haskell-src/Concordium/Types/Execution.hs
```diff
@@ -2088,7 +2088,9 @@ instance S.Serialize RejectReason where
       rejectReason <- S.getInt32be
       contractAddress <- S.get
       receiveName <- S.get
-      parameter <- S.get
+      -- Get the parameter without validating its length. This is OK here, because the length has been checked before reaching this point.
+      -- Checking requires the protocol version as input, which is why it is skipped here.
```
abizjak, line 2092: Perhaps add that "before reaching this point" means that this is always an outcome of a valid transaction.

Which means that validation has already happened, so we don't need to do it again.

haskell-src/Concordium/Wasm.hs
```diff
@@ -665,7 +680,7 @@ getActionsTree' size = go HM.empty 0
                            erAddr <- get
                            erName <- get
                            erAmount <- get
-                           erParameter <- get
+                           erParameter <- getParameterUnchecked -- TODO: Ensure interpreter checks the length.
```
EmilLa1, line 683: There's a todo here.

abizjak, line 683: There is this in the `send` function
```rust
        ensure!(
            parameter_bytes.len() <= constants::MAX_PARAMETER_SIZE,
            "Parameter exceeds max size."
        );
```

abizjak: So this TODO can be resolved.

smart-contracts/wasm-chain-integration/src/constants.rs
abizjak, line 12: In protocol version 4 and lower.

smart-contracts/wasm-chain-integration/src/v0/mod.rs
```diff
@@ -26,9 +26,9 @@ impl Logs {
     /// - 0 if data was not logged because it would exceed maximum number of
     ///   logs
     /// - 1 if data was logged.
-    pub fn log_event(&mut self, event: Vec<u8>) -> i32 {
+    pub fn log_event(&mut self, event: Vec<u8>, limit_num_logs: bool) -> i32 {
         let cur_len = self.logs.len();
-        if cur_len < constants::MAX_NUM_LOGS {
+        if !limit_num_logs || cur_len < constants::MAX_NUM_LOGS {
```
abizjak, line 31: For extra safety please check that `cur_len < u32::MAX` so that we can serialize this thing.

smart-contracts/wasm-chain-integration/src/v0/mod.rs
```diff
@@ -192,36 +190,44 @@ impl State {
...
     /// The parameter to the receive method.
-    pub param:             ParamType,
+    pub param:                    ParamType,
     /// Outcomes of the execution, i.e., the actions tree.
-    pub outcomes:          Outcome,
+    pub outcomes:                 Outcome,
     /// The receive context for this call.
-    pub receive_ctx:       Ctx,
+    pub receive_ctx:              Ctx,
+    /// The maximum parameter size. Is constant for a given protocol version.
```
abizjak, line 226: It would make sense to document here that for P1-P4 it was ... and P5 it is ...

smart-contracts/wasm-chain-integration/src/v1/mod.rs
```diff
@@ -245,7 +256,11 @@ mod host {
             .checked_add(length)
             .ok_or_else(|| anyhow::anyhow!("Writing past the end of memory."))?
             as usize;
-        let end = std::cmp::min(end, constants::MAX_CONTRACT_STATE as usize) as u32;
+        let end = if limit_return_value_size {
+            std::cmp::min(end, constants::MAX_CONTRACT_STATE as usize) as u32
+        } else {
+            end as u32
```
abizjak, line 262: Why are we converting back to u32 here? THat is a bit dangerous although probably impossible to trigger in practice.

I recommend that we explicitly check that `bytes.len()` fits into a u32.

And we don't do conversion from usize to u32 at all.

abizjak (commented):
Looks good. A few corner cases here and there that should be handled more robustly.

Bargsteen:
> Great job, looks good!
> 
> I was wondering about another solution where you pass the protocol version through the FFI as the whole context is given from the PV anyway in this case. So if we in the future are to add more features/conditionals/ then they could be handled in the rust side and the FFI API arguments wouldn't grow as much. Also the smart contract logic would be more biased towards the rust side (which I think in general would be a good idea wrt. maintenance. )
> 
> I think it is at least worth a thought for future features.

Thanks.
That's an interesting suggestion, which might make sense. But I don't want to change it now due to the deadline approaching ;)

abizjak:
> > Great job, looks good!
> > I was wondering about another solution where you pass the protocol version through the FFI as the whole context is given from the PV anyway in this case. So if we in the future are to add more features/conditionals/ then they could be handled in the rust side and the FFI API arguments wouldn't grow as much. Also the smart contract logic would be more biased towards the rust side (which I think in general would be a good idea wrt. maintenance. )
> > I think it is at least worth a thought for future features.
> 
> Thanks. That's an interesting suggestion, which might make sense. But I don't want to change it now due to the deadline approaching ;) Perhaps we can change it if/when we need to pass in more PV-related parameters to the interpreter.

We have discussed that solution also for the upgrades. I think it is better to pass specific data like this so as to not duplicate the logic about what is or is not enabled too much.

haskell-src/Concordium/Constants.hs
Bargsteen: I actually thought `_` would be more maintainable as we wouldn't need to update this code for future updates unless we change the limits. But I guess we also risk forgetting something with the wildcard ;)

haskell-src/Concordium/Types/Execution.hs
Bargsteen: I've rephrased it, let me know if it is ok.

haskell-src/Concordium/Types/Execution.hs
```diff
@@ -2088,7 +2088,9 @@ instance S.Serialize RejectReason where
       rejectReason <- S.getInt32be
       contractAddress <- S.get
       receiveName <- S.get
-      parameter <- S.get
+      -- Get the parameter without validating its length. This is OK here, because the length has been checked before reaching this point.
```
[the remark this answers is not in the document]
Bargsteen: I've rephrased it. Let me know if it is ok.

smart-contracts/wasm-chain-integration/src/constants.rs
Bargsteen: Yes, I tried to make that clear, but apparently, it wasn't :p. I've rephrased it now ;)

smart-contracts/wasm-chain-integration/src/v1/mod.rs
Bargsteen: I've tried to clean up all the conversion shenanigans ;)

haskell-src/Concordium/Constants.hs
abizjak: It is more maintainable from the perspective of having to make fewer changes if we add a new protocol version. However this is largely a trivial thing to do and the compiler complains.

It is less maintainable from the perspective that now we have an implicit dependency on the protocol version that the compiler will not complain about.

smart-contracts/wasm-chain-integration/src/v0/mod.rs
```diff
@@ -26,9 +26,9 @@ impl Logs {
     /// - 0 if data was not logged because it would exceed maximum number of
     ///   logs
     /// - 1 if data was logged.
-    pub fn log_event(&mut self, event: Vec<u8>) -> i32 {
+    pub fn log_event(&mut self, event: Vec<u8>, limit_num_logs: bool) -> i32 {
         let cur_len = self.logs.len();
-        if cur_len < constants::MAX_NUM_LOGS {
+        if (!limit_num_logs && cur_len < u32::MAX as usize) || cur_len < constants::MAX_NUM_LOGS {
```
abizjak, line 31:
```suggestion
        if (!limit_num_logs && cur_len <= u32::MAX as usize) || cur_len < constants::MAX_NUM_LOGS {
```

abizjak (approved):

PR #250, opened 2022-10-28
Simplify range proof verifier

chrmatt:
## Purpose

There have been three different verification methods for range proofs: A naive one, an efficient one (which is actually used), and a more efficient one. This pull request changes the efficient verification to use the same inner-product verification as the set-membership proofs. This makes that method essentially as simple and modular as the naive one and almost as efficient as the more efficient one (the performance benefits of the more efficient one appear negligible, < 1% on a desktop PC).

## Changes

- Use inner-product verification from set-membership proof for range-proof verification.
- Remove naive verification.
- Remove more efficient verification.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/bulletproofs/src/inner_product_proof.rs
+++ b/rust-src/bulletproofs/src/inner_product_proof.rs
@@ -1,6 +1,5 @@
 //! Logarithmic sized inner product proof used as base for the other proofs in
 //! this crate
-use crate::utils::*;
 use crypto_common::*;
 use crypto_common_derive::*;
 use curve_arithmetic::{multiexp, Curve};
@@ -29,7 +28,7 @@ pub struct InnerProductProof<C: Curve> {
 /// - Q - the elliptic curve point Q
 /// - a_slice - a slice to the vector a of scalars
 /// - b_slice - a slice to the vector b of scalars
-/// Precondictions:
+/// Preconditions:
 /// G_slice, H_slice, a_slice and b_slice should all be of the same length, and
 /// this length must be a power of 2.
 #[allow(non_snake_case)]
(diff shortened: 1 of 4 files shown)


rust-src/bulletproofs/src/inner_product_proof.rs
```diff
@@ -295,44 +297,29 @@ pub fn verify_inner_product<C: Curve>(
...
+    verify_inner_product_with_scalars(
+        transcript,
+        &H_exponents,
+        &P_prime_bases,
+        &P_prime_exponents,
+        proof,
+    )
 }
 
 /// This function is an optimized variant of the above.
```
tschudid, line 325:
```suggestion
/// This function is the actual verification function used at the core of the above.
```

tschudid (approved):
Looks good

rust-src/bulletproofs/src/inner_product_proof.rs
```diff
@@ -295,81 +297,60 @@ pub fn verify_inner_product<C: Curve>(
...
-        .plus_point(&Q.mul_by_scalar(&ab))
-        .minus_point(&sum);
-    P_prime.minus_point(&RHS).is_zero_point()
+    let H_exponents = vec![C::Scalar::one(); n];
+    // P_prime is also given directly, so set P_prime_bases to P_prime with exponent
+    // 1. Since it is assumed that P_prime_bases starts with G, H and Q, add those
+    // first and set the first 2n+1 exponents to 0.
+    let mut P_prime_bases = Vec::with_capacity(2 * n + 2);
+    P_prime_bases.extend(G_vec);
+    P_prime_bases.extend(H_vec);
```
abizjak, line 309: I think it would be better to use `extend_from_slice`. 

But it does not matter in practice. It is just more precise a method that does exactly what you need here more directly than the generic extend.

rust-src/bulletproofs/src/inner_product_proof.rs
```diff
@@ -295,81 +297,60 @@ pub fn verify_inner_product<C: Curve>(
...
     transcript: &mut RandomOracle,
-    gens: &Generators<C>,
     H_exponents: &[C::Scalar],
     P_prime_bases: &[C],
     P_prime_exponents: &[C::Scalar],
-    Q: &C,
     proof: &InnerProductProof<C>,
 ) -> bool {
-    let n = gens.G_H.len();
+    let n = H_exponents.len();
```
abizjak, line 353: This seems like a significant change. Why was it previously not an issue? Do we always supply generators of the right size here?

abizjak (commented):
This looks good to me, and I always like removing code. :)

Just to make sure, since we have not changed the prover here at all, the semantics of the verification has not changed, right?

Have you also done any benchmarking to make sure there are no accidental regressions? I don't expect them from looking at the changes.

rust-src/bulletproofs/src/inner_product_proof.rs
chrmatt: It was an issue before. It was undetected because all tests supplied generators of the right size (and this function was only used for set-membership, which is not used in practice, yet), but it is not a precondition. We should probably add an explicit test case for having additional generators.

chrmatt:
> Just to make sure, since we have not changed the prover here at all, the semantics of the verification has not changed, right?
> 
> Have you also done any benchmarking to make sure there are no accidental regressions? I don't expect them from looking at the changes.

Yes, the verification should verify/reject the same proofs as before.

The new code has been benchmarked against the more efficient, randomized variant from before and is essentially equally efficient.

rust-src/bulletproofs/src/inner_product_proof.rs
abizjak: Could you add a small test for that in this PR?

abizjak (approved):

rust-src/bulletproofs/src/inner_product_proof.rs
chrmatt: Done.

PR #251, opened 2022-10-31
Add BP set-non-membership proofs.

chrmatt:
## Purpose

This pull request adds set-non-membership proofs based on bulletproofs.

## Changes

Add set-non-membership proofs, similar to set-membership proofs.

## Checklist

      hard-to-understand areas.


rust-src/bulletproofs/src/set_non_membership_proof.rs
```diff
@@ -0,0 +1,668 @@
...
+    let (mut G, mut H): (Vec<_>, Vec<_>) = gens.G_H.iter().take(n).cloned().unzip();
+    GH_B_tilde.append(&mut G);
+    GH_B_tilde.append(&mut H);
+
+    // Generators for single commitments and blinding
+    let B = v_keys.g;
+    let B_tilde = v_keys.h;
+    GH_B_tilde.push(B_tilde);
+
+    // define aliases to make G and H available again
```
abizjak, line 99: I don't understand this comment. What does "available again" mean?

rust-src/bulletproofs/src/set_non_membership_proof.rs
```diff
@@ -0,0 +1,668 @@
...
+    }
+}
+
+/// Error messages detailing why proof verification failed
+#[derive(Debug, PartialEq, Eq)]
+pub enum VerificationError {
+    /// The length of `gens` was less than `|the_set|`
+    NotEnoughGenerators,
+    /// The consistency check for `t_0` failed
+    InconsistentT0,
```
abizjak, line 306: Can this be made more understandable? Is it possible to state in a sentence or two what T0 is?

abizjak (approved):

rust-src/bulletproofs/src/set_non_membership_proof.rs
chrmatt: `GH_B_tilde.append(&mut G);` above moves the elements out of G; this redefines it as the corresponding sub-slice. Probably that's not the right way to do it, though. ;) We tried some things and tested for performance, but generally the fact that `gens` contains `G` and `H` zipped does not help here...

abizjak: I see. Probably a more appropriate solution would be
```
let (G, H) = gens.G_H.iter().take(n).unzip();
GH_B_tilde.extend(G.chain(H).copied();
let G = &GH_B_tilde[0..n];
let H = &GH_B_tilde[n..2 * n];
```

chrmatt: Rust doesn't let me unzip without cloning...

chrmatt: Cool, we'll use that then. But yeah, it doesn't really matter, and we copy things multiple times in other places.

rust-src/bulletproofs/src/set_non_membership_proof.rs
```diff
@@ -0,0 +1,674 @@
...
+use pedersen_scheme::*;
+use rand::*;
+use random_oracle::RandomOracle;
+use std::iter::once;
+
+/// Bulletproof style set-non-membership proof
+#[derive(Clone, Serialize, SerdeBase16Serialize, Debug)]
+#[allow(non_snake_case)]
+pub struct SetNonMembershipProof<C: Curve> {
+    /// Commitment to the multiplicative inverse of `v-s_i` for each `i`
```
hamiidreza, line 16: I would call A a commitment to the multiplicative inverse of `v-s_i` for each `i`, and an `all-v` vector

rust-src/bulletproofs/src/set_non_membership_proof.rs
```diff
@@ -0,0 +1,674 @@
...
+    transcript.append_message(b"tx", &tx);
+    transcript.append_message(b"tx_tilde", &tx_tilde);
+    transcript.append_message(b"e_tilde", &e_tilde);
+
+    // Part 5: Inner product proof for tx = <lx,rx>
+    // get challenge w from transcript
+    let w: C::Scalar = transcript.challenge_scalar::<C, _>(b"w");
+    // get generator q
+    let Q = B.mul_by_scalar(&w);
+    // compute scalars c such that c*H = H', that is H_prime_scalars = (1, y^-1,..,
```
hamiidreza, line 278:
```suggestion
    // compute scalars c such that H^c = H', that is H_prime_scalars = (1, y^-1,..,
```

rust-src/bulletproofs/src/set_non_membership_proof.rs
```diff
@@ -0,0 +1,674 @@
...
+        })
+    } else {
+        Err(ProverError::InnerProductProofFailure)
+    }
+}
+
+/// Error messages detailing why proof verification failed
+#[derive(Debug, PartialEq, Eq)]
+pub enum VerificationError {
+    /// The length of `gens` was less than `|the_set|`
```
hamiidreza, line 309:
```suggestion
    /// The length of the generator vector `gens` was too short
```

rust-src/bulletproofs/src/set_non_membership_proof.rs
chrmatt: Indeed! The same mistake was there for the set-membership and range-proofs as well. I've updated all of the comments. Please check.

rust-src/bulletproofs/src/set_non_membership_proof.rs
```diff
@@ -0,0 +1,674 @@
...
+
+/// This function produces a set-non-membership proof, i.e., a proof of
+/// knowledge of a value v that is not in a given set `the_set` and that is
+/// consistent with the commitment `V` to `v`. The arguments are
+/// - `transcript` - the random oracle for Fiat Shamir
+/// - `csprng` - cryptographic safe randomness generator
+/// - `the_set` - the set as a vector of scalars
+/// - `v` the value, a scalar
+/// - `gens` - generators containing vectors `G` and `H` both of at least length
+///   `k` where k is the smallest power of two >= `|the_set|`
```
hamiidreza, line 56:
```suggestion
///   `k` where k is the smallest power of two >= `|the_set|` (bold **g**,**h**
///   in bluepaper)
```

rust-src/bulletproofs/src/set_non_membership_proof.rs
```diff
@@ -0,0 +1,674 @@
...
+    let yn = z_vec(y, 0, n);
+    let one_vec = vec![C::Scalar::one(); n];
+    let ip_one_yn = inner_product(&one_vec, &yn);
+    let mut zip_s_yn = inner_product(&set_vec, &yn);
+    zip_s_yn.mul_assign(&z);
+    let mut delta_yz = ip_one_yn;
+    delta_yz.sub_assign(&zip_s_yn);
+
+    // Part 2: Verify consistency of t_0
+    // i.e., check 0 = V^(z <1, yn>) * g^(delta_yz - tx) * T_1^x * T_2^x^2 *
```
hamiidreza, line 401: with multiplicative notation, it should be '1' on the LHS; but that would also make the reader confused!

rust-src/bulletproofs/src/set_non_membership_proof.rs
```diff
@@ -0,0 +1,674 @@
...
+    let t0_check_base_points = vec![V.0, v_keys.g, T_1, T_2, v_keys.h];
+    let t0_check_exponents = vec![zip_one_yn, delta_minus_tx, x, x2, minus_tx_tilde];
+
+    let rhs = multiexp(&t0_check_base_points, &t0_check_exponents);
+    if !rhs.is_zero_point() {
+        return Err(VerificationError::InconsistentT0);
+    }
+
+    // Part 3: Verify inner product
+    // First compute helper variables g_hat, h_prime, and P_prime
```
hamiidreza, line 421: Any reason why we call it g_hat here and Q in the prover side?

rust-src/bulletproofs/src/set_non_membership_proof.rs
chrmatt: Yeah. This also applies to the other proofs. We could say instead 1_G as in the bluepaper or write verbosely "check that ... is neutral element".

rust-src/bulletproofs/src/set_non_membership_proof.rs
chrmatt: Legacy. It used to be Q everywhere for the range proof because it comes from Dalek. I always tried to use bluepaper notation instead and sometimes added comments saying that this corresponds to that in bluepaper.

rust-src/bulletproofs/src/set_non_membership_proof.rs
```diff
@@ -0,0 +1,674 @@
...
+
+    let y_inv = match y.inverse() {
+        Some(inv) => inv,
+        None => return Err(VerificationError::DivisionError),
+    };
+    let y_inv_n = z_vec(y_inv, 0, n);
+    let mut minus_e_tilde = e_tilde;
+    minus_e_tilde.negate();
+
+    // get exponent for g, i.e., [z, z, ..., z]
```
hamiidreza, line 432:
```suggestion
    // get exponent for bold **g**, i.e., [z, z, ..., z]
```

rust-src/bulletproofs/src/set_non_membership_proof.rs
```diff
@@ -0,0 +1,674 @@
...
+    let mut gexp = vec![z; n];
+
+    let mut P_prime_exps = Vec::with_capacity(2 * n + 4);
+    P_prime_exps.append(&mut gexp);
+
+    // compute exponent for h, i.e., -s, and add it to P_prime_exps
```
hamiidreza, line 438:
```suggestion
    // compute exponent for bold **h**, i.e., -s, and add it to P_prime_exps
```

rust-src/bulletproofs/src/set_non_membership_proof.rs
```diff
@@ -0,0 +1,674 @@
...
+        P_prime_exps.push(hexpi);
+    }
+
+    // add remaining exponents
+    P_prime_exps.push(tx);
+    P_prime_exps.push(minus_e_tilde);
+    P_prime_exps.push(C::Scalar::one());
+    P_prime_exps.push(x);
+
+    // P_prime_bases starts with G, H, and Q = g_hat
```
hamiidreza, line 451:
```suggestion
    // P_prime_bases starts with G=bold **g**, H=bold **h**, and Q = g_hat
```

rust-src/bulletproofs/src/set_non_membership_proof.rs
```diff
@@ -0,0 +1,674 @@
...
+    ) -> Commitment<G1> {
+        let v_value = Value::<SomeCurve>::new(v);
+        let v_com = v_keys.hide(&v_value, &v_rand);
+
+        v_com
+    }
+
+    #[test]
+    /// Test whether verifying an honestly generated proof works
+    fn test_smp_prove_verify() {
```
hamiidreza, line 515: should we call it 'test_snmp_prove_verify()' instead?

rust-src/bulletproofs/src/set_non_membership_proof.rs
```diff
@@ -0,0 +1,674 @@
...
+        // verify
+        let v_com = get_v_com(v, v_keys, v_rand);
+        let mut transcript = RandomOracle::empty();
+        let result = verify(&mut transcript, &the_set, &v_com, &proof, &gens, &v_keys);
+        assert!(result.is_ok());
+    }
+
+    /// Test that sets with sizes not a power of two work
+    #[test]
+    fn test_smp_prove_not_power_of_two() {
```
hamiidreza, line 538: test_snmp_prove_not_power_of_two()?

hamiidreza (approved):
All looks correct and also consistent with the bluepaper.
I just added some minor comments.

rust-src/bulletproofs/src/set_non_membership_proof.rs
tschudid: Yes

rust-src/bulletproofs/src/set_non_membership_proof.rs
hamiidreza: I checked and it looks fine.

PR #252, opened 2022-11-01
Wallet connect mobile lib adjustments

shjortConcordium:
## Purpose

Adjustments to mobile lib for wallet connect.

## Changes

Added a function for converting serialized parameters for smart contract updates (as bytes) into JSON.
Added a function `create_unsigned_transaction` for creating a PreAccountTransaction.

## Checklist

      hard-to-understand areas.


mobile_wallet/android/mobile_wallet_lib/src/main/java/com/concordium/mobile_wallet_lib/wallet.kt
```diff
@@ -20,8 +20,9 @@ external fun link_check(input: String) : String
 external fun get_identity_keys_and_randomness(input: String) : ReturnValue
 external fun get_account_keys_and_randomness(input: String) : ReturnValue
 external fun sign_transaction(input: String) : ReturnValue
-external fun transaction_to_json(input: String) : ReturnValue
+external fun parameters_to_json(input: String) : ReturnValue
```
abizjak, line 23: Isn't it one parameter only? So `parameter_to_json`?

mobile_wallet/android/mobile_wallet_lib/src/main/java/com/concordium/mobile_wallet_lib/wallet.kt
```diff
@@ -20,8 +20,9 @@ external fun link_check(input: String) : String
...
 external fun sign_message(input: String) : ReturnValue
+external fun create_unsigned_transaction(input: String) : ReturnValue
```
abizjak, line 25: Is there a reason we should not call this as we do internally, `create_pre_account_transaction`?

mobile_wallet/src/lib.rs
```diff
@@ -349,6 +325,98 @@ fn sign_message_aux(input: &str) -> anyhow::Result<String> {
     ))?)
 }
 
+#[derive(common::SerdeDeserialize)]
+struct MicroCCDAmount {
+    #[serde(rename = "microGtuAmount")]
+    amount: u64,
+}
```
abizjak, line 332: I am not sure why we need this? There is already an `Amount` type with the intended serialization in JSON.

mobile_wallet/src/lib.rs
```diff
@@ -349,6 +325,98 @@ fn sign_message_aux(input: &str) -> anyhow::Result<String> {
...
+
+#[derive(common::SerdeDeserialize)]
+struct AccountAddressJSON {
+    address: String,
+}
```
abizjak, line 337: We also have `AccountAddress` defined in `contracts_common` with the right serialziation.

mobile_wallet/src/lib.rs
```diff
@@ -349,6 +325,98 @@ fn sign_message_aux(input: &str) -> anyhow::Result<String> {
...
+
+#[derive(common::SerdeDeserialize)]
+#[serde(untagged)]
+enum JSONPayload {
+    UpdateContract {
+        amount:       MicroCCDAmount,
```
abizjak, line 343: This should just be `Amount`./

mobile_wallet/src/lib.rs
```diff
@@ -349,6 +325,98 @@ fn sign_message_aux(input: &str) -> anyhow::Result<String> {
...
+        #[serde(rename = "contractAddress")]
+        address:      ContractAddress,
+        #[serde(rename = "receiveName")]
+        receive_name: String,
```
abizjak, line 347:
```suggestion
        receive_name: common::OwnedReceiveName,
```

mobile_wallet/src/lib.rs
```diff
@@ -349,6 +325,98 @@ fn sign_message_aux(input: &str) -> anyhow::Result<String> {
...
+        parameter:    String, // Hex encoded bytes
+        #[serde(rename = "maxContractExecutionEnergy")]
+        energy:       u64,
```
abizjak, line 350: We have `Energy` as a type in concordium-base crate, which is availble in this crate.

mobile_wallet/src/lib.rs
```diff
@@ -349,6 +325,98 @@ fn sign_message_aux(input: &str) -> anyhow::Result<String> {
...
+
+/// Context to create an unsigned transaction.
+#[derive(common::SerdeDeserialize)]
+#[serde(rename_all = "camelCase")]
+struct TransactionContext {
+    pub from:             AccountAddressJSON,
+    pub expiry:           TransactionTime,
+    pub nonce:            Nonce,
+    #[serde(rename = "type")]
+    pub transaction_type: i32,
```
abizjak, line 366: What is this? There is a type called `TransactionType` in concordium-base, which has a reasonable JSON instance. So please use that.

abizjak: It is also unclear to me why this is needed since the `JSONPayload` is already an enum.

mobile_wallet/src/lib.rs
```diff
@@ -349,6 +325,98 @@ fn sign_message_aux(input: &str) -> anyhow::Result<String> {
...
+    amount: u64,
+}
+
+#[derive(common::SerdeDeserialize)]
+struct AccountAddressJSON {
+    address: String,
+}
+
+#[derive(common::SerdeDeserialize)]
+#[serde(untagged)]
```
abizjak, line 340: Instead of this you could do
```
#[serde(tag = "type", contents = "payload"]
```
to get the structure

```json
{
   "type": "UpdateContract",
   "payload": {
        "amount': ...,
        "address": {...},
   }
```

I would recommend using the same names as in the `TransactionType` though. That would be `Update` and `Transfer`

mobile_wallet/src/lib.rs
```diff
@@ -349,6 +325,98 @@ fn sign_message_aux(input: &str) -> anyhow::Result<String> {
...
+        amount:       MicroCCDAmount,
+        #[serde(rename = "contractAddress")]
+        address:      ContractAddress,
+        #[serde(rename = "receiveName")]
+        receive_name: String,
+        parameter:    String, // Hex encoded bytes
+        #[serde(rename = "maxContractExecutionEnergy")]
+        energy:       u64,
+    },
+    SimpleTransfer {
```
abizjak, line 352:
```suggestion
    Transfer {
```

mobile_wallet/src/lib.rs
```diff
@@ -349,6 +325,98 @@ fn sign_message_aux(input: &str) -> anyhow::Result<String> {
...
+        amount: MicroCCDAmount,
+        to:     AccountAddressJSON,
```
abizjak, line 354:
```suggestion
        amount: Amount,
        to:    AccountAddress,
```

abizjak (changes requested):

mobile_wallet/src/lib.rs
abizjak, line 343: Actually we already have a struct which is exactly what you need here I think. And has sensible JSON
```rust
UpdateContract {
    #[serde(flatten)]
    payload: base::UpdateContractPayload
}
```

mobile_wallet/src/lib.rs
shjortConcordium: Yeah, I guess we don't need it at all, we use it JS to know how the payload's structure shouldbe.

mobile_wallet/src/lib.rs
shjortConcordium: I wanted to do this, but the tag we set in JS is the number, not the transaction type name.
But I guess we can just not supply the type, because the payloads are distinct?

mobile_wallet/android/mobile_wallet_lib/src/main/java/com/concordium/mobile_wallet_lib/wallet.kt
shjortConcordium: I was not sure that the preAccountTransaction naming was something we used generally?

mobile_wallet/src/lib.rs
shjortConcordium: Yeah, except that the parameter/message fields have different names, but that might be more of a JS/me issue :thinking:

shjortConcordium: The correct thing would be to change the field name to message in javascript, so we are aligned. I guess.

mobile_wallet/android/mobile_wallet_lib/src/main/java/com/concordium/mobile_wallet_lib/wallet.kt
shjortConcordium: I changed it.

mobile_wallet/src/lib.rs
shjortConcordium: (I changed it to this)

rust-bins/wallet-notes/README.md
```diff
@@ -641,29 +641,49 @@ The returned value is a JSON object containing a list with signatures of the mes
...
 
-The returned value is a JSON object with the following fields:
+The returned value is a JSON representation of the parameter.
+
+An example input to this request is in the file [parameter_to_json-input.json](files/parameter_to_json-input.json).
+An example output to this request is in the file [parameter_to_json-output.json](files/parameter_to_json-output.json).
+
+## create_pre_account_transaction_ext
+
+Semantics: Creates a transfer transaction with the provided values.
```
abizjak, line 665: The second sentence is kind of in contradiction to the first.

rust-bins/wallet-notes/README.md
```diff
@@ -641,29 +641,49 @@ The returned value is a JSON object containing a list with signatures of the mes
...
+
+- `"expiry"` ... unix timestamp of the expiry date of the transaction.
+
+- `"nonce"` ... nonce of the sender account.
 
-- `"hashToSign"` ... hex encoding of the hash of the transaction
+- `"numKeys"` ... the number of keys that will be used to sign.
 
-- `"header"` ... the transaction header
+- `"type"` ... the transaction type tag (i.e. for a simple transfer the value should be "Transfer").
```
abizjak, line 679: And for contract update it should be ...

mobile_wallet/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 0.18.0
```
abizjak, line 3: Can you update the version in Cargo.toml & Cargo.lock to match this.

Then merge.

abizjak (approved):

PR #260, opened 2022-11-08
Add serialize_token_transfer-parameters to mobile_wallet_lib

shjortConcordium:
## Purpose

Add a function to serialize parameters for CIS-2 token transfers.

## Checklist

      hard-to-understand areas.


rust-bins/wallet-notes/README.md
```diff
@@ -676,6 +676,29 @@ The returned value is a JSON object with the following fields:
...
+This function takes as input a NUL-terminated UTF8-encoded string. The string
+must be a valid JSON object with fields
+
+- `"from"` ... address of the sender account.
+
+- `"to"` ... address of the receiver account or contract. This should be a JSON object with two fields:
+  
+  - `"type"` ... The value should be "AddressAccount" or "AddressContract" depending on which type the receiver is.
+
+  - `"address"` ... If the receiver is an account, this should be the base58 encoding of the account's address. If it is a contract, this field should be a JSON object with the fields `index` and `subindex`, both numbers.
```
abizjak, line 691: I have changed this (at least temporarily) so that `to` has to be an account address. 

If it is a contract address then there must also be an entrypoint supplied. That is invoked on that contract. We can keep it just an account for this version, otherwise we need to add a JSON/Serde instance for the `Receiver` type in base.

rust-bins/wallet-notes/files/serialize_token_transfer_parameters-input.json
```diff
@@ -0,0 +1,9 @@
+{
+    "tokenId": "10",
+    "amount": "1",
+    "from": "4EdBeGmpnQZWxaiig7FGEhWwmJurYmYsPWXo6owMDxA7ZtJMMH",
+    "to": {
+        "type":  "AddressAccount",
+        "address": "4EdBeGmpnQZWxaiig7FGEhWwmJurYmYsPWXo6owMDxA7ZtJMMH"
+    }
+}
```
abizjak, line 9: This no longer works.

rust-bins/wallet-notes/files/serialize_token_transfer_parameters-output.json
```diff
@@ -0,0 +1,3 @@
+{
+    parameter: "010001010100000000000000004dcc43ed4d3bd27e2c0918ee75b61e61807d77cf587f7b447996b983971a56f000aa05dd88c26078e6315d051fa1e80cf67e9fcec66709cc2688e959864cef2ca50000"
```
abizjak, line 2: This is probably incorrect now as well.

rust-bins/wallet-notes/files/serialize_token_transfer_parameters-output.json
```diff
@@ -0,0 +1,3 @@
+{
+    parameter: "0000000000000049010001100100aa05dd88c26078e6315d051fa1e80cf67e9fcec66709cc2688e959864cef2ca500aa05dd88c26078e6315d051fa1e80cf67e9fcec66709cc2688e959864cef2ca50000"
```
shjortConcordium, line 2: I've adjusted the input/output, but I don't think it is supposed to have the length prepended? :thinking: 
I guess that is the difference between the to_bytes method used.

abizjak: Yeah, the base16 should not be there. It is necessary that we use the contracts_common serial though since serialization of lengths inside the parameter, and integers, differs (it  is little endian).

abizjak (approved):

PR #262, opened 2022-11-09
Fix description of to parameter for serialize_token_transfer_parameters

shjortConcordium:
## Purpose

Fix mobile wallet lib readme :upside_down_face:


abizjak (approved):
I see you ignored my comment in the previous PR :)

PR #264, opened 2022-11-10
Add fourmolu formatting

limemloh:
## Purpose

This will add the [`fourmolu`](https://github.com/fourmolu/fourmolu) Haskell code formatter.
Related to https://github.com/Concordium/concordium-node/issues/599

## Changes

- Add a config for formatting `fourmolu.yaml`
- Add CI job checking formatting on PRs.
- Update readme related to formatter for Haskell.
- Run formatter on code base.


.github/workflows/fourmolu.yaml
```diff
@@ -0,0 +1,44 @@
+# Workflow triggered by pushes to the main branch and PRs targeting it
```
abizjak, line 1: Can we specify that the build job should wait for this to complete?

The build job is very resource intensive and we should ideally not run it if formatting fails.

abizjak (approved):

.github/workflows/fourmolu.yaml
limemloh: I'll address this in another PR

PR #272, opened 2022-11-21
Mark functions that only apply to sending p1-p3 transactions as deprecated

abizjak:
## Purpose

Make it less confusing to use baker related transactions in the rust SDK.

## Changes

Mark baker related transactions that only apply to P1-P3 as deprecated, and hide them from generated documentation.

## Checklist

      hard-to-understand areas.


limemloh (approved):
Which version is this referring to? The `concordium-base` crate is version 0.1.0

abizjak:
> Which version is this referring to? The `concordium-base` crate is version 0.1.0

It is referring to rust-sdk version since that is where these are re-exported from.

PR #283, opened 2022-12-14
Upgrade compiler tooling and libraries.

vikt0r0:
## Purpose

Upgrade stackage snapshot to `lts-20.4` and ghc to 9.2.5.

## Changes

Change `resolver`in `stack.yaml` and fix build errors and warnings.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-tests/Types/PayloadSerializationSpec.hs
+++ b/haskell-tests/Types/PayloadSerializationSpec.hs
@@ -60,12 +60,14 @@ checkPayload spv e =
             Left err -> counterexample err False
             Right e' -> label (groupIntoSize (fromIntegral (BS.length bs))) $ e === e'
 
+-- Modify the bitmap portion of a payload. Assumes that the input bytestring
+-- is at least 24 bits long.
 modifyPayloadBitmap :: (Word16 -> Word16) -> BS.ByteString -> BS.ByteString
 modifyPayloadBitmap f bs =
     let res = S.runGetState ((,) <$> S.getWord8 <*> S.getWord16be) bs 0
         ((header, bitmap), rest) = case res of
             Right v -> v
-            -- This does not happen
+            -- This happens only when bs is 23 bits or less.
             Left _ -> error "res should be Right"
     in  S.runPut (S.putWord8 header <> S.putWord16be (f bitmap)) `BS.append` rest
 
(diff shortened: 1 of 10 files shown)


haskell-tests/Types/PayloadSerializationSpec.hs
```diff
@@ -62,7 +62,11 @@ checkPayload spv e =
 
 modifyPayloadBitmap :: (Word16 -> Word16) -> BS.ByteString -> BS.ByteString
 modifyPayloadBitmap f bs =
-    let Right ((header, bitmap), rest) = S.runGetState ((,) <$> S.getWord8 <*> S.getWord16be) bs 0
+    let res = S.runGetState ((,) <$> S.getWord8 <*> S.getWord16be) bs 0
+        ((header, bitmap), rest) = case res of
+            Right v -> v
+            -- This does not happen
```
abizjak, line 68: That is only true if the input bytestring has at least 24 bytes. State that as a precondition I suppose.

abizjak (approved):
Looks good. But we should not merge this before the other PRs are ready.

EmilLa1 (approved):
Looks good 👍 
I think the outstanding comment should be addressed before merging.

haskell-tests/Types/PayloadSerializationSpec.hs
```diff
@@ -60,9 +60,15 @@ checkPayload spv e =
...
 
+-- Modify the bitmap portion of a payload. Assumes that the input bytestring
+-- is at least 24 bits long.
 modifyPayloadBitmap :: (Word16 -> Word16) -> BS.ByteString -> BS.ByteString
 modifyPayloadBitmap f bs =
-    let Right ((header, bitmap), rest) = S.runGetState ((,) <$> S.getWord8 <*> S.getWord16be) bs 0
+    let res = S.runGetState ((,) <$> S.getWord8 <*> S.getWord16be) bs 0
+        ((header, bitmap), rest) = case res of
+            Right v -> v
+            -- This happens only when bs is 23 bits or less.
```
abizjak, line 70:
```suggestion
            -- This happens only when bs is 23 bytes or less.
```

vikt0r0: No, 3 bytes right?

abizjak: Yes, 3 bytes.  Better to use bytes than bits since that is what we are reading.

PR #288, opened 2022-12-22
Consensus chain parameters

EmilLa1:
## Purpose

Introduce new chain parameters for supporting the new consensus protocol. We also adopt the [singletons library](https://hackage.haskell.org/package/singletons) and broadly use that where we have previously defined singletons directly.

Addresses https://github.com/Concordium/concordium-node/issues/638 together with
https://github.com/Concordium/concordium-node/pull/653 

## Changes

Changes to the parameters:
- Election difficulty is generalised to "consensus parameters". For <P6, this is just the election difficulty. For P6, this consists of:
  - Timeout parameters: Timeout base time, timeout increase factor, timeout decrease factor.
  - Minimum block time.
  - Block energy limit.
- The consensus parameters are updatable with the same keys as for the election difficulty:
  - The access structure `asParamElectionDifficulty` is renamed to `asParamConsensusParameters`.
  - For JSON serialization of the level 2 authorizations, `electionParameters` is still used. This is to maintain compatibility, and since the old GRPC API is deprecated.
  - (It is renamed to `parameter_consensus` in the GRPC2 API. Renaming does not break binary compatibility.)
  - Each of the three new consensus parameters is updatable independently, though with the same keys.
- The GAS reward for including finalization proofs has been removed.

Technical changes:
- Use the singletons library to construct the existing singletons for `ProtocolVersion`, `ChainParametersVersion`, `AccountVersion` and `TransactionOutcomesVersion`. Generally, the same interface is maintained, however:
  - The constructors of `SChainParametersVersion` are renamed, e.g. `SCPV0` to `SChainParametersV0`.
  - Where the `IsProtocolVersion` constraint is used, the module should use `{-# LANGUAGE MonoLocalBinds #-}` (or `TypeFamilies` or `GADTs`, which imply this) to avoid warnings.
- `JustForCPV1` is removed. It is replaced with two new types:
  - `Conditionally`, with constructors `CTrue` and `CFalse`, holds a value if the type-level boolean parameters is `True`.
  - `OParam`, with constructors `SomeParam` and `NoParam`, holds a value if the chain parameters version supports the given parameter type.
- A number of new versioning types that are dependent on `ChainParametersVersion` are introduced for fine-grained control of version specific parameters:
  - `MintDistributionVersion`:
    - `MintDistributionVersion0` (`ChainParametersV0`): supports mint per slot.
    - `MintDistributionVersion1` (`ChainParametersV1`, `ChainParametersV2`): does not support mint per slot.
  - `GASRewardsVersion`:
    - `GASRewardsVersion0` (`ChainParametersV0`, `ChainParametersV1`): supports GAS reward for finalization proofs.
    - `GASRewardsVersion1` (`ChainParametersV2`): does not support GAS reward for finalization proofs.
  - `CooldownParametersVersion`:
    - `CooldownParametersVersion0` (`ChainParametersV0`): baker cooldown specified in epochs.
    - `CooldownParametersVersion1` (`ChainParametersV1`, `ChainParametersV2`): baker and delegator cooldowns specified in seconds.
  - `PoolParametersVersion`:
    - `PoolParametersVersion0`: (`ChainParametersV0`): minimum stake for registering a baker.
    - `PoolParametersVersion1` (`ChainParametersV1`, `ChainParametersV2`):
      - passive commission rates
      - bounds on pool commission rates
      - minimum baker equity capital
      - leverage bound for a baker
  - `ConsensusParametersVersion`:
    - `ConsensusParametersVersion0` (`ChainParametersV0`, `ChainParametersV1`): election difficulty
    - `ConsensusParametersVersion1` (`ChainParametersV2`):
        - Timeout parameters
        - Minimum block time
        - Block energy limit
  - `AuthorizationsVersion`:
    - `AuthorizationsVersion0` (`ChainParametersV0`).
    - `AuthorizationsVersion1` (`ChainParametersV1`, `ChainParametersV2`): add access structures for cooldown parameters and time parameters.

## Checklist

      hard-to-understand areas.


haskell-src/Concordium/Genesis/Parameters.hs
```diff
@@ -91,17 +92,54 @@ parseJSONForGCPV1 =
...
             _ppPassiveCommissions = CommissionRates{..}
             _ppCommissionBounds = CommissionRanges{..}
         return GenesisChainParameters{..}
 
+-- |Parse 'GenesisChainParameters' from JSON for 'ChainParametersV2'.
+parseJSONForGCPV2 :: Value -> Parser (GenesisChainParameters' 'ChainParametersV2)
+parseJSONForGCPV2 =
+    withObject "GenesisChainParametersV2" $ \v -> do
+        _erEuroPerEnergy <- v .: "euroPerEnergy"
+        _erMicroGTUPerEuro <- v .: "microGTUPerEuro"
```
chrmatt, line 107: Would it be possible/a good time to replace GTU by CCD at least for the new things we do?

chrmatt:
It seems this does not include the parameters for which bakers are finalizers we have discussed. Will this be part of another PR?

EmilLa1:
@chrmatt Yes. There is an issue open on it: https://github.com/Concordium/concordium-base/issues/303. We will start on that later when it's specified.

td202 (approved):

limemloh (approved):
Looks good, I have mainly comments nitpicking on naming and documentation.

haskell-src/Concordium/Types/UpdateQueues.hs
```diff
@@ -44,17 +43,7 @@ data UpdateQueue e = UpdateQueue
...
-
-putUpdateQueueForCPV1 :: (Serialize e) => Putter (UpdateQueueForCPV1 cpv e)
-putUpdateQueueForCPV1 NothingForCPV1 = return ()
-putUpdateQueueForCPV1 (JustForCPV1 uq) = putUpdateQueueV0 uq
-
-getUpdateQueueForCPV1 :: forall cpv e. (Serialize e, IsChainParametersVersion cpv) => Get (UpdateQueueForCPV1 cpv e)
-getUpdateQueueForCPV1 = case chainParametersVersion @cpv of
-    SCPV0 -> return NothingForCPV1
-    SCPV1 -> JustForCPV1 <$> getUpdateQueueV0
+type OUpdateQueue (pt :: ParameterType) (cpv :: ChainParametersVersion) e = OParam pt cpv (UpdateQueue e)
```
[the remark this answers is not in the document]
EmilLa1: O = Optional

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -20,38 +32,253 @@
...
-    SP2 :: SProtocolVersion 'P2
-    SP3 :: SProtocolVersion 'P3
-    SP4 :: SProtocolVersion 'P4
-    SP5 :: SProtocolVersion 'P5
-    SP6 :: SProtocolVersion 'P6
+import Data.Word
 
+$( singletons
+    [d|
+        -- \|An enumeration of the supported versions of the consensus protocol.
```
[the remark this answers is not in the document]
EmilLa1: One cannot generate haddoc documentation within template haskell, so that is why it's in the header and export list. 
The reason for having it here in the template haskell part is to also provide doc when e.g. jumping to definition.

I admit it's a bit duplication, but I think it's worth it?

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -27,42 +397,280 @@ import qualified Concordium.Crypto.SHA256 as Hash
...
+        data MintDistributionVersion
+            = MintDistributionVersion0
+            | MintDistributionVersion1
+
+        mintDistributionVersionFor :: ChainParametersVersion -> MintDistributionVersion
+        mintDistributionVersionFor ChainParametersV0 = MintDistributionVersion0
+        mintDistributionVersionFor ChainParametersV1 = MintDistributionVersion1
+        mintDistributionVersionFor ChainParametersV2 = MintDistributionVersion1
+
+        supportsMintPerSlot :: MintDistributionVersion -> Bool
```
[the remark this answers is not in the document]
EmilLa1: That is kind of what we did before. The problem with that approach is that (with the new introduced versions, parameter types etc.) it would become quite a lot of boiler plate code.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -27,42 +397,280 @@ import qualified Concordium.Crypto.SHA256 as Hash
...
+maybeWhenSupported b _ NoParam = b
+maybeWhenSupported _ f (SomeParam a) = f a
+
+-- * Mint distribution
+
+-- |Constraint on a type level 'MintDistributionVersion' that can be used to get a corresponding
+-- 'SMintDistributionVersion'. An alias for 'SingI'.
+type IsMintDistributionVersion (mdv :: MintDistributionVersion) = SingI mdv
+
+-- |Witness a @SingI (SupportsMintPerSlot mdv)@ constraint using a 'SMintDistributionVersion mdv'.
```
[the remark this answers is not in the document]
EmilLa1: Added some doc.

haskell-src/Concordium/Types/Updates.hs
```diff
@@ -716,6 +714,9 @@ instance Serialize UpdateType where
     put UpdateAddIdentityProvider = putWord8 14
     put UpdateCooldownParameters = putWord8 15
     put UpdateTimeParameters = putWord8 16
+    put UpdateTimeoutParameters = putWord8 17
+    put UpdateMinBlockTime = putWord8 19
```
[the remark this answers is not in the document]
EmilLa1: Great spotted. I think this stems from me having a hard time counting when I'm out of fingers :) .

haskell-src/Concordium/Types/ProtocolVersion.hs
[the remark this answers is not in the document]
td202: The reason for having it here is more so you can see what it's meant to do when you're reading the code, rather than having to look elsewhere.

haskell-src/Concordium/Types/Parameters.hs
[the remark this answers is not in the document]
limemloh: Just to be clear, my comment is not about using the singleton library, but more of using template haskell for generating this part, instead of doing it manually and still using the library.
I don't have a good grasp of how much work and whether it would be worth it

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -1,23 +1,393 @@
...
+    sCooldownParametersVersionFor,
+    IsCooldownParametersVersion,
+    withIsCooldownParametersVersionFor,
+    CooldownParameters' (..),
+    CooldownParameters,
+    cpBakerExtraCooldownEpochs,
+    cpPoolOwnerCooldown,
+    cpDelegatorCooldown,
+    putCooldownParameters,
+    getCooldownParameters,
```
abizjak, line 209: Why are these not documented?

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -1,23 +1,393 @@
...
+
+    -- * Defunctionalisation symbols
+    PTElectionDifficultySym0,
+    PTTimeParametersSym0,
+    PTMintPerSlotSym0,
+    PTTimeoutParametersSym0,
+    PTMinBlockTimeSym0,
+    PTBlockEnergyLimitSym0,
+    PTCooldownParametersAccessStructureSym0,
+    PTFinalizationProofSym0,
```
abizjak, line 380: Explain in detail and on examples what defunctionalization symbols are. What are they for, what problems do they solve.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -27,42 +397,313 @@ import qualified Concordium.Crypto.SHA256 as Hash
...
--- |Lens for '_mpsMintPerSlot'
-{-# INLINE mpsMintPerSlot #-}
-mpsMintPerSlot :: Lens' (MintPerSlotForCPV0 'ChainParametersV0) MintRate
-mpsMintPerSlot =
-    lens _mpsMintPerSlot (\mps x -> mps{_mpsMintPerSlot = x})
-
-deriving instance Eq (MintPerSlotForCPV0 cpv)
-deriving instance Show (MintPerSlotForCPV0 cpv)
+-- The Template Haskell below generates the quoted definitions, together with the singletons
+-- and liftings to singletons. The module header is used to give the haddock documentation.
```
abizjak, line 405: Explain what exactly is being generated, e.g., what is a "quoted definition"?

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -27,42 +397,313 @@ import qualified Concordium.Crypto.SHA256 as Hash
...
+        isSupported PTBlockEnergyLimit ChainParametersV2 = True
+        isSupported PTCooldownParametersAccessStructure cpv = supportsCooldownParametersAccessStructure (authorizationsVersionFor cpv)
+        isSupported PTFinalizationProof ChainParametersV0 = True
+        isSupported PTFinalizationProof ChainParametersV1 = True
+        isSupported PTFinalizationProof ChainParametersV2 = False
+        |]
+ )
+
+-- |Constraint on a type level 'ParameterType' that can be used to get a corresponding
+-- 'SParameterType'. An alias for 'SingI'.
```
abizjak, line 541: I don't think it's useful to state that this is an alias for something. That's clear from the definition.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -27,42 +397,313 @@ import qualified Concordium.Crypto.SHA256 as Hash
...
+    CFalse == CFalse = True
+    CTrue a == CTrue b = a == b
+
+instance (Ord a) => Ord (Conditionally b a) where
+    compare CFalse CFalse = EQ
+    compare (CTrue x) (CTrue y) = compare x y
+
+instance (Show a) => Show (Conditionally b a) where
+    show CFalse = "<CFalse>"
+    show (CTrue v) = show v
```
abizjak, line 589: That's not clear `(show (CTrue CFalse))`. Please add some indicator to indicate the variant.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -27,42 +397,313 @@ import qualified Concordium.Crypto.SHA256 as Hash
...
+
+instance (Serialize a, SingI b) => Serialize (Conditionally b a) where
+    put CFalse = return ()
+    put (CTrue a) = put a
+
+    get = case sing @b of
+        SFalse -> pure CFalse
+        STrue -> CTrue <$> get
+
+-- |Wrap a value in a 'Conditionally' depending on the supplied 'SBool'.
```
abizjak, line 599: Please explain where SBool is coming from. At the TH where it is generated.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -27,42 +397,313 @@ import qualified Concordium.Crypto.SHA256 as Hash
...
+
+-- |A lens for accessing the contents of a 'Conditionally' when the guard is known to be 'True'.
+unconditionally :: Lens (Conditionally 'True a) (Conditionally 'True b) a b
+unconditionally f (CTrue a) = CTrue <$> f a
+
+-- |An @OParam pt cpv a@ is an @a@ if the parameter type @pt@ is supported at @cpv@, and @()@
+-- otherwise.
+data OParam (pt :: ParameterType) (cpv :: ChainParametersVersion) a where
+    NoParam :: (IsSupported pt cpv ~ 'False) => OParam pt cpv a
+    SomeParam :: (IsSupported pt cpv ~ 'True) => !a -> OParam pt cpv a
```
abizjak, line 619: Why is this a separate type, and not just an alias for Conditionally (IsSupported pt cpv). Please explain.

haskell-src/Concordium/Types/Parameters.hs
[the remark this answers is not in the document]
abizjak: This comment should be provided to all these "with*" functions. Not just this one.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -27,42 +397,313 @@ import qualified Concordium.Crypto.SHA256 as Hash
...
+
+-- * Mint distribution
+
+-- |Constraint on a type level 'MintDistributionVersion' that can be used to get a corresponding
+-- 'SMintDistributionVersion'. An alias for 'SingI'.
+type IsMintDistributionVersion (mdv :: MintDistributionVersion) = SingI mdv
+
+-- |Witness a @SingI (SupportsMintPerSlot mdv)@ constraint using a 'SMintDistributionVersion mdv'.
+-- Concretely this provides the passed in action @a@ with the context 'SupportsMintPerSlot mdv'.
+withSupportsMintPerSlot :: forall mdv a. SMintDistributionVersion mdv -> (SingI (SupportsMintPerSlot mdv) => a) -> a
```
abizjak, line 692: Why is this one without an alias (i.e., why is it using SingI) explicitly.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -158,14 +799,28 @@ instance HashableTo Hash.Hash TransactionFeeDistribution where
 
 instance Monad m => MHashableTo m Hash.Hash TransactionFeeDistribution
 
+-- * GAS rewards
+
+-- |Constraint on a type level 'GASRewardsVersion' that can be used to get a corresponding
+-- 'SGASRewardsVersion'. An alias for 'SingI'.
```
abizjak, line 805: "An alias for singI" does not provide value to the reader. That's clear from the definition.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -307,9 +987,23 @@ instance HasExchangeRates ExchangeRates where
     {-# INLINE energyRate #-}
     energyRate = to _erEnergyRate
 
+-- * Cooldown parameters
+
+-- |Constraint on a type level 'CooldownParametersVersion' that can be used to get a corresponding
+-- 'SCooldownParametersVersion'. An alias for 'SingI'.
```
abizjak, line 993: Alias again.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -340,65 +1037,76 @@ instance ToJSON (CooldownParameters cpv) where
...
+parseCooldownParametersJSON = case sing @cpv of
+    SCooldownParametersVersion0 -> withObject "CooldownParametersV0" $ \v -> CooldownParametersV0 <$> v .: "bakerCooldownEpochs"
+    SCooldownParametersVersion1 -> withObject "CooldownParametersV1" $ \v ->
         CooldownParametersV1
-            <$> v .: "poolOwnerCooldown"
-            <*> v .: "delegatorCooldown"
+            <$> v
+                .: "poolOwnerCooldown"
+            <*> v
+                .: "delegatorCooldown"
```
abizjak, line 1052: I don't understand why you felt the need to add newlines.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -407,59 +1115,47 @@ data TimeParameters (cpv :: ChainParametersVersion) where
...
-    lens _tpRewardPeriodLength (\tp x -> tp{_tpRewardPeriodLength = x})
+        TimeParameters
+    deriving (Eq, Show)
 
--- |Lens for '_tpMintPerPayday'
-{-# INLINE tpMintPerPayday #-}
-tpMintPerPayday :: Lens' (TimeParameters 'ChainParametersV1) MintRate
-tpMintPerPayday =
-    lens _tpMintPerPayday (\tp x -> tp{_tpMintPerPayday = x})
+makeClassy ''TimeParameters
```
abizjak, line 1121: Explain in detail waht functions and classes this generates. The documentation of makeClassy is poor.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -687,34 +1404,184 @@ putPoolParameters PoolParametersV1{..} = do
...
-deriving instance Eq (PoolParameters cpv)
-deriving instance Show (PoolParameters cpv)
+instance Serialize TimeoutParameters where
+    put TimeoutParameters{..} = do
+        put tpTimeoutBase
+        put tpTimeoutIncrease
+        put tpTimeoutDecrease
+    get = do
+        tpTimeoutBase <- get
+        tpTimeoutIncrease <- get
```
abizjak, line 1445: We are deviating from what we do in other such cases by using an instance that has implicit endianness.

Moreover in contrast to other cases with ratios, where we check that numerator and denominator are coprime we do not do so here. Which I would consider a bug. Same for decrease.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -687,34 +1404,184 @@ putPoolParameters PoolParametersV1{..} = do
...
+              "timeoutIncrease" AE..= tpTimeoutIncrease,
+              "timeoutDecrease" AE..= tpTimeoutDecrease
+            ]
+
+instance FromJSON TimeoutParameters where
+    parseJSON = withObject "TimeoutParameters" $ \o -> do
+        tpTimeoutBase <- o .: "timeoutBase"
+        tpTimeoutIncrease <- o .: "timeoutIncrease"
+        unless (tpTimeoutIncrease > 1) $ fail "timeoutIncrease must be greater than 1."
+        tpTimeoutDecrease <- o .: "timeoutDecrease"
```
abizjak, line 1465: Similar issue with coprime test than above.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -687,34 +1404,184 @@ putPoolParameters PoolParametersV1{..} = do
...
+          -- |Maximum energy allowed per block.
+          _cpBlockEnergyLimit :: !Energy
+        } ->
+        ConsensusParameters' 'ConsensusParametersVersion1
+
+-- |Convenience type for a 'ConsensusParameters'' parametrised by the 'ChainParametersVersion'.
+type ConsensusParameters (cpv :: ChainParametersVersion) =
+    ConsensusParameters' (ConsensusParametersVersionFor cpv)
+
+-- |Lens for '_cpElectionDifficulty'
```
abizjak, line 1506: Could we try to explain things in non-jargon words. It is clear taht this is a Lens. That's what the type says.

But if this comment was to be useful it needs to add something so that non-experts can also understand.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -744,19 +1611,30 @@ instance HasExchangeRates (ChainParameters' cpv) where
...
+getChainParameters =
+    ChainParameters
+        <$> withIsConsensusParametersVersionFor (chainParametersVersion @cpv) get
+        <*> get
+        <*> withIsCooldownParametersVersionFor (chainParametersVersion @cpv) get
+        <*> get
+        <*> get
+        <*> get
+        <*> get
+        <*> withIsPoolParametersVersionFor (chainParametersVersion @cpv) get
```
abizjak, line 1637: Please don't do this. Name the variables you are parsing. This is always so hard to maintain since you have to count the order to figure out what is what.

haskell-src/Concordium/Genesis/Parameters.hs
EmilLa1: Yes. But this is actually just reusing the existing `ExchangeRates` type we have defined. We can change it if we want.. But I'm not so sure we want to do that as it will lead to more "needless" work later on for dependencies. 

I suggest we do this in a subsequent task in order to keep this feature somewhat focused on the actual new consensus protocol. @abizjak can perhaps chime in on his thoughts?

haskell-src/Concordium/Types/Parameters.hs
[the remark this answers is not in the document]
EmilLa1: Resolving this. I will add more documentation for the splice and what it generates. There's also another comment pointing out this issue.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -687,34 +1367,184 @@ putPoolParameters PoolParametersV1{..} = do
...
+        SConsensusParametersVersion1 -> do
+            _cpTimeoutParameters <- get
+            _cpMinBlockTime <- get
+            _cpBlockEnergyLimit <- get
+            return ConsensusParametersV1{..}
+
+-- * Chain parameters
+
+-- |Witness the constraints implied by an 'SChainParametersVersion'.
+withCPVConstraints ::
```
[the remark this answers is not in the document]
EmilLa1: Hm. I'm unsure what you mean here? What other functions? All the functions that provides a context for an action @a@ is currently prefixed with `withFooBar` ? 

I have documented the function a bit more with what it concretely provides the user.

haskell-src/Concordium/Genesis/Parameters.hs
```diff
@@ -91,17 +92,54 @@ parseJSONForGCPV1 =
...
+        _ppMinimumEquityCapital <- v .: "minimumEquityCapital"
+        _ppCapitalBound <- v .: "capitalBound"
+        _ppLeverageBound <- v .: "leverageBound"
+        _tpRewardPeriodLength <- v .: "rewardPeriodLength"
+        _tpMintPerPayday <- v .: "mintPerPayday"
+        tpTimeoutBase <- v .: "timeoutBase"
+        tpTimeoutIncrease <- v .: "timeoutIncrease"
+        tpTimeoutDecrease <- v .: "timeoutDecrease"
+        _cpMinBlockTime <- v .: "minBlockTime"
+        _cpBlockEnergyLimit <- v .: "blockEnergyLimit"
```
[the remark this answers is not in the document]
EmilLa1: No, not except that is currently what is being done for cpv1. https://github.com/Concordium/concordium-base/blob/2223fa461ceef8a0203926e206f211a8c68de922/haskell-src/Concordium/Genesis/Parameters.hs#L114

haskell-src/Concordium/Types/Migration.hs
```diff
@@ -73,6 +73,22 @@ migratePoolParameters (StateMigrationParametersP3ToP4 migration) _ =
     P4.updatePoolParameters (P4.migrationProtocolUpdateData migration)
 migratePoolParameters StateMigrationParametersP4ToP5 poolParams = poolParams
 
+-- |Apply a state migration to a 'GASRewards' structure.
+--
+-- [P4 to P5]: the finalization proof reward is removed.
```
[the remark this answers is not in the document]
EmilLa1: No. I removed this comment for now. It was meant for being P5->P6. 

The migration will be handled in an upcoming PR which will address the protocol update.

haskell-src/Concordium/Types/Migration.hs
```diff
@@ -73,6 +73,22 @@ migratePoolParameters (StateMigrationParametersP3ToP4 migration) _ =
...
+    StateMigrationParameters oldpv pv ->
+    GASRewards (GasRewardsVersionFor (ChainParametersVersionFor oldpv)) ->
+    GASRewards (GasRewardsVersionFor (ChainParametersVersionFor pv))
+migrateGASRewards StateMigrationParametersTrivial gr = gr
+migrateGASRewards StateMigrationParametersP1P2 gr = gr
+migrateGASRewards StateMigrationParametersP2P3 gr = gr
+migrateGASRewards StateMigrationParametersP3ToP4{} gr = gr
+migrateGASRewards StateMigrationParametersP4ToP5 gr = gr
+
+-- FIXME: Add P5 -> P6
```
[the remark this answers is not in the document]
EmilLa1: Yes. So was the comment above. 
I removed the comment for now so in order to avoid needless confusion.

haskell-src/Concordium/Types/Queries.hs
```diff
@@ -266,9 +279,12 @@ instance FromJSON BlockSummary where
...
             BlockSummary
-                <$> v .: "transactionSummaries"
-                <*> v .: "specialEvents"
-                <*> v .: "finalizationData"
+                <$> v
+                    .: "transactionSummaries"
+                <*> v
+                    .: "specialEvents"
+                <*> v
+                    .: "finalizationData"
```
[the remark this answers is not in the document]
EmilLa1: Changed it back.

haskell-src/Concordium/Types/Updates.hs
```diff
@@ -836,31 +848,53 @@ putUpdatePayload (CooldownParametersCPV1UpdatePayload u) = putWord8 14 >> putCoo
...
+        14
+            | CooldownParametersVersion1 <- cooldownParametersVersionFor cpv ->
+                CooldownParametersCPV1UpdatePayload <$> getCooldownParameters SCooldownParametersVersion1
+        15
+            | PoolParametersVersion1 <- poolParametersVersionFor cpv ->
+                PoolParametersCPV1UpdatePayload <$> getPoolParameters SPoolParametersVersion1
+        16
+            | isSupported PTTimeParameters cpv ->
+                TimeParametersCPV1UpdatePayload <$> getTimeParameters
+        17
```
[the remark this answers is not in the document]
EmilLa1: Fourmolu.

haskell-src/Concordium/Types/Updates.hs
```diff
@@ -836,31 +848,53 @@ putUpdatePayload (CooldownParametersCPV1UpdatePayload u) = putWord8 14 >> putCoo
...
+putUpdatePayload (MinBlockTimeUpdatePayload u) = putWord8 19 >> put u
+putUpdatePayload (BlockEnergyLimitUpdatePayload u) = putWord8 20 >> put u
+putUpdatePayload (GASRewardsCPV2UpdatePayload u) = putWord8 21 >> put u
 
 getUpdatePayload :: SProtocolVersion pv -> Get UpdatePayload
 getUpdatePayload spv =
     getWord8 >>= \case
         1 -> ProtocolUpdatePayload <$> get
-        2 -> ElectionDifficultyUpdatePayload <$> get
+        2
```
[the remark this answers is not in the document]
EmilLa1: Fourmolu.

haskell-src/Concordium/Types/Updates.hs
```diff
@@ -180,21 +185,21 @@ data Authorizations cpv = Authorizations
       -- |Parameter keys: IdentityProviderIdentity and IpInfo
       asAddIdentityProvider :: !AccessStructure,
       -- |Parameter keys: Cooldown periods for pool owners and delegators
-      asCooldownParameters :: !(AccessStructureForCPV1 cpv),
+      asCooldownParameters :: !(Conditionally (SupportsCooldownParametersAccessStructure auv) AccessStructure),
       -- |Parameter keys: Length of reward period / payday
-      asTimeParameters :: !(AccessStructureForCPV1 cpv)
+      asTimeParameters :: !(Conditionally (SupportsTimeParameters auv) AccessStructure)
```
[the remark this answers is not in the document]
EmilLa1: True. But they also reflect different things. 
`SupportsTimeParameters` reflects that the time parameters were introduced to the set of chain parameters, while `SupportsCooldownParametersAccessStructure` reflects when it became _possible_ to actually update the cooldown parameters, hence the cooldown parameters were existing **before** it became possible to update them. 

At least that is how I understand the flow of things - and why they are named like they are. @td202 Can you correct me if I'm wrong?

haskell-src/Concordium/Types/Parameters.hs
EmilLa1: I have added an alias for a `SingI` i.e. ` type MintPerSlotSupported (mdv :: MintDistributionVersion) = SingI (SupportsMintPerSlot mdv) ` in order to make things consistent.

haskell-src/Concordium/Types/Parameters.hs
EmilLa1: There are documentation at the actual definitions. Is this not adequate? It seems to me that duplicating the stubs of documentation will just lead to an unnecessary overhead when maintaining the code base.

EmilLa1: I have put a comment at top stating that TH definitions are documented in the header of the module while the regular definitions are simply documented where they are defined.

haskell-src/Concordium/Types/Parameters.hs
EmilLa1: It's a predefined type coming from `Data.Bool.Singletons`

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -20,38 +32,253 @@
...
+    ProtocolVersion (..),
+    -- |The singleton type associated with 'ProtocolVersion'.
+    -- There is a unique constructor of 'SProtocolVersion' for
+    -- each constructor of 'ProtocolVersion'.
+    SProtocolVersion (..),
+    IsProtocolVersion,
+    protocolVersion,
+    SomeProtocolVersion (..),
+    promoteProtocolVersion,
+    demoteProtocolVersion,
```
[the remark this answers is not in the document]
EmilLa1: TH generated definitions are documented here in the header of the module while the regular definitions are documented where they are defined. I've added a line about this in the top of the file.

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -20,38 +32,253 @@
...
+    --         * Minimum block time.
+    --         * Block energy limit.
+    --
+    --     - The finalization proof reward is removed from the GAS rewards.
+    ChainParametersVersion (..),
+    -- |Singleton type corresponding to 'ChainParametersVersion'.
+    SChainParametersVersion (..),
+    IsChainParametersVersion,
+    chainParametersVersion,
+    demoteChainParameterVersion,
```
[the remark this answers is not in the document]
EmilLa1: Same comment as above.

haskell-src/Concordium/Types/Parameters.hs
EmilLa1: Added a comment.

haskell-src/Concordium/Genesis/Parameters.hs
abizjak: We cannot change JSON. I also agree we should not be renaming stuff in this PR.

abizjak: We "can", but what that means is that all consumers of the API now have to support the two versions. It's not hard to do this, but it is work.

Since V1 API is in maintenance mode we are not going to be spending time on this.

haskell-src/Concordium/Types/Parameters.hs
EmilLa1: Fix inc

haskell-src/Concordium/Types/Parameters.hs
EmilLa1: Fix inc

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -20,38 +32,253 @@
...
+    supportsDelegation,
+    -- |Determine whether delegation is supported for a particular account version (at the type level).
+    SupportsDelegation,
+    -- |Determine whether delegation is supported for a particular account version (on singletons).
+    sSupportsDelegation,
+    SAVDelegationSupport (..),
+    AVSupportsDelegation,
+    PVSupportsDelegation,
+    delegationSupport,
+    protocolSupportsDelegation,
```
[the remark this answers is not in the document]
EmilLa1: TH generated haskell code is defined in the head of the module while regular definitions are defined on site.

haskell-src/Concordium/Types/Parameters.hs
```diff
@@ -27,42 +397,280 @@ import qualified Concordium.Crypto.SHA256 as Hash
...
+
+-- |Wrap a value in a 'Conditionally' depending on the supplied 'SBool'.
+conditionally :: SBool b -> a -> Conditionally b a
+conditionally SFalse _ = CFalse
+conditionally STrue a = CTrue a
+
+-- |Perform an action conditionally on the supplied 'SBool'.
+-- This is typically used for monadic actions.
+-- The action is not performed in the 'SFalse' case.
+conditionallyA :: (Applicative f) => SBool b -> f a -> f (Conditionally b a)
```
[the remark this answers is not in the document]
EmilLa1: I renamed for consistency. pureWhenSupported -> whenSupported and the old whenSupported -> whenSupportedA now.

haskell-src/Concordium/Types/Parameters.hs
EmilLa1: I have added insightful comments for this now :-).... Thanks.

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -20,38 +32,253 @@
...
-data SProtocolVersion (pv :: ProtocolVersion) where
-    SP1 :: SProtocolVersion 'P1
-    SP2 :: SProtocolVersion 'P2
-    SP3 :: SProtocolVersion 'P3
-    SP4 :: SProtocolVersion 'P4
-    SP5 :: SProtocolVersion 'P5
-    SP6 :: SProtocolVersion 'P6
+import Data.Word
 
+$( singletons
```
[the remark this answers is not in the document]
EmilLa1: There's a pointer to the documentation for the splice in `Parameters.hs`.

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -20,38 +32,253 @@
...
+
+    -- * Feature guards
+    supportsMemo,
+    supportsV1Contracts,
+    supportsUpgradableContracts,
+    supportsChainQueryContracts,
+    supportsAccountAliases,
+    supportsDelegationPV,
+
+    -- * Defunctionalisation symbols
```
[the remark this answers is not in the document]
EmilLa1: I have put documentation before the splice at the `Parameters.hs` and pointed to that. Please check that it is adequate.

haskell-src/Concordium/Types/Parameters.hs
EmilLa1: I have put documentation before the splice. Please check that it is adequate.

haskell-src/Concordium/Types/Parameters.hs
EmilLa1: Elaborated.

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -20,38 +32,253 @@
...
+    IsChainParametersVersion,
+    chainParametersVersion,
+    demoteChainParameterVersion,
+    -- |The chain parameters version for a given protocol version.
+    chainParametersVersionFor,
+    -- |The chain parameters version for a given protocol version (at the type level).
+    ChainParametersVersionFor,
+    -- |The chain parameters version for a given protocol version (on singletons).
+    sChainParametersVersionFor,
+    -- |Defunctionalisation symbol for 'ChainParametersVersionFor'.
```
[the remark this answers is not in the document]
EmilLa1: I have tried documenting this in Parameters.hs above the splice.

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -10,6 +19,9 @@
 -- for pattern matching. (See: https://gitlab.haskell.org/ghc/ghc/-/issues/20896)
 {-# OPTIONS_GHC -Wno-redundant-constraints #-}
 
+-- The class 'IsProtocolVersion' identifies the singleton for each 'ProtocolVersion',
+-- and should have an instance for each constructor of 'ProtocolVersion'.
```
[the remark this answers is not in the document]
EmilLa1: Added some more doc.

abizjak (approved):
I will approve this, but I still state my concerns.

The number of constraints and different fine-grained "versions" is very complex. Approaching this codebase has become even harder now.

The fact that MonoLocalBinds must magically be enabled in modules is another magical thing that requires expert level understanding of advanced type theory. Or just accepting you don't understand the code. Another hurdle.

The fact that now we have these magical "defunctionalization symbols" that are mentioned, but never used by our code, is another hurdle.

The fact that we use an obscure singletons library which, by necessity, is extremely generic, and so the documentation is essentially useless unless you spend a lot of time reading background research. Which will be hard because it is written for other researchers and not for programmers.

I wish I had time to spend to propose another approach and explain the tradeoffs in detail without just complaining. But I don't, so I have to begrudgingly accept this.

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -140,162 +326,74 @@ promoteProtocolVersion P4 = SomeProtocolVersion SP4
...
+-- |Constraint on a type level 'ChainParametersVersion' that can be used to get a corresponding
+-- 'SChainParametersVersion' (see 'chainParametersVersion'). (An alias for 'SingI'.)
+type IsChainParametersVersion (cpv :: ChainParametersVersion) = SingI cpv
+
+-- |Constraint on a type level 'TransactionOutcomesVersion' that can be used to get a corresponding
+-- 'STransactionOutcomesVersion' (see 'transactionOutcomesVersion'). (An alias for 'SingI'.)
+type IsTransactionOutcomesVersion (tov :: TransactionOutcomesVersion) = SingI tov
+
+-- |Constraint on a type level 'TransactionOutcomesVersion' that can be used to get a corresponding
+-- 'STransactionOutcomesVersion' (see 'transactionOutcomesVersion'). This wraps 'SingI', but also
```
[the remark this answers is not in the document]
EmilLa1: Good catch, I think in the first part of the doc it should be `ProtocolVersion` instead of `TransactionOutcomesVersion` in order to make sense. At least then the sentence parses .. :) 

Fixed.

PR #295, opened 2023-01-13
Add implementation for std::hash::Hash for HashBytes

soerenbf:
## Purpose

Added Hash trait to hashbytes type to be able to use it in data structures requiring the Hash trait (such as HashMap)

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/src/hashes.rs
+++ b/rust-src/concordium_base/src/hashes.rs
@@ -4,14 +4,14 @@ use crate::constants::*;
 use crypto_common::{Deserial, SerdeDeserialize, SerdeSerialize, Serial};
 use std::{
     convert::{TryFrom, TryInto},
-    fmt,
+    fmt, hash,
     marker::PhantomData,
     ops::Deref,
     str::FromStr,
 };
 use thiserror::Error;
 
-#[derive(Eq, PartialEq, Ord, PartialOrd, Copy, SerdeSerialize)]
+#[derive(Eq, Ord, PartialOrd, Copy, SerdeSerialize)]
 #[serde(into = "String")]
 /// A general wrapper around Sha256 hash. This is used to add type safety to
 /// a hash that is used in different context. The underlying value is always the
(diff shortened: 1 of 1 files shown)


abizjak (approved):

EmilLa1 (approved):

rust-src/concordium_base/src/hashes.rs
```diff
@@ -23,6 +23,16 @@ pub struct HashBytes<Purpose> {
     _phantom:         PhantomData<Purpose>,
 }
 
+impl<Purpose> PartialEq for HashBytes<Purpose> {
```
abizjak, line 26: I think you added this after the initial review. I am not sure what the purpose is. All Marker types we have implement PartialEq and Eq, so you will get this for those.

In any case, if this is a manual implementation then also `Eq` should be. Since the `Eq` instance now is `<Purpose: Eq> Eq<HashBytes<Purpose>>`.

soerenbf: I added an explicit implementation for `Eq` as well now.

soerenbf: @abizjak , assuming all the checks complete successfully, could you confirm that things are OK now?

rust-src/concordium_base/src/hashes.rs
```diff
@@ -23,6 +23,18 @@ pub struct HashBytes<Purpose> {
...
+    fn eq(&self, other: &Self) -> bool {
+        self.bytes == other.bytes && self._phantom == other._phantom
```
abizjak, line 28:
```suggestion
        self.bytes == other.bytes
```
apart from this it is fine.

abizjak (approved):

rust-src/concordium_base/src/hashes.rs
abizjak: See new comment. THe PartialEq instance should be fixed. Otherwise good.

soerenbf: On another note, an issue has been filed regarding the `derive_hash_xor_eq` clippy check: https://github.com/rust-lang/rust-clippy/issues/2627

Apparently it doesn't make much sense to deny deriving `PartialEq` with an explicit definition of `Hash`, but it's only just been closed recently.

abizjak: Ok.

PR #296, opened 2023-01-13
Move grpc2 types to base.

abizjak:
## Purpose

Move GRPC2 type definitions to concordium-base so that we don't have to duplicate generation and conversion both in the node and in the client.

There are only necessary changes apart from moving the module. Any further desired changes should be done 
afterwards.

## Checklist

      hard-to-understand areas.


limemloh (approved):

limemloh (approved):
The readme should probably mention the new dependency on protoc

Bargsteen (approved):

PR #307, opened 2023-01-23
Add FFI functions for creating `ArInfo`, `IpInfo` and `GlobalContext` instances.

vikt0r0:
## Purpose

This PR adds C ABI functions to the Rust codebase for constructing heap-allocated `ArInfo`, `IpInfo` and `GlobalContext` instances from their field values. It furthermore adds Haskell FFI bindings and wrappers for said functions. This is useful since the field values for instances of these are marshalled in the Node API, and this allows for creating heap-allocated copies of the instances from which the fields were extracted from - for instance in the client

## Changes
- Rust codebase:
  - `ar_info_create`, `ip_info_create` and `global_context_create`, 
- Haskell codebase:
  - Corresponding bindings an wrappers for the above.

## Checklist

      hard-to-understand areas.


vikt0r0:
I am currently writing some tests, and I guess I need to specify an output size as well. Looking at other FFI functions it seems sufficient to add a `size_t`pointer as the last parameter and then set the value of the location accordingly, e.g.

```
extern "C" fn my_function(
    ...,
    output_len: *mut size_t
) -> *const u8 {
    ...
    unsafe { *output_len = output_instance_bytes.len() as size_t  }
    output_instance_ptr
}
```

Does anyone know if this is the case?

abizjak:
> 

What do you 

> I am currently writing some tests, and I guess I need to specify an output size as well. Looking at other FFI functions it seems sufficient to add a `size_t`pointer as the last parameter and then set the value of the location accordingly, e.g.
> 
> ```
> extern "C" fn my_function(
>     ...,
>     output_len: *mut size_t
> ) -> *const u8 {
>     ...
>     unsafe { *output_len = output_instance_bytes.len() as size_t  }
>     output_instance_ptr
> }
> ```
> 
> I do not see explicit mention of the output length parameters in the Haskell codebase. Will doing so just work?

We do use this pattern a lot. What do you mean that you don't see it in the Haskell codebase. For example

```
foreign import ccall unsafe "bls_sk_to_bytes" toBytesSecretKey :: Ptr SecretKey -> Ptr CSize -> IO (Ptr Word8)
```
 is like this, and most other serialization functions.

haskell-src/Concordium/ID/AnonymityRevoker.hs
```diff
@@ -38,6 +39,31 @@ foreign import ccall unsafe "ar_info_name" arNameFFI :: Ptr ArInfo -> Ptr CSize
 foreign import ccall unsafe "ar_info_url" arUrlFFI :: Ptr ArInfo -> Ptr CSize -> IO (Ptr Word8)
 foreign import ccall unsafe "ar_info_description" arDescriptionFFI :: Ptr ArInfo -> Ptr CSize -> IO (Ptr Word8)
 foreign import ccall unsafe "ar_info_public_key" arPublicKeyFFI :: Ptr ArInfo -> Ptr CSize -> IO (Ptr Word8)
+foreign import ccall unsafe "ar_info_create"
+    createArInfoFFI ::
+        Word32 ->
+        Ptr Word8 -> CSize ->
+        Ptr Word8 -> CSize ->
+        Ptr Word8 -> CSize ->
+        Ptr Word8 -> CSize ->
```
abizjak, line 48: Please document all these arguments.

haskell-src/Concordium/ID/AnonymityRevoker.hs
```diff
@@ -38,6 +39,31 @@ foreign import ccall unsafe "ar_info_name" arNameFFI :: Ptr ArInfo -> Ptr CSize
...
+createArInfo :: ArIdentity -> BS.ByteString -> Text -> Text -> Text -> Maybe ArInfo
+createArInfo arId pubKey name url desc = unsafePerformIO ( do
+    let (ArIdentity idW) = arId
+    (pkPtr, pkLen) <- toByteArrayInput pubKey
+    (nPtr, nLen) <- toByteArrayInput (Text.encodeUtf8 name)
+    (urlPtr, urlLen) <- toByteArrayInput (Text.encodeUtf8 url)
+    (descPtr, descLen) <- toByteArrayInput (Text.encodeUtf8 desc)
+    ptr <- createArInfoFFI idW pkPtr pkLen nPtr nLen urlPtr urlLen descPtr descLen
+    if ptr == nullPtr
+    then return Nothing
```
abizjak, line 60: Is this ever necessary? When does this return a null pointer?

haskell-src/Concordium/ID/AnonymityRevoker.hs
```diff
@@ -38,6 +39,31 @@ foreign import ccall unsafe "ar_info_name" arNameFFI :: Ptr ArInfo -> Ptr CSize
...
+    else Just . ArInfo <$> newForeignPtr freeArInfo ptr)
+    where
+        toByteArrayInput bs =
+            unsafeUseAsCStringLen bs
+                $ \(p, l) -> return (castPtr p :: Ptr Word8, fromIntegral l :: CSize)
+
```
abizjak, line 66: This won't work correctly. It might seem to work most of the time though.

The pointer you have inside `unsafeUseAsCStringLen` cannot escape the closure .So you have to nest all of these. See  how we use this elsewhere for an example.

haskell-src/Concordium/ID/IdentityProvider.hs
```diff
@@ -38,6 +39,31 @@ foreign import ccall unsafe "ip_info_url" ipUrlFFI :: Ptr IpInfo -> Ptr CSize ->
...
+    (urlPtr, urlLen) <- toByteArrayInput (Text.encodeUtf8 url)
+    (descPtr, descLen) <- toByteArrayInput (Text.encodeUtf8 desc)
+    ptr <- createIpInfoFFI idW vkPtr vkLen cvkPtr cvkLen nPtr nLen urlPtr urlLen descPtr descLen
+    if ptr == nullPtr
+    then return Nothing
+    else Just . IpInfo <$> newForeignPtr freeIpInfo ptr)
+    where
+        toByteArrayInput bs =
+            unsafeUseAsCStringLen bs
+                $ \(p, l) -> return (castPtr p :: Ptr Word8, fromIntegral l :: CSize)
```
abizjak, line 66: Same comments as for anonymity revokers.

haskell-src/Concordium/ID/Parameters.hs
```diff
@@ -27,6 +30,26 @@ foreign import ccall safe "global_context_to_bytes" globalContextToBytes :: Ptr
...
+    (bpgPtr, bpgLen) <- toByteArrayInput bulletProofGens
+    (occPtr, occLen) <- toByteArrayInput onChainComm
+    ptr <- createGlobalContextFFI gsPtr gsLen bpgPtr bpgLen occPtr occLen
+    if ptr == nullPtr
+    then return Nothing
+    else Just . GlobalContext <$> newForeignPtr freeGlobalContext ptr)
+    where
+        toByteArrayInput bs =
+            unsafeUseAsCStringLen bs
+                $ \(p, l) -> return (castPtr p :: Ptr Word8, fromIntegral l :: CSize)
```
abizjak, line 52: Same comments as above.

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +298,203 @@ pub extern "C" fn ip_info_cdi_verify_key(
...
+    desc_ptr: *const u8,
+    desc_len: size_t,
+) -> *mut ArInfo<G1> {
+    // Description.
+    let name;
+    if let Ok(_name) = from_utf8(slice_from_c_bytes!(name_ptr, name_len as usize)) {
+        name = _name.to_string();
+    } else {
+        return std::ptr::null_mut();
+    }
```
abizjak, line 320: You should just state a precondition to this function and not handle these problems here. This is an internal function and we know it will always be called with valid utf8 strings.

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +298,203 @@ pub extern "C" fn ip_info_cdi_verify_key(
...
+extern "C" fn global_context_create(
+    genesis_string_ptr: *const u8,
+    genesis_string_len: size_t,
+    bulletproof_generators_ptr: *const u8,
+    bulletproof_generators_len: size_t,
+    on_chain_commitment_ptr: *const u8,
+    on_chain_commitment_len: size_t,
+) -> *mut GlobalContext<G1> {
+    let genesis_string;
+    if let Ok(_genesis_string) = from_utf8(slice_from_c_bytes!(
```
abizjak, line 379: Same as above. Assume valid input.

abizjak (changes requested):

haskell-src/Concordium/ID/AnonymityRevoker.hs
```diff
@@ -38,6 +39,31 @@ foreign import ccall unsafe "ar_info_name" arNameFFI :: Ptr ArInfo -> Ptr CSize
...
+foreign import ccall unsafe "ar_info_create"
+    createArInfoFFI ::
+        Word32 ->
+        Ptr Word8 -> CSize ->
+        Ptr Word8 -> CSize ->
+        Ptr Word8 -> CSize ->
+        Ptr Word8 -> CSize ->
+        IO (Ptr ArInfo)
+
+createArInfo :: ArIdentity -> BS.ByteString -> Text -> Text -> Text -> Maybe ArInfo
```
Bargsteen, line 51: Please document 😊

haskell-src/Concordium/ID/IdentityProvider.hs
```diff
@@ -38,6 +39,31 @@ foreign import ccall unsafe "ip_info_url" ipUrlFFI :: Ptr IpInfo -> Ptr CSize ->
...
 foreign import ccall unsafe "ip_info_verify_key" ipVerifyKeyFFI :: Ptr IpInfo -> Ptr CSize -> IO (Ptr Word8)
 foreign import ccall unsafe "ip_info_cdi_verify_key" ipCdiVerifyKeyFFI :: Ptr IpInfo -> Ptr CSize -> IO (Ptr Word8)
+foreign import ccall unsafe "ip_info_create" createIpInfoFFI ::
+    Word32 ->
+    Ptr Word8 -> CSize ->
+    Ptr Word8 -> CSize ->
+    Ptr Word8 -> CSize ->
+    Ptr Word8 -> CSize ->
+    Ptr Word8 -> CSize ->
+    IO (Ptr IpInfo)
```
Bargsteen, line 49: Docs

haskell-src/Concordium/ID/IdentityProvider.hs
```diff
@@ -38,6 +39,31 @@ foreign import ccall unsafe "ip_info_url" ipUrlFFI :: Ptr IpInfo -> Ptr CSize ->
...
+
+createIpInfo :: IdentityProviderIdentity -> BS.ByteString -> BS.ByteString -> Text -> Text -> Text -> Maybe IpInfo
```
Bargsteen, line 51: Docs

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +298,208 @@ pub extern "C" fn ip_info_cdi_verify_key(
...
+    };
+
+    let ar_info = ArInfo {
+        ar_identity,
+        ar_description,
+        ar_public_key,
+    };
+
+    let bytes = to_bytes(&ar_info);
+    *output_len = to_bytes(&ar_info).len() as size_t;
```
Bargsteen, line 360:
```suggestion
    *output_len = bytes.len() as size_t;
```

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +298,208 @@ pub extern "C" fn ip_info_cdi_verify_key(
...
+        }
+    };
+
+    let ar_info = ArInfo {
+        ar_identity,
+        ar_description,
+        ar_public_key,
+    };
+
+    let bytes = to_bytes(&ar_info);
```
Bargsteen, line 359: Please document what is happening here, and why you use `mem::forget`

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +298,208 @@ pub extern "C" fn ip_info_cdi_verify_key(
...
+    *output_len = to_bytes(&ar_info).len() as size_t;
+    std::mem::forget(bytes);
+    Box::into_raw(Box::new(ar_info))
+}
+
+/// Create a GlobalContext instance from byte array pointers.
```
Bargsteen, line 365:
```suggestion
/// Create a [`GlobalContext`] instance from byte array pointers.
```

Bargsteen: This will make the docs include a link to the GlobalContext type. Please add links when relevant 😊

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +298,208 @@ pub extern "C" fn ip_info_cdi_verify_key(
...
+    };
+
+    let global_context = GlobalContext {
+        on_chain_commitment_key,
+        bulletproof_generators,
+        genesis_string,
+    };
+
+    let bytes = to_bytes(&global_context);
+    *output_len = to_bytes(&global_context).len() as size_t;
```
Bargsteen, line 419:
```suggestion
    *output_len = bytes.len() as size_t;
```
And please also document the bytes & forgetting here.

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +298,208 @@ pub extern "C" fn ip_info_cdi_verify_key(
...
+
+    let ip_info = IpInfo {
+        ip_identity,
+        ip_description,
+        ip_verify_key,
+        ip_cdi_verify_key,
+    };
+
+    let bytes = to_bytes(&ip_info);
+    *output_len = to_bytes(&ip_info).len() as size_t;
```
Bargsteen, line 498:
```suggestion
    *output_len = bytes.len() as size_t;
```

Bargsteen (commented):
Had a few comments, but looks like it is nearly there 😊

rust-src/id/src/ffi.rs
```diff
@@ -387,6 +583,73 @@ mod test {
...
+                    ip_info_url_bs.len(),
+                    ip_info_desc_bs.as_ptr(),
+                    ip_info_desc_bs.len(),
+                )
+            }
+        };
+
+        // FIXME: How to check equality; can I simple derive PartialEq and Eq,
+        //        and it will compare the field values (recursively) using
+        //        their comparison functions?
```
abizjak, line 647: Yes.

rust-src/id/src/ffi.rs
```diff
@@ -387,6 +583,73 @@ mod test {
...
+
+            // Get the identity.
+            let ip_info_identity = ip_info_ip_identity(ip_info_ptr);
+
+            // Get the description fields.
+            let mut ip_info_desc_bs_len: size_t = 0;
+            let ip_info_desc: Description = {
+                let ip_info_desc_ptr = ip_info_description(ip_info_ptr, &mut ip_info_desc_bs_len);
+                let buf = &mut slice_from_c_bytes!(ip_info_desc_ptr, ip_info_desc_bs_len as usize);
+                from_bytes(buf).expect("Bad stuff happened.")
```
abizjak, line 599:
```suggestion
                let buf = slice_from_c_bytes!(ip_info_desc_ptr, ip_info_desc_bs_len as usize);
                let desc = String::from_utf8(buf).expect("Not valid UTF8);
```

rust-src/id/src/ffi.rs
vikt0r0: Thanks. I still find it odd that I can simply compare the ps_sig::PublicKey<P> instances when P is Bls12.

But in that case I will scrap it; I wanted to ensure the FFI functions work as intended since I am not very experienced in Rust.

abizjak: We manually implement it
```
impl<C: Pairing> PartialEq for PublicKey<C> {
    fn eq(&self, other: &Self) -> bool {
        self.g == other.g
            && self.g_tilda == other.g_tilda
            && self.ys == other.ys
            && self.y_tildas == other.y_tildas
            && self.x_tilda == other.x_tilda
    }
}

```
giving the constraint `C: Pairing`. If that is what you mean.

haskell-src/Concordium/ID/IdentityProvider.hs
```diff
@@ -38,6 +39,42 @@ foreign import ccall unsafe "ip_info_url" ipUrlFFI :: Ptr IpInfo -> Ptr CSize ->
...
+    Ptr Word8 -> CSize -> -- Pointer to a byte array returned by ipCdiVerifyKeyFFI and its length. The
+                          -- array must be the binary representation of a `ps_sig::PublicKey<Bls12>`
+                          -- Rust instance.
+    Ptr Word8 -> CSize -> -- Pointer to a byte array returned by ipNameFFI and its length. The array
+                          -- must correspond to a utf8 encoded string.
+    Ptr Word8 -> CSize -> -- Pointer to a byte array returned by ipUrlFFI and its length. The array
+                          -- must correspond to a utf8 encoded string.
+    Ptr Word8 -> CSize -> -- Pointer to a byte array returned by ipDescriptionFFI and its length. The
+                          -- array must correspond to a utf8 encoded string.
+    IO (Ptr IpInfo) -- Pointer to an `IpInfo Rust instance with its corresponding fields set to the above
```
Bargsteen, line 56:
```suggestion
    IO (Ptr IpInfo) -- Pointer to an `IpInfo` Rust instance with its corresponding fields set to the above
```

Bargsteen (approved):
LGTM 👍

haskell-src/Concordium/ID/Parameters.hs
```diff
@@ -27,6 +30,30 @@ foreign import ccall safe "global_context_to_bytes" globalContextToBytes :: Ptr
...
+
+-- Create an GlobalContext intance from bytestrings and texts.
+-- This function is a wrapper for `createGlobalContextFFI`, and is used for creating heap-allocated
+-- `GlobalContext` instances from bytestrings. See the `createGlobalContextFFI` import declaration
+-- for information about the function inputs and preconditions.
+createGlobalContext :: Text -> BS8.ByteString -> BS8.ByteString -> GlobalContext
+createGlobalContext genString bulletProofGens onChainComm = unsafePerformIO ( do
+    ptr <- unsafeUseAsCStringLen (Text.encodeUtf8 genString) $ \(gsPtr, gsLen) ->
+        unsafeUseAsCStringLen bulletProofGens $ \(bpgPtr, bpgLen) ->
+            unsafeUseAsCStringLen onChainComm $ \(occPtr, occLen) ->
```
abizjak, line 51: It is important to be aware that if the string is empty what the Rust side gets is an arbitrary pointer. So it is important to document that this situation si handled appropriately on the Rust side. `slice_from_c_bytes` handles it correctly. But please state this as we do in other cases.

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +298,196 @@ pub extern "C" fn ip_info_cdi_verify_key(
...
+
+    // Public key.
+    let ar_public_key = {
+        let buf = &mut slice_from_c_bytes!(public_key_ptr, public_key_len as usize);
+
+        from_bytes::<elgamal::PublicKey<G1>, &[u8]>(buf).expect(
+            "Precondition failed: Unable to create PublicKey<G1> instance from byte array at \
+             public_key_ptr.",
+        )
+    };
```
abizjak, line 329: Are you sure this precondition is always satisfied?

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +298,196 @@ pub extern "C" fn ip_info_cdi_verify_key(
...
+    let ip_identity = IpIdentity::try_from(identity)
+        .expect("Precondition failed: Unable to create IpIdentity instance from u32.");
+
+    // Identity provider verify key.
+    let ip_verify_key = {
+        let ps_buf = &mut slice_from_c_bytes!(verify_key_ptr, verify_key_len as usize);
+
+        from_bytes::<ps_sig::PublicKey<Bls12>, &[u8]>(ps_buf).expect(
+            "Precondition failed: Unable to create PublicKey<Bls12> instance from byte array at \
+             verify_key_ptr.",
```
abizjak, line 446: Are you sure this precondition is always satisfied when you call this?

haskell-src/Concordium/ID/Parameters.hs
```diff
@@ -27,6 +30,30 @@ foreign import ccall safe "global_context_to_bytes" globalContextToBytes :: Ptr
...
 foreign import ccall safe "global_context_to_json" globalContextToJSONFFI :: Ptr GlobalContext -> Ptr CSize -> IO (Ptr Word8)
 foreign import ccall safe "global_context_from_json" globalContextFromJSONFFI :: Ptr Word8 -> CSize -> IO (Ptr GlobalContext)
+foreign import ccall unsafe "global_context_create"
+    createGlobalContextFFI ::
+        Ptr Word8 -> CSize -> -- Pointer to a byte array representing the genesis string and its length.
+                              -- The array must correspond to a utf8 encoded string.
+        Ptr Word8 -> CSize -> -- Pointer to a byte array and its length. The array must be the binary
+                              -- representation of a `Generators<G1>` Rust instance.
+        Ptr Word8 -> CSize -> -- Pointer to a byte array and its length. The array must be the binary
+                              -- representation of a `PedersenKey<G1>` Rust instance.
```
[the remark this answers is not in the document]
vikt0r0: I'm not sure I understand. I can not guarantee anything about the input data, but according to your comments for the `*_create `-functions, I believe you mentioned that it should be stated as a precondition that the input is a valid serialization. Before that, I returned null-pointers and checked whether the returned pointer was null in the FFI wrapper.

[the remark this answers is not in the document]
abizjak: You do know that the strings are valid utf8 encoding. Protobuf guarantees that.

But the public keys are not necessarily well-formed. So failure there needs to be handled.

Sorry for misleading comments.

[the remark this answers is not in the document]
vikt0r0: I see, thanks. I see examples of "verification" FFI-functions, for instance `verifyCDIFFI`? I suppose I use these and implement where needed?

haskell-src/Concordium/ID/IdentityProvider.hs
```diff
@@ -38,6 +39,57 @@ foreign import ccall unsafe "ip_info_url" ipUrlFFI :: Ptr IpInfo -> Ptr CSize ->
...
+    -- `ps_sig::PublicKey<Bls12>` Rust-instance, for instance one returned
+    -- by ipCdiVerifyKeyFFI, and its length.
+    Ptr Word8 -> CSize ->
+    -- Pointer to a byte array which is the binary representation of an
+    -- utf8 encoded string, for instance one returned by ipNameFFI, and
+    -- its length.
+    Ptr Word8 -> CSize ->
+    -- Pointer to a byte array which is the binary representation of an
+    -- utf8 encoded string, for instance one returned by ipUrlFFI, and
+    -- its length.
```
abizjak, line 59: As I wrote before, it is not relevant where the pointer is coming from. And in any case in the place this is used the byte array is not coming from "ipUrlFFI". So please remove these "for instance" since they do not bring value.

What is relevant is that the pointer is of the right size, and the data is valid utf8.

haskell-src/Concordium/ID/IdentityProvider.hs
```diff
@@ -38,6 +39,57 @@ foreign import ccall unsafe "ip_info_url" ipUrlFFI :: Ptr IpInfo -> Ptr CSize ->
...
+    Ptr Word8 -> CSize ->
+    -- Pointer to a byte array which is the binary representation of an
+    -- utf8 encoded string, for instance one returned by ipDescriptionFFI,
+    -- and its length.
+    Ptr Word8 -> CSize ->
+    -- Pointer to an `IpInfo Rust instance with its corresponding fields set
+    -- to the above values. This is a null-pointer on failure.
+    IO (Ptr IpInfo)
+
+-- Create an @IpInfo@ instance from bytestrings and texts.
```
abizjak, line 69:
```suggestion
-- Create an @IpInfo@ instance from constituent parts.
```

haskell-src/Concordium/ID/IdentityProvider.hs
```diff
@@ -38,6 +39,57 @@ foreign import ccall unsafe "ip_info_url" ipUrlFFI :: Ptr IpInfo -> Ptr CSize ->
...
+-- This function is a wrapper for `createIpInfoFFI`, and is used for creating heap-allocated @IpInfo@
+-- instances from raw and utf8-encoded bytestrings. The inputs should conform to field values extracted
+-- from @IpInfo@ instances using FFI functions, see the `createIpInfoFFI` import declaration
+-- for information about the function inputs and preconditions. Returns @Nothing@ on failure.
+createIpInfo:: IdentityProviderIdentity -> BS.ByteString -> BS.ByteString -> Text -> Text -> Text -> Maybe IpInfo
```
abizjak, line 74:
```suggestion
createIpInfo:: IdentityProviderIdentity ->
                         -- |Serialized Pointcheval-Sanders public key.
                         BS.ByteString -> 
                         -- |Serialized Ed25519 public key
                         BS.ByteString -> 
                         -- |Name of the identity provider
                         Text -> 
                         -- |URL of the identity provider.
                         Text -> 
                         -- |Description of the provider.
                         Text -> 
                         -- |If the public keys cannot be deserialized this returns 'Nothing'. Otherwise an identity provider is returned.
                         Maybe IpInfo
```

haskell-src/Concordium/ID/Parameters.hs
```diff
@@ -27,6 +30,41 @@ foreign import ccall safe "global_context_to_bytes" globalContextToBytes :: Ptr
...
+        -- A pointer to an @GlobalContext@ instance with its corresponding
+        -- fields set to the above values. This is a null-pointer on failure.
+        IO (Ptr GlobalContext)
+
+-- Create a @GlobalContext@ instance from bytestrings and texts.
+-- This function is a wrapper for `createGlobalContextFFI`, and is used for creating heap-allocated
+-- @GlobalContext@ instances from bytestrings. See the `createGlobalContextFFI` import declaration
+-- for information about the function inputs and preconditions. Returns @Nothing@ on failure.
+createGlobalContext :: Text -> BS8.ByteString -> BS8.ByteString -> Maybe GlobalContext
+createGlobalContext genString bulletProofGens onChainComm = unsafePerformIO ( do
```
abizjak, line 54: Same as for ip-info

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +299,274 @@ pub extern "C" fn ip_info_cdi_verify_key(
     ptr
 }
 
+/// Attempt to create an [`ArInfo`] instance from byte array pointers.
+/// Assumes that public_key_ptr points to a PublicKey<G1>
```
abizjak, line 303: I don't see that this function assumes that public key points to that. Please update documentation.

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +299,274 @@ pub extern "C" fn ip_info_cdi_verify_key(
...
+        name_ptr,
+        name_len,
+        url_ptr,
+        url_len,
+        desc_ptr,
+        desc_len,
+    ) {
+        return ptr;
+    }
+    std::ptr::null_mut()
```
abizjak, line 332:
```suggestion
    ) {
        ptr
    } else {
        std::ptr::null_mut()
    }
```
Rust is expression based. Please use that.

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +299,274 @@ pub extern "C" fn ip_info_cdi_verify_key(
...
+        genesis_string_ptr,
+        genesis_string_len,
+        bulletproof_generators_ptr,
+        bulletproof_generators_len,
+        on_chain_commitment_ptr,
+        on_chain_commitment_len,
+    ) {
+        return ptr;
+    }
+    std::ptr::null_mut()
```
abizjak, line 416: Same here.

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +299,274 @@ pub extern "C" fn ip_info_cdi_verify_key(
...
+}
+
+/// Attempt to create a [`GlobalContext`] instance from byte array pointers.
+/// Assumes that genesis_string_ptr points to a valid
+/// utf8 string, bulletproof_generators_ptr to a
+/// Generators<G1> instance and on_chain_commitments to a
+/// PedersenKey<G1> instance, and returns a null-pointer
+/// otherwise.
```
abizjak, line 424: This documentation does not reflect reality. Please update and explain that null is returned if data cannot be deserialized.

rust-src/id/src/ffi.rs
```diff
@@ -292,6 +299,274 @@ pub extern "C" fn ip_info_cdi_verify_key(
...
+        on_chain_commitment_key,
+        bulletproof_generators,
+        genesis_string,
+    })))
+}
+
+/// Attempt to create an [`IpInfo`] instance from byte array pointers.
+/// Assumes that verify_key_ptr points to a
+/// ps_sig::PublicKey<Bls12> instance,
+/// cdi_verify_key_ptr to a ed25519_dalek::PublicKey
```
abizjak, line 471: Same documentation issue here.

abizjak (commented):
It's looking OK, but documentation needs to be fixed so as to not be misleading.

abizjak (approved):
Please fix "assumes" documentation. And then merge.

PR #308, opened 2023-01-26
Fix encrypted amount ToProto instance to include incoming amounts.

abizjak:
## Purpose

Fix a bug in `ToProto` instance that omitted encrypted transfers.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/GRPC2.hs
+++ b/haskell-src/Concordium/GRPC2.hs
@@ -14,6 +14,7 @@ where
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Short as BSS
 import Data.Coerce
+import Data.Foldable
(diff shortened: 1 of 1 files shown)


vikt0r0 (approved):
Looks good if this pleases GHC.

haskell-src/Concordium/GRPC2.hs
```diff
@@ -14,6 +14,7 @@ where
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Short as BSS
 import Data.Coerce
+import Data.Foldable
```
Bargsteen, line 17: Consider:
```suggestion
import Data.Foldable (toList)
```
Or qualified.

Bargsteen (approved):
One comment, but otherwise looks good 😊

PR #310, opened 2023-01-30
Add token address string format.

abizjak:
## Purpose

Add a text format for "token address", the full identifier of a token.

This is based on base58check encoding with version 2 to achieve the following
- comparing the first few and the last few characters should be sufficient in most cases to determine whether tokens are distinct.
- the format is short and copyable
- the address can never be confused with an account address since version is 2.

## Checklist

      hard-to-understand areas.


rust-src/concordium_base/src/cis2_types.rs
```diff
@@ -82,6 +82,53 @@ impl TokenAmount {
...
+    id:       TokenId,
+}
+
+impl TryFrom<&str> for TokenAddress {
+    type Error = anyhow::Error;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        let bytes = bs58::decode(value).with_check(Some(2)).into_vec()?;
+        if let Some((&2, mut source)) = bytes.split_first() {
+            let index = leb128::read::unsigned(&mut source)?;
```
limemloh, line 100: Is the LEB128 really needed here? We will probably end up implementing this in multiple languages, so unless it provides good value, I think we should keep it simple.

limemloh (approved):

rust-src/concordium_base/src/cis2_types.rs
abizjak: It makes for much shorter token IDs, which I think is a good tradeoff.

limemloh: fair enough

PR #315, opened 2023-02-09
Testnet variant.

ZB-concordium:
Hi Team
We got a message from a user that it would be nice 2 separate user manuals for ID creation.

"I think generate request instructions is wrong. It's only for mainnet. If testnet, I have to replace some words: cryptographic-parameters.json -> cryptographic-parameters-testnet.json ars.json -> ars-testnet.json
ip-info.json -> ip-info-testnet.json
Documents should be modified."

Thank you

## Purpose

Not clear to everyone the difference...

## Changes

Testnet added in the file names

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-bins/docs/user-cli.md
+++ b/rust-bins/docs/user-cli.md
@@ -28,6 +28,17 @@ user_cli generate-request --cryptographic-parameters cryptographic-parameters.js
                           --id-use-data-out id-use-data.json \ # data that enables use of the identity object
                           --request-out request.json # request to send to the identity provider
 ```
+
+For Testnet woule be the following: 
+```console
+user_cli generate-request --cryptographic-parameters cryptographic-parameters-testnet.json \
+                          --ars ars-testnet.json \
+                          --ip-info ip-info-testnet.json \
+                          --initial-keys-out initial-keys.json \ # keys of the initial account together with its address.
+                          --id-use-data-out id-use-data.json \ # data that enables use of the identity object
+                          --request-out request.json # request to send to the identity provider
+```
+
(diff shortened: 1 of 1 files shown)


rust-bins/docs/user-cli.md
```diff
@@ -28,6 +28,17 @@ user_cli generate-request --cryptographic-parameters cryptographic-parameters.js
                           --id-use-data-out id-use-data.json \ # data that enables use of the identity object
                           --request-out request.json # request to send to the identity provider
 ```
+
+For Testnet woule be the following:
```
abizjak, line 32:
```suggestion
For Testnet would be the following: 
```

abizjak: This is also n the "Version 0 flow" section. If we are to add this we need to add one also in the V1 section.

abizjak (changes requested):

abizjak:
The thing is, I don't think this fits into the documentation. For example in the beginning of the documentation we have
```

- A file with cryptographic parameters for the chain. We'll refer to this file as `cryptographic-parameters.json` below.
- A file with the list of anonymity revokers supported by the identity provider. We refer to this file as `ars.json` below.
- A file with identity provider public keys. We refer to this file as `ip-info.json` below.
- For creating accounts from the identity the `concordium-client` tool and access to an up-to-date node.
```

so we have to work with this. Just adding these examples you added now does not fit into the rest of the documentation.

If you want to add these specific examples with `-testnet` in the name we need to revise more of the documentation.

abizjak:
Since we have added this to the developer documentation now, this document should not be updated with network specific configurations. So I am closing this PR.

PR #319, opened 2023-02-14
Mobile wallet lib: Add support for receiving just the parameter's schema in parameter_to_json

shjortConcordium:
## Purpose

Support schemas for just the function's parameters in the mobile wallet lib's `parameter_to_json` function.

## Changes

 - Bumped mobile_wallet version
 - Changed parameter_to_json to handle differerent formats for the schema field.

## Checklist

      hard-to-understand areas.


mobile_wallet/src/lib.rs
```diff
@@ -224,7 +236,11 @@ fn parameter_to_json_aux(input: &str) -> anyhow::Result<String> {
     let serialized_parameter: String = try_get(&v, "parameter")?;
     let receive_name: OwnedReceiveName = try_get(&v, "receiveName")?;
     let parameter: Parameter = Parameter::new_unchecked(hex::decode(serialized_parameter)?);
-    let schema: String = try_get(&v, "schema")?;
+    let schema: SchemaInputType = match try_get(&v, "schema") {
+        Ok(v) => v,
+        // To support the legacy format we also attempt to parse the schema as a string directly:
+        Err(e) => SchemaInputType::Module(try_get(&v, "schema").map_err(|_| e)?),
+    };
```
abizjak, line 243: It would be cleaner to get a `serde_json::Value` first, then check if that is an object or string and then do the appropriate parsing. I don't like this "try and see" approach.

Do we also actually need to handle this legacy case here? Won't the wallet itself have to deal with this already?

abizjak (approved):
Ok, let's leave the legacy handling as well. Not sure it will be needed, but it does not hurt if itis there.

PR #323, opened 2023-02-20
New treestate

EmilLa1:
## Purpose

Move `MonadProtocolVersion` into base.

This supports the new consensus protocol.

## Changes

- Moved `MonadProtocolVersion` to `Concordium.Types`

## Checklist

      hard-to-understand areas.


haskell-src/Concordium/Genesis/Data/BaseV1.hs
```diff
@@ -58,3 +65,25 @@ instance Serialize RegenesisDataV1 where
         genesisTerminalBlock <- get
         genesisStateHash <- get
         return RegenesisDataV1{..}
+
+-- |Convert 'GenesisParametersV2' to genesis data.
```
EmilLa1, line 69:
```suggestion
-- |Convert 'GenesisParametersV3' to genesis data.
```

haskell-src/Concordium/Genesis/Data/BaseV1.hs
```diff
@@ -58,3 +65,25 @@ instance Serialize RegenesisDataV1 where
...
+    genesisAccounts = Vec.fromList gp3InitialAccounts
+    genesisCryptographicParameters = gp3CryptographicParameters
+    genesisIdentityProviders =
+        case filter (\(k, v) -> k /= ipIdentity v) (Map.toList (idProviders gp3IdentityProviders)) of
+            [] -> gp3IdentityProviders
+            ips -> error $ "Inconsistent identity provider ids: " ++ show ips
+    genesisAnonymityRevokers =
+        case filter (\(k, v) -> k /= arIdentity v) (Map.toList (arRevokers gp3AnonymityRevokers)) of
+            [] -> gp3AnonymityRevokers
+            ars -> error $ "Inconsistent anonymity revoker ids: " ++ show ars
```
abizjak, line 87: I think I had this comment in a previous PR already, but checking these seems arbitrary here.

For example we don't check that the anonymity revokers and identity providers that appear in the accounts are the right ones either.

So I think checking this is arbitrary and unclear why we need to bother.

haskell-src/Concordium/Genesis/Data/P6.hs
```diff
@@ -73,6 +74,9 @@ putVersionedGenesisData gd = do
     putVersion 8
     putGenesisDataV8 gd
 
+parametersToGenesisData :: GenesisParametersV3 'P6 -> GenesisDataP6
+parametersToGenesisData = uncurry GDP6Initial . BaseV1.parametersToState
```
abizjak, line 78: In a previous PR this function was removed with the argument that it is not needed. Why is it being reintroduced?

td202: This was an oversight as this branched from an old version. `GenesisParametersV3` has now been (re)removed.

td202 (approved):

haskell-src/Concordium/Types.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE ConstraintKinds #-}
```
abizjak, line 1: this does not seem to be needed

abizjak (approved):

haskell-src/Concordium/Types.hs
EmilLa1: Nice catch!

limemloh (approved):

PR #324, opened 2023-02-21
New consensus leader election

td202:
## Purpose

Revise the seed state with fields to support leader election in the new consensus version.

## Changes

- Rather than having fields `Conditionally`, `SeedState` is spit into constructors for different `SeedStateVersion`s.
- `SeedStateV1` removes `epochLength` and adds:
  - `triggerBlockTime`: the timestamp at or after which the trigger block for this epoch occurs.
  - `epochTransitionTriggered`: flag indicating if a trigger block has already been encountered for this epoch.
- Add a `hashToInteger` function for SHA256 hashes. This is used for leader election calculations.

## Checklist

      hard-to-understand areas.


haskell-src/Concordium/Types/SeedState.hs
```diff
@@ -53,36 +51,103 @@ withIsSeedStateVersionFor :: SProtocolVersion pv -> (IsSeedStateVersion (SeedSta
...
-      -- |Current leadership election nonce
-      currentLeadershipElectionNonce :: !LeadershipElectionNonce,
-      -- |The leadership election nonce updated with the block nonces
-      -- of blocks in the first 2/3 of the current epoch.
-      updatedNonce :: !Hash
-    }
-    deriving (Eq, Show)
+--
+-- (Implementation note: the constructors have equational constraints so that they have the same
+-- return time. This allows them them to have common field names (e.g. 'epoch'), which is not
```
chrmatt, line 56: Typo "them them".

I don't understand what is meant here by "return time".

td202: Should be "return type".

EmilLa1 (approved):
Good job

eb-concordium (approved):

haskell-src/Concordium/Types/SeedState.hs
```diff
@@ -52,36 +51,103 @@ withIsSeedStateVersionFor :: SProtocolVersion pv -> (IsSeedStateVersion (SeedSta
...
-      -- |The leadership election nonce updated with the block nonces
-      -- of blocks in the first 2/3 of the current epoch.
-      updatedNonce :: !Hash
-    }
-    deriving (Eq, Show)
+--
+-- (Implementation note: the constructors have equational constraints so that they have the same
+-- return type. This allows them to have common field names (e.g. 'epoch'), which is not
+-- allowed if the return types differ (e.g. @SeedState 'SeedStateVersion0@ and
+-- @SeedState 'SeedStateVersion1@).)
```
abizjak, line 58: Does this provide a lot of convenience? 

What happens to `epochTransitionTriggered`? In normal ADTs you would get a partial function from a field that only appears in one of the variants. Which is annoyingly error prone. 

Is that better here in that you get a function
```
epochTransitionTriggered :: (ssv ~ 'SeedStateVersion1) -> SeedStateVersion ssv -> Bool
```
?

haskell-src/Concordium/Types/SeedState.hs
```diff
@@ -52,36 +51,103 @@ withIsSeedStateVersionFor :: SProtocolVersion pv -> (IsSeedStateVersion (SeedSta
...
+          -- |Current leadership election nonce
+          currentLeadershipElectionNonce :: !LeadershipElectionNonce,
+          -- |The leadership election nonce updated with the block nonces
+          -- of blocks in the first 2/3 of the current epoch.
+          updatedNonce :: !Hash
+        } ->
+        SeedState ssv
+    SeedStateV1 ::
+        (ssv ~ 'SeedStateVersion1) =>
+        { -- |Current epoch
```
abizjak, line 76:
```suggestion
        { -- |Current epoch.
```

abizjak (approved):

haskell-src/Concordium/Types/SeedState.hs
EmilLa1: Yes. Instead of throwing an exception at runtime then it will error out at compile time telling you that the the constraints can't be met (if you were to define that function like :

```
epochTransitionTriggered :: (ssv ~ 'SeedStateVersion1) => SeedStateVersion ssv -> Bool

```

EmilLa1: Not except for naming.

td202: I have reworked to use lenses instead that are type safe.

PR #332, opened 2023-03-03
Move smart contract tests from Concordium Node (with history)

Bargsteen:
## Purpose

Move the smart contract test files from concordium node.
The reason for this is that the files are also needed in the smart contract integration testing library.

## Changes

- Move the files (now including the history ;) )
- Add the consistency checker to the CI
- Change the consistency checker so it won't validate modules that are invalid on purpose.

## Checklist

      hard-to-understand areas.


abizjak (approved):
Excellent.

PR #336, opened 2023-03-09
Ensure Cargo.lock files are up-to-date.

abizjak:
## Purpose

It happens too many times that only some lock files are updated when dependencies are updated.

This addition should ensure that the CI catches those cases.

## Checklist

      hard-to-understand areas.


EmilLa1 (approved):
:+1:

PR #338, opened 2023-03-13
Fix CIS2 token Display implementation.

abizjak:
## Purpose

Fix CIS2 token event Display implementation.

## Checklist

      hard-to-understand areas.


limemloh:
One issue here, you'll need some information from the metadata for displaying the amount.
But this is the best we can do

abizjak:
> One issue here, you'll need some information from the metadata for displaying the amount. But this is the best we can do

Yeah, we can't display it in decimal. However it is better to have the information available. I had a small script for something else and I could not see the amounts printed which limited the utility.

We could have (maybe we already do) have a separate function `fn display(&self, num_decimals: u8) -> impl Display` or some such.

Issue #352, opened 2023-03-30
Upgrade tooling in `concordium-base`.


vikt0r0:
- Upgrade the stackage snapshot to `lts-20.16` and GHC to `9.2.7`.
- Bump `rustc` to `1.68`.

PR #353, opened 2023-03-30
Upgrade tooling

vikt0r0:
## Purpose

Resolves https://github.com/Concordium/concordium-base/issues/352

Depends on https://github.com/Concordium/concordium-contracts-common/pull/80

## Changes

- Upgrade the stackage snapshot to `lts-20.16` and GHC to `9.2.7`.
- Bump `rustc` to `1.68`.

## Checklist

      hard-to-understand areas.


smart-contracts/wasm-chain-integration/src/v1/ffi.rs
```diff
@@ -478,7 +476,7 @@ unsafe extern "C" fn resume_receive_v1(
         // the state has changed, so this is crucial.
         let state_ref = std::mem::replace(&mut *state_ptr_ptr, std::ptr::null_mut());
         // The clone is cheap since this is reference counted.
-        let mut state = (&*state_ref).clone();
+        let mut state = (*state_ref).clone();
```
abizjak, line 479: Clone is very overloaded so this is not obviously correct as a change. I'm on my phone so cannot check now but please make sure that the right clone is called here. It probably is but I want you to make sure.

abizjak: Stating the type of stateref explicitly would help make this clearer.

abizjak (approved):
Please make sure to properly address the 4 comments. In particular the one about clone. The rest looks good.

smart-contracts/wasm-chain-integration/src/v1/ffi.rs
vikt0r0: So just to confirm, this should be `*mut MutableState`, no?

smart-contracts/wasm-chain-integration/src/v0/types.rs
```diff
@@ -202,7 +202,7 @@ impl InitResult {
             } => {
                 let mut out = Vec::with_capacity(5 + state.len() as usize + 8);
                 out.push(2);
-                out.extend_from_slice(&(state.len() as u32).to_be_bytes());
+                out.extend_from_slice(&state.len().to_be_bytes());
```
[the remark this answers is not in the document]
vikt0r0: I believe it does, but perhaps it is better to `#[allow(clippy::unnecessary_cast)]` if it is not obvious?

rust-src/concordium_base/src/aggregate_sig/ffi.rs
```diff
@@ -77,7 +77,7 @@ pub extern "C" fn bls_sign(
     m_len: size_t,
     sk_ptr: *mut SecretKey<Bls12>,
 ) -> *mut Signature<Bls12> {
-    let m_len = m_len as usize;
+    let m_len = m_len;
```
td202, line 80: With this change, the assignment seems entirely unnecessary, and similarly for the others in this file. However, I am not entirely sure that removing the cast is the right option. I think for all platforms we care about, we can rely on `usize` and `size_t` being the same, but it seems at least historically we have tried to avoid making this assumption. I think probably the thing to do is keep the cast and suppress the warning.

rust-src/concordium_base/src/eddsa_ed25519/ffi.rs
```diff
@@ -56,7 +56,7 @@ extern "C" fn eddsa_public_to_bytes(
 #[no_mangle]
 #[allow(clippy::not_unsafe_ptr_arg_deref)]
 extern "C" fn eddsa_public_from_bytes(input_bytes: *mut u8, input_len: size_t) -> *mut PublicKey {
-    let len = input_len as usize;
+    let len = input_len;
```
td202, line 59: As in the other ffi, we should probably retain the case from `size_t` to `usize` in case they are different.

rust-src/concordium_base/src/encrypted_transfers/ffi.rs
```diff
@@ -189,8 +189,8 @@ unsafe extern "C" fn verify_encrypted_transfer(
         encryptions: [*transfer_low, *transfer_high],
     };
 
-    let transfer_proof = slice_from_c_bytes!(transfer_proof_ptr, transfer_proof_len as usize);
```
td202, line 192: As in other ffi, we may wish to retain casts from `size_t` to `usize`.

rust-src/concordium_base/src/id/ffi.rs
```diff
@@ -32,7 +32,7 @@ extern "C" fn verify_initial_cdi_ffi(
     initial_cdi_len: size_t,
     expiry: u64,
 ) -> i32 {
-    let cdi_bytes = slice_from_c_bytes!(initial_cdi_ptr, initial_cdi_len as usize);
```
td202, line 35: As before, we may want to retain these casts.

smart-contracts/wasm-chain-integration/src/v0/ffi.rs
```diff
@@ -37,19 +37,19 @@ unsafe extern "C" fn call_init_v0(
     energy: InterpreterEnergy,
     output_len: *mut size_t,
 ) -> *mut u8 {
-    let artifact_bytes = slice_from_c_bytes!(artifact_ptr, artifact_bytes_len as usize);
```
td202, line 40: Possibly retain coercion (see other comments).

smart-contracts/wasm-chain-integration/src/v1/ffi.rs
```diff
@@ -116,7 +116,7 @@ unsafe extern "C" fn call_init_v1(
     output_len: *mut size_t,
     output_state_ptr: *mut *mut MutableState,
 ) -> *mut u8 {
-    let artifact_bytes = slice_from_c_bytes!(artifact_ptr, artifact_bytes_len as usize);
```
td202, line 119: Possibly retain coercion (see other comments).

td202 (commented):
For the most part, this looks fine. My main concern is the issue of `size_t` being treated as if it's the same as `usize`, and I think should be addressed one way or the other.

rust-src/concordium_base/src/aggregate_sig/ffi.rs
abizjak: https://docs.rs/libc/latest/libc/type.size_t.html

On all platforms that I can find there size_t is defined as usize (as in type alias). Since the only reason we need to convert is so we can index in Rust there does not seem to be a downside to avoiding this cast.

If at some point Rust and Haskell add support for a platform where size_t will not be usize then that will lead to a compilation error for this module. So I think it is safe to remove the cast.

vikt0r0: Good point, I removed them.

smart-contracts/wasm-chain-integration/src/v0/types.rs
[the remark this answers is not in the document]
vikt0r0: The warning happens since the implementation is `pub fn len(&self) -> u32 { self.state.len() as u32 }` so casting this is unnecessary.

PR #354, opened 2023-04-09
Fix wasm test.

abizjak:
The return value u64::MAX is returned when there is no next value for the iterator. u64::MAX with the second bit unset is returned in case of "error" (such as the iterator not existing).

## Checklist

      hard-to-understand areas.


Bargsteen (approved):
Looks fine, but how did the test succeed before?

abizjak:
> Looks fine, but how did the test succeed before?

No changes were made, which is what this test is checking. Not that the invocation actually succeeded. We check that b_forward succeeds, which it always does.

PR #371, opened 2023-05-11
Web3id key derivation

orhoj:
## Purpose
Add support for deriving keys for verifiable credentials. In particular also expose this functionality to the mobile libraries for when they are going to implement Web3 ID functionality.

## Changes
- Updated `key_derivation` with methods for getting the public key, signing key and encryption key required for a verifiable credential.
- Updated mobile libraries to expose the new functionality in `key_derivation`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/mobile_wallet/CHANGELOG.md
+++ b/mobile_wallet/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.25.0
+- Added a function `get_verifiable_credential_keys` for deriving the keys for a verifiable credential.
+
 ## 0.24.0
 - Changed parameter_to_json to support parameter schemas.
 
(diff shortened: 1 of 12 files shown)


orhoj:
As part of reviewing this, in particular @tschudid and @chrmatt, please compare to what is written in the bluepaper. 

I was considering if it would be nicer to move the encryption key to index `1` as we are not using attribute randomness at this time.

chrmatt (approved):
I can confirm that this matches the description in the bluepaper. I'm fine with changing the index for the encryption key to 1 instead of 2. Let me know if you want to do that, then I can update the bluepaper accordingly.

abizjak (approved):
Looks good.

tschudid (approved):
Looks good to me.
I am also fine with changing the indices.

orhoj:
> 

I moved the encryption key to index `1`, so please update the bluepaper.

PR #373, opened 2023-05-12
Make sure chrono is sufficiently up to date.

abizjak:
## Purpose

Make sure that the manifest file requires a sufficiently recent version of chrono since we use Months which has only been introduced recently.

## Checklist

      hard-to-understand areas.


rust-src/concordium_base/Cargo.toml
```diff
@@ -20,7 +20,7 @@ rand_core = "=0.5"
 serde = {version = "1.0", features = ["derive"]}
 serde_json = "1.0"
 libc = "0.2"
-chrono = "0.4"
```
limemloh, line 23: Consider adding a comment to why this is the right version.
In a month or two someone (probably me) will think we can remove the patch version

limemloh (approved):

rust-src/concordium_base/Cargo.toml
abizjak: https://en.wiktionary.org/wiki/Chesterton%27s_fence :) 

I have added comments to toml files.

Issue #327, opened 2023-02-28
New Consensus: M1 Fix TimeoutParameters.


EmilLa1:
First problem currently is, that we do not check whether the denominator is zero when deserializing the `TimeoutParameters` structure.

This should be fixed and an appropriate error should be returned if it is the case.

Secondly we should probably not use `Ratio` for these values because `Ratio` does not play along well with silent overflows. 

e.g. 

```
ghci> (251 % 252 :: Ratio Word8) >= 2 % 3
False
```

So let's introduce our own type for these and expose the few capabilities we need.

abizjak:
What do you mean it does not play along with multiplication? How does the example you have demonstrate that?

EmilLa1:
Yea sorry for vague description, but it's due to silent overflow e.g.

```
(251 % 253 :: Ratio Word8) >= 2 % 3  <=/>   251 * 3    >= 2 * 252  =>  241 >= 248
;```

**Comment by EmilLa1** [comment]
More context: https://gitlab.haskell.org/ghc/ghc/-/issues/19500

**Comment by abizjak** [comment]
Very nice. :P 

The comment
```
(251 % 253 :: Ratio Word8) > 2 % 3
```
was confusing since, well, that relation is correct :P

abizjak:
Is this still relevant or was it fixed?

EmilLa1:
Yes. 

I have posted this PR which addresses the missing denominator checks. 

However, I have not introduced a new type *yet* to use instead of the `Ratio` as I don't think it matters when we normalise parameters.

If you think it's worth to add a new type, then I'll do it.

EmilLa1:
Closing this now as https://github.com/Concordium/concordium-base/pull/390 is merged into main. 

We should however be careful in the future if we start to compare the growth and shrink factors after the deserialization.

PR #390, opened 2023-06-11
Check denominator when deserializing TimeoutParameters

EmilLa1:
## Purpose

Partly fixes https://github.com/Concordium/concordium-base/issues/327 by checking the denominators when deserializing the `TimeoutParameters`.

## Changes

Introduced two checks that make sure that we fail properly if the provided denominator of either the `timeoutIncrease` or `timeoutDecrease` are zero.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Types/Parameters.hs
+++ b/haskell-src/Concordium/Types/Parameters.hs
@@ -1512,12 +1512,14 @@ instance Serialize TimeoutParameters where
         -- Get the timeout increase ratio.
         tiNum <- get
         tiDen <- get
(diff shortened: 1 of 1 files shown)


abizjak:
Regarding the other, overflow property, where is the comparison of timeout parameters relevant?

We only compare to 0/1 in deserialization which is always fine.

EmilLa1:
@abizjak  We don't do it (compare the growth and shrink factor) elsewhere - that's why I didn't introduce a new type for it now.

abizjak (approved):

PR #400, opened 2023-06-25
Add support for sign extension instructions.

### LINKED ISSUE #386: Add support for mutable globals and sign extension instructions
**Task description**

In Rust 1.70 the default code generation (that is not possible to override) has changed so that the emitted code can contain sign extension instructions not present in the MVP version.

See https://releases.rs/docs/1.70.0/#compatibility-notes for the changelog.

Since these instructions are a conservative extension we should support them since that is the simplest way forward to ensure compatibility with newer Rust versions, but this requires a protocol update.

See also https://github.com/rust-lang/rust/issues/109807 for a tracking issue on why the behaviour cannot be disabled.

Followup tasks:
- [ ] Once this task is completed and the node is released with the new protocol version we must remove the mentions of Rust 1.69 from the documentation.

abizjak:
## Purpose

Add support for the sign extension operator proposal, see https://github.com/WebAssembly/sign-extension-ops/blob/master/proposals/sign-extension-ops/Overview.md

## Changes

- Add another option to ValidationConfig whether to support or not the sign extension instructions. Sign extension instructions are allowed or not here. The interpreter always supports them.
- Implement support for sign extension instructions in the execution engine.
- Add benchmarks to make sure the cost assignment of the new instructions is adequate.

Closes #386 

## Checklist

      hard-to-understand areas.


smart-contracts/wasm-transform/src/validate.rs
```diff
@@ -934,7 +946,7 @@ pub fn validate_module(
     // The global section is valid as long as it's well-formed.
     // We already check that all the globals are initialized with
     // correct expressions.
-    let global: GlobalSection = parse_sec_with_default(EMPTY_CTX, &skeleton.global)?;
+    let global: GlobalSection = parse_sec_with_default(config, &skeleton.global)?;
     ensure!(
         global.globals.len() <= MAX_NUM_GLOBALS,
         "The number of globals must not exceed {}.",
```
limemloh, line 952: Don't we also need the non-empty context when parsing the functions section further below?
I expect that we do, since (to my understanding) it can contain instructions

limemloh: Okay, having read a bit more about how the binary representation is, I take this back.
I see now that all the instructions are put in the code section

limemloh (approved):

smart-contracts/wasm-transform/src/validate.rs
abizjak: Function section does not contain the code. It contains the type declarations of functions, i.e., 

```
void myFunction();
```

The code section contains instructions, and we pass in the relevant context to the instruction parser

smart-contracts/wasm-transform/src/validate.rs
```diff
@@ -991,8 +1003,11 @@ pub fn validate_module(
                     memory: memory.memory_type.is_some(),
                     table: table.table_type.is_some(),
                 };
-                let (opcodes, max_height) =
-                    validate(&ctx, &mut OpCodeIterator::new(c.expr_bytes), Vec::new())?;
+                let (opcodes, max_height) = validate(
+                    &ctx,
+                    &mut OpCodeIterator::new(config.allow_sign_extension_instr, c.expr_bytes),
```
abizjak, line 1008: @limemloh Here.

smart-contracts/wasm-transform/src/validate.rs
abizjak: I submitted my comment without seeing your own second comment. All good then.

smart-contracts/wasm-transform/src/parse.rs
```diff
@@ -815,8 +849,11 @@ impl<'a> Parseable<'a, Option<&GlobalSection>> for Data {
     }
 }
 
-impl<'a> Parseable<'a, Option<&GlobalSection>> for DataSection {
-    fn parse(ctx: Option<&GlobalSection>, cursor: &mut Cursor<&'a [u8]>) -> ParseResult<Self> {
+impl<'a> Parseable<'a, InstructionValidationContext<'_>> for DataSection {
```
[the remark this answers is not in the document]
limemloh: Right, so it seems that all the implementations are either of a generic context or this `InstructionValidationContext`, but seeing that we need it to be empty for some of the sections this is probably how it has to be.

Bargsteen (approved):
LGTM

smart-contracts/wasm-transform/src/validate.rs
```diff
@@ -871,17 +879,21 @@ pub struct ValidationConfig {
     /// data and element sections. In protocols 1-5 this was allowed, but we
     /// need to disallow it in following protocols since the Wasm spec has been
     /// updated to not allow this anymore. See [issue](https://github.com/WebAssembly/spec/issues/1522) on the Wasm spec repository.
-    pub allow_globals_in_init: bool,
+    pub allow_globals_in_init:      bool,
+    /// Allow sign extension instructions.
```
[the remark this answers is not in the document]
abizjak: Added

limemloh (approved):

smart-contracts/wasm-transform/src/parse.rs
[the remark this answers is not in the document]
abizjak: Some are also for `ValidationConfig`.

We could probably do things differently, but I think this design is reasonable.

Issue #388, opened 2023-06-07
Add account key signature checking to V1 smart contracts


abizjak:
**Task description**

Expose account keys and checking signatures with account keys to smart contracts

Add two new operations to be called via `invoke` to check an account signature, and get the account keys.

PR #404, opened 2023-06-29
Support for checking account signatures, and retrieving account public keys.

abizjak:
## Purpose

Adds support for implementing https://github.com/Concordium/concordium-base/issues/388 in the node.

## Checklist

      hard-to-understand areas.


haskell-src/Concordium/Cost.hs
```diff
@@ -264,3 +264,24 @@ contractInstanceQueryContractBalanceCost = 200
 -- |Cost of querying the current exchange rates from a within smart contract instance.
 contractInstanceQueryExchangeRatesCost :: Energy
 contractInstanceQueryExchangeRatesCost = 100
+
+-- |Base cost of querying the account keys from a within smart contract instance.
```
Bargsteen, line 268:
```suggestion
-- |Base cost of querying the account keys from within smart contract instance.
```

smart-contracts/testdata/contracts/v1/account-signature-checks.wat
```diff
@@ -0,0 +1,52 @@
...
+           (then nop)
+           (else unreachable)))
+
+ ;; Contract
+
+ ;; Initialize contract.
+ (func $init (export "init_contract") (param i64) (result i32)
+       (return (i32.const 0))) ;; Successful init
+
+ ;; Retrieve the parameter and get the account keys for the parameter
```
Bargsteen, line 28:
```suggestion
 ;; Retrieve the parameter (an account address) and get the account keys for the parameter. Then return the keys.
```

smart-contracts/testdata/contracts/v1/account-signature-checks.wat
```diff
@@ -0,0 +1,52 @@
...
+      (call $get_parameter_section (i32.const 0) (i32.const 0) (i32.const 32) (i32.const 0))
+      ;; query account keys, tag of the operation is 6.
+      (call $invoke (i32.const 6) (i32.const 0) (i32.const 32))
+      ;; The keys are output as parameter 1. Read them into linear memory.
+      (call $get_parameter_section (i32.const 1) (i32.const 0) (call $get_parameter_size (i32.const 1)) (i32.const 0))
+      ;; Output 4 bytes from linear memory
+      (call $write_output (i32.const 0) (call $get_parameter_size (i32.const 1)) (i32.const 0))
+      (return (i32.const 0)))
+
+ ;; Get the signature and data from the parameter, and invoke the check signature operation.
```
Bargsteen, line 40:
```suggestion
 ;; Get the signature and data from the parameter, and invoke the check signature operation. Then return the signature check result.
```

smart-contracts/wasm-chain-integration/src/v1/ffi.rs
```diff
@@ -237,6 +237,8 @@ unsafe extern "C" fn call_receive_v1(
     // in case execution terminated normally. Normally here means without a runtime exception.
     output_state_changed: *mut u8,
     support_queries_tag: u8, // non-zero to enable support of chain queries.
+    support_account_signature_checks: u8, /* non-zero to enable support for queryin account keys
```
Bargsteen, line 240:
```suggestion
    support_account_signature_checks: u8, /* non-zero to enable support for querying account keys
```

smart-contracts/wasm-chain-integration/src/v1/mod.rs
```diff
@@ -303,6 +335,8 @@ mod host {
     const QUERY_ACCOUNT_BALANCE_TAG: u32 = 2;
     const QUERY_CONTRACT_BALANCE_TAG: u32 = 3;
     const QUERY_EXCHANGE_RATE_TAG: u32 = 4;
+    const CHECK_ACCOUNT_SIGNATURE_TAG: u32 = 5;
+    const QUERY_ACCOUNT_KEYS_TAG: u32 = 6;
```
Bargsteen, line 339: Why are these 5 and 6, when they are 6 and 7 in `to_bytes` above? Seems like a constant for `upgrade` is missing.

smart-contracts/wasm-chain-integration/src/v1/mod.rs
```diff
@@ -486,6 +522,43 @@ mod host {
                 );
                 Ok(Interrupt::QueryExchangeRates.into())
             }
+            CHECK_ACCOUNT_SIGNATURE_TAG if support_account_signature_checks => {
+                ensure!(
+                    length >= ACCOUNT_ADDRESS_SIZE,
+                    "Account balance queries must have exactly 32 bytes of payload, but was {}",
```
Bargsteen, line 528:
```suggestion
                    "Account signature check queries must have at least the 32 bytes for an account address, but was {}",
```

smart-contracts/wasm-chain-integration/src/v1/mod.rs
```diff
@@ -486,6 +522,43 @@ mod host {
...
+                    length
+                );
+                // Overflow is not possible in the next line on 64-bit machines.
+                ensure!(start + length <= memory.len(), "Illegal memory access.");
+                let mut addr_bytes = [0u8; ACCOUNT_ADDRESS_SIZE];
+                addr_bytes.copy_from_slice(&memory[start..start + ACCOUNT_ADDRESS_SIZE]);
+                let address = AccountAddress(addr_bytes);
+                if energy.tick_energy(constants::copy_parameter_cost(length_u32)).is_err() {
+                    bail!(OutOfEnergy);
+                }
```
Bargsteen, line 538: We could check this before copying the account address so we avoid doing unnecessary work when the energy is too low.

smart-contracts/wasm-chain-integration/src/v1/mod.rs
```diff
@@ -486,6 +522,43 @@ mod host {
...
+                Ok(Interrupt::CheckAccountSignature {
+                    address,
+                    payload,
+                }
+                .into())
+            }
+            QUERY_ACCOUNT_KEYS_TAG if support_account_signature_checks => {
+                ensure!(
+                    length == ACCOUNT_ADDRESS_SIZE,
+                    "Account balance queries must have exactly 32 bytes of payload, but was {}",
```
Bargsteen, line 549:
```suggestion
                    "Account keys queries must have exactly 32 bytes of payload, but was {}",
```

smart-contracts/wasm-chain-integration/src/v1/mod.rs
Bargsteen: I see that we don't charge for moving the account keys below, so perhaps this is consistent afterall.

smart-contracts/wasm-chain-integration/src/v1/mod.rs
```diff
@@ -1394,6 +1468,10 @@ pub enum InvokeFailure {
     UpgradeInvalidContractName,
     /// Attempt to upgrade a V1 contract to a V0 contract.
     UpgradeInvalidVersion,
+    /// Could not parse the signature and message
```
Bargsteen, line 1471:
```suggestion
    /// Could not parse the signature and message.
```

smart-contracts/wasm-chain-integration/src/v1/mod.rs
```diff
@@ -1708,31 +1790,46 @@ where
...
+    pub max_parameter_size:               usize,
     /// Whether the amount of logs a contract may produce, and the size of the
     /// logs, is limited.
-    pub limit_logs_and_return_values: bool,
+    pub limit_logs_and_return_values:     bool,
     /// Whether queries should be supported or not. Queries were introduced in
     /// protocol 5.
-    pub support_queries:              bool,
+    pub support_queries:                  bool,
+    /// Whether querying account public keys checking account signatures is
```
Bargsteen, line 1800:
```suggestion
    /// Whether querying account public keys and checking account signatures is
```

Bargsteen (commented):
Looks good, I just have a question about the tags used.

smart-contracts/wasm-chain-integration/src/v1/mod.rs
abizjak: Right, the reason for that is that `upgrade` is not handled by the `invoke`, it is instead a separate host function so that it is visible from the contract code that it is being called.

But internally we handle it via the `invoke` handler. 

The other tags in `to_bytes` are purely an internal thing in how we pass data through FFI.  Whereas the tags here are part of the smart contract API.

smart-contracts/wasm-chain-integration/src/v1/mod.rs
abizjak: That's a good point, we should do it before when we can.

smart-contracts/wasm-chain-integration/src/v1/mod.rs
limemloh: There are two places where we have to serialize the invoke call, Wasm to rust and rust to Haskell, since upgrade is a separate host function there is no tag for it in Wasm to rust, but there still is in the rust to Haskell part.

Bargsteen: Ah ok, thanks for the explanation. 👍

limemloh (approved):

Bargsteen (approved):
Assuming the minor comments about comments will be fixed, this is good to go ;)

smart-contracts/wasm-chain-integration/src/v1/mod.rs
abizjak: That's changed now.

PR #416, opened 2023-08-02
Revise credentials

abizjak:
## Purpose

Revise credentials

## Changes

- The issuer now additionally signs the contract address.
- The JSON format of Web3IdCredential now follows W3C "standard".
- There are no more attribute tags (u8). String names are used in proofs and inputs now for Web3Id stuff.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/src/web3id/mod.rs
+++ b/rust-src/concordium_base/src/web3id/mod.rs
@@ -11,6 +11,7 @@ pub mod did;
 use crate::{
     base::CredentialRegistrationID,
     cis4_types::IssuerKey,
+    common::{base16_decode_string, base16_encode_string},
     curve_arithmetic::Curve,
     id::{
         constants::{ArCurve, AttributeKind},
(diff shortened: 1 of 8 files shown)


eb-concordium (approved):

rust-src/concordium_base/src/web3id/mod.rs
```diff
@@ -1059,7 +1059,12 @@ impl CredentialSchema {
     }
 }
```
annenkov, line 1060: Seems like `CredentialSchema` is not used anywhere now, at least not in this file. Maybe it's imported somewhere else, but `attribute_tag` seems irrelevant since tags are not indices now.

annenkov (approved):
Looks good.

rust-src/concordium_base/src/web3id/mod.rs
abizjak: Forgot to delete this. Will remove.

PR #423, opened 2023-08-09
Get block certificates

EmilLa1:
## Purpose

Introduced types to support the `GetBlockCertificates` endpoint.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Types/KonsensusV1.hs
+++ b/haskell-src/Concordium/Types/KonsensusV1.hs
@@ -39,9 +39,9 @@ data TimeoutCertificate = TimeoutCertificate
       -- |The minimum epoch for which we include signatures.
       tcMinEpoch :: !Epoch,
       -- |The rounds for which finalizers have their best QCs in the epoch 'tcMinEpoch'.
-      tcFinalizerQCRoundsFirstEpoch :: !FinalizerRound,
+      tcFinalizerQCRoundsFirstEpoch :: ![FinalizerRound],
       -- |The rounds for which finalizers have their best QCs in the epoch @tcMinEpoch + 1@.
-      tcFinalizerQCRoundsSecondEpoch :: !FinalizerRound,
+      tcFinalizerQCRoundsSecondEpoch :: ![FinalizerRound],
       -- |Aggregate of the finalizers' 'TimeoutSignature's on the round and QC round.
       tcAggregateSignature :: !Bls.Signature
     }
--- a/haskell-src/Concordium/Types/KonsensusV1.hs
+++ b/haskell-src/Concordium/Types/KonsensusV1.hs
@@ -1,13 +1,11 @@
-{-# LANGUAGE DeriveTraversable #-}
-
 -- |Types that are relevant only for ConsensusV1.
--- Common for types defined here is that they are
--- exposed through the node API, otherwise they should
--- just be defined in the node.
-module Concordium.Types.ConsensusV1 where
+-- Common for types defined here is that they are exposed through the node API,
+-- otherwise definitions should be defined in the node.
+module Concordium.Types.KonsensusV1 where
 
 import Control.Monad
 import Data.Bits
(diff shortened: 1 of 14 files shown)


haskell-src/Concordium/Types/KonsensusV1.hs
```diff
@@ -0,0 +1,134 @@
...
+        object
+            [ "finalizedQC" .= efeFinalizedQC,
+              "successorQC" .= efeSuccessorQC,
+              "successorProof" .= efeSuccessorProof
+            ]
+
+-- |Block certificates for a block in 'ConsensusV1'.
+data BlockCertificates = BlockCertificates
+    { -- |Quorum certificate for the block.
+      -- This is only present if the block is not a genesis block.
```
abizjak, line 115:
```suggestion
      -- This is present if and only if the block is not a genesis block.
```

haskell-src/Concordium/Types/KonsensusV1.hs
```diff
@@ -0,0 +1,134 @@
...
+
+instance ToJSON EpochFinalizationEntry where
+    toJSON EpochFinalizationEntry{..} =
+        object
+            [ "finalizedQC" .= efeFinalizedQC,
+              "successorQC" .= efeSuccessorQC,
+              "successorProof" .= efeSuccessorProof
+            ]
+
+-- |Block certificates for a block in 'ConsensusV1'.
```
abizjak, line 112: Is it `C` or `K` then?

abizjak (approved):

haskell-src/Concordium/Types/KonsensusV1.hs
```diff
@@ -0,0 +1,134 @@
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+-- |Types that are relevant only for endpoints exposed in ConsensusV1.
```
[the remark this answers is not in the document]
EmilLa1: Are you referring to the module name?

The module is called "KonsensusV1"  - because that is how it's named in the node repository. 
I.e. the types that are quite similar to the ones defined here.. (`Concordium/Types/KonsensusV1.hs`)

'ConsensusV1' is what we call the actual versioned parameter in `Parameters.hs`, i.e. `ConsensusVersion` in `Parameters.hs`.


:man_shrugging:  I can rename it to `ConsensusV1.hs` i.e. the module - I'm fine either way. @td-concordium got an opinion on this matter?

haskell-src/Concordium/Types/KonsensusV1.hs
EmilLa1: See above.

haskell-src/Concordium/Types/KonsensusV1.hs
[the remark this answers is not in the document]
abizjak: I don't care enough to change naming now since it's obviously going to cause confusion elsewhere.

But it is confusing to have both K and C in various places for the same thing.

haskell-src/Concordium/Types/KonsensusV1.hs
```diff
@@ -0,0 +1,134 @@
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+-- |Types that are relevant only for endpoints exposed in 'ConsensusV1'.
+module Concordium.Types.KonsensusV1 where
```
td202, line 5: I think it might be best to rename to `Concordium.Types.Queries.KonsensusV1` to emphasise that these types exist for supporting queries, and are distinct to the types used internally by the consensus.

haskell-src/Concordium/Types/KonsensusV1.hs
td202: As above, recommend "consensus version 1".

td202 (approved):

PR #426, opened 2023-08-15
Support for epoch winning bakers and first block queries.

td202:
## Purpose

Add support for getting the winning bakers of an epoch and the first block of an epoch.

## Changes

- Update `concordium-grpc-api`.
- Add `EpochRequest` and `WinningBaker` types, with `ToProto` support.
- Implement an interpolation search. This is used for finding the first block of an epoch, which is used by both queries.

## Checklist

      hard-to-understand areas.


concordium-base.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version: 2.0
 
--- This file has been generated from package.yaml by hpack version 0.35.0.
+-- This file has been generated from package.yaml by hpack version 0.35.1.
```
EmilLa1, line 3: Just out of curiosity, is this because of a new stack version we're using or?

EmilLa1 (approved):
Looking good!

haskell-src/Concordium/Utils/InterpolationSearch.hs
```diff
@@ -0,0 +1,66 @@
...
+    (index -> m (key, value)) ->
+    -- |Key to search for @target@.
+    key ->
+    -- |Lower bound of the search range @(lowIx, (lowKey, lowVal))@.
+    (index, (key, value)) ->
+    -- |Upper bound of the search range @(highIx, (highKey, highVal))@.
+    (index, (key, value)) ->
+    m (Maybe (index, value))
+{-# INLINE interpolationSearchFirstM #-}
+interpolationSearchFirstM lookupIx target low@(lowIx, (lowKey, lowVal)) high@(_, (highKey, _))
```
abizjak, line 33: Could you please add some unit tests for this.

haskell-src/Concordium/Types/Queries.hs
```diff
@@ -1028,3 +1028,30 @@ instance ToJSON BakerRewardPeriodInfo where
...
+          erEpoch :: !Epoch
+        }
+    | -- |Query the epoch of a specified block.
+      EpochOfBlock
+        { -- |The block whose epoch is to be used for the query.
+          erBlock :: !BlockHashInput
+        }
+
+-- |Details of which baker won the lottery in a given round in consensus version 1.
+data WinningBaker = WinningBaker
```
abizjak, line 1049: We should probably add ToJSON instance for this so we can display it in the client.

abizjak (approved):

concordium-base.cabal
td202: Yes. `stack` version 2.9.3 uses `hpack` 0.35.1.

haskell-src/Concordium/Utils/InterpolationSearch.hs
td202: Done.

haskell-src/Concordium/Types/Queries.hs
td202: Done.

PR #428, opened 2023-08-16
ZKP refactoring

chrmatt:
## Purpose

This PR moves all sigma protocols into a common folder and moves the functions for range proofs to the bulletproof crate.

## Changes

- Move all sigma protocols from ID to a common sigma-protocol crate.
- Move all sigma protocols from Encrypted Transfers to the common sigma-protocol crate.
- Move the helper functions for proving and verifying range proofs from `id_prover` and `id_verifier` to `range_proofs`.

**Note:** The sigma protocol in `dlog_ed25519` was not moved because it does not implement the SigmaProtocol trait and is specific to that curve.

## Checklist

      hard-to-understand areas.


hamiidreza (approved):
Looks good!

abizjak (approved):
Please update the changelog as well.

PR #442, opened 2023-09-13
Merge Concordium Contracts Common

Bargsteen:
## Purpose

Merge Concordium Contracts Common(CCC) into base.

Currently, the only place CCC is used as a submodule on its own is in the rust-smart-contracts repo. But soon we will move the integration testing library from the contract-tools repo to rust-smart-contracts. At that point, no repositories depend on contracts common in isolation. Thus the merge.

## Changes

- Merge in the repo with its full history to the folder `smart-contracts/contracts-common` (the subfolders still have a `concordium-` prefix)
- Merge the Github workflow actions

## Checklist

      hard-to-understand areas.


smart-contracts/contracts-common/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+- ```cargo clippy --all``` produces no warnings
+- ```rustfmt``` makes no changes.
+
+Everything in this repository should build with rust version 1.60 however the `fmt` tool must be from a nightly release since some of the configuration options are not stable. One way to run the `fmt` tool is
+```
+cargo +nightly-2023-04-01 fmt
+```
+
+(the exact version used by the CI can be found in [.github/workflows/linter.yml](.github/workflows/linter.yml) file).
```
limemloh, line 47: Link needs to be fixed, but maybe better is to remove this section since base have it already

smart-contracts/contracts-common/concordium-contracts-common/Cargo.toml
```diff
@@ -0,0 +1,101 @@
+[package]
+name = "concordium-contracts-common"
+version = "8.0.0"
+authors = ["Concordium <developers@concordium.com>"]
+edition = "2021"
+rust-version = "1.65"
+license = "MPL-2.0"
+description = "Common functionality used by smart contracts and the host environment on the Concordium blockchain."
+homepage = "https://github.com/Concordium/concordium-contracts-common"
+repository = "https://github.com/Concordium/concordium-contracts-common"
```
limemloh, line 10: Needs to be updated

limemloh (approved):
LGTM assuming everything builds

abizjak (approved):
Looks good, but please make a separate workflow.yml for checking contracts-common so we can only run it when it has changed.

abizjak (changes requested):
Actually I now looked at the commits and if I understand correctly you had one massive commit renaming files to put them under contracts-common. This is less than ideal since git blame becomes harder, and we lose tags.

It would be better to use git filter-repo to just move everything and rewrite history that way.
We should ideally also copy over the release tags we used, and then also push those tags.

I think git-filter-repo should make this relatively easy and this is what we've done before.

https://github.com/newren/git-filter-repo#solving-this-with-filter-repo

Bargsteen:
> Actually I now looked at the commits and if I understand correctly you had one massive commit renaming files to put them under contracts-common. This is less than ideal since git blame becomes harder, and we lose tags.
> 
> It would be better to use git filter-repo to just move everything and rewrite history that way. We should ideally also copy over the release tags we used, and then also push those tags.
> 
> I think git-filter-repo should make this relatively easy and this is what we've done before.
> 
> https://github.com/newren/git-filter-repo#solving-this-with-filter-repo

Ah, I didn't notice that you could see it on the files. Thanks. I've created https://github.com/Concordium/concordium-base/pull/443 instead, in which I've used filter-repo and pushed the tags.

PR #445, opened 2023-09-19
mobile_wallet: Use conventional naming for transaction types

bisgardo:
## Purpose

The dApp lib `@concordium/wallet-connectors` was changed to use lower case transaction type in https://github.com/Concordium/concordium-dapp-libraries/pull/42 to match the convention set by wallet-proxy and the internal types ([iOS](https://github.com/Concordium/concordium-reference-wallet-ios/blob/7200500f64084b8c0ce64ccc701f092859a3c6c1/ConcordiumWallet/Model/TransferType.swift)) of the mobile wallets.

This PR aligns the change with the crypto library to use the same type names.

## Changes

Make the mobile wallet crypto library use the conventional type names (in camelCase).

Part of https://concordium.atlassian.net/browse/CBW-1360.

### DIFF
--- a/mobile_wallet/src/lib.rs
+++ b/mobile_wallet/src/lib.rs
@@ -250,6 +250,7 @@ enum SpecifiedEnergy {
 
 #[derive(common::SerdeDeserialize)]
 #[serde(tag = "type", content = "payload")]
+#[serde(rename_all = "camelCase")]
 enum JSONPayload {
     InitContract {
         #[serde(flatten)]


abizjak:
This change would mean that we are introducing an incompatibility that we need to deal with. Why are we doing this exactly?

It would mean that the new wallet only supports the camelcase format, and the old wallet supports the old format.  What is a dapp going to do?

bisgardo:
> This change would mean that we are introducing an incompatibility that we need to deal with. Why are we doing this exactly?

The wallets are currently not calling the lib with the correct type names. It's not introducing an incompatibility, it's part of fixing the currently incompatibility.

The problems are not the same on iOS and Android and it's of course part of the task to ensure that both work.

> It would mean that the new wallet only supports the camelcase format, and the old wallet supports the old format.

The dapp payload doesn't/shouldn't go here directly. It gets parsed into the internal type that I linked in the description (iOS). It's the internal type that gets converted into the payload for this lib.

> What is a dapp going to do?

Use the new type names. The special case `"Update"` will be supported for backwards comp.

PR #453, opened 2023-10-01
Fix VersionedModuleSchema Display given multiple contracts

schwartz-concordium:
## Purpose

When invoking `Display` on a `VersionedModuleSchema` with multiple contracts only the last contract was shown. This because contracts wasn’t appended to final output string in loop.

Also fix test since they didn’t run green. From the PR where they where added, and from this comment, https://github.com/Concordium/concordium-contracts-common/pull/99#discussion_r1250555893, I can see that at some time in the development the indention was set to 30. If I change the indention to 30 the test runs green. Hence I guess the test has been forgotten to run after the indention was changed from 30 -> 11.

## Changes
- Changed loop to append contracts to final string
- Fix test which wasn’t running green.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
+++ b/smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased changes
+- Fixed `Display` trait on `VersionedModuleSchema` when module contained multiple contracts to render all of them.
+- Add `--all-features` to contracts common CI test execution.
 
 ## concordium-contracts-common 8.0.0 (2023-08-21)
 
--- a/smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
+++ b/smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## Unreleased changes
 - Fixed `Display` trait on `VersionedModuleSchema` when module contained multiple contracts to render all of them.
-- Add `--all-features` to contracts common CI test execution.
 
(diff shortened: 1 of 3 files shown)


abizjak (commented):
The reason the test was not passing is that it did not run as part of CI.

Please fix that by adding `--all-features` to `cargo test` for contracts-common

```
      - name: Run cargo test
        uses: actions-rs/cargo@v1
        with:
          command: test
          args: --workspace --all-features
```

limemloh (commented):
Looks good, but don't forget to update the changelog

smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased changes
+- Fixed `Display` trait on `VersionedModuleSchema` when module contained multiple contracts to render all of them.
+- Add `--all-features` to contracts common CI test execution.
```
limemloh, line 5: This last line should not be in the changelog, only relevant information for the users

limemloh (approved):
Looks good to me when my last comment is addressed :+1:

smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased changes
+- Fixed `Display` trait on `VersionedModuleSchema` when module contained multiple contracts to render all of them.
```
abizjak, line 4: This is a nitpick, but the style in the changelog is to use present tense, not past, so "Fix ...".

This is also the style we try to use in commit messages.

PR #454, opened 2023-10-05
Dry run

td202:
## Purpose

Support for dry-running transactions and contract operations.

## Changes

- Supporting times and conversions for the dry run endpoint.

## Checklist

      hard-to-understand areas.


haskell-src/Concordium/Types/Queries.hs
```diff
@@ -1062,3 +1063,51 @@ data WinningBaker = WinningBaker
...
+    | -- | The amount to mint would overflow the total CCD supply.
+      DryRunErrorAmountOverLimit {dreMaximumMintAmount :: !Amount}
+    | -- | The balance of the sender account is not sufficient to pay for the operation.
+      DryRunErrorBalanceInsufficient {dreRequiredAmount :: !Amount, dreAvailableAmount :: !Amount}
+    | -- | The energy supplied for the transaction is not sufficient to check the transaction
+      -- header.
+      DryRunErrorEnergyInsufficient {dreEnergyRequired :: !Energy}
+
+-- | A successful result of a dry run execution.
+-- These do not cover all successful results.
```
EmilLa1, line 1086: What do you mean by this comment?

EmilLa1 (approved):

haskell-src/Concordium/Types/Queries.hs
td202: There are some other successful results that are not covered by this type. The reason for this is that those cases have a (potentially) fallible Protobuf encoding, whereas these ones will always succeed.

haskell-src/Concordium/Types/Execution.hs
```diff
@@ -1894,18 +1895,39 @@ data TransactionSummary' a = TransactionSummary
...
+    transactionSuccess :: [Event] -> a
+
+    -- | Construct a rejecting result.
+    transactionReject :: RejectReason -> a
+
+    -- | Attach a return value to the transaction.
+    setTransactionReturnValue :: BS.ByteString -> a -> a
+
+    -- | Construct a result from a 'ValidResult'.
+    fromValidResult :: ValidResult -> a
```
abizjak, line 1917: `setTransactionReturnValue` seems like an error-prone operation. Especially since this is implemented for ValidResult where that operation does nothing.

Similarly it seems error-prone to have `fromValidResult` since it defaults a value which is usually a source of problems.

Maybe there's no better solution, but this is not a nice abstraction.

limemloh (approved):
Nothing to add, which have not already been brought up.
Seems fine to me when the comments have been addressed.

haskell-src/Concordium/Types/Queries.hs
```diff
@@ -1062,3 +1063,51 @@ data WinningBaker = WinningBaker
...
+          drsProtocolVersion :: !ProtocolVersion
+        }
+    | -- | The account info was successfully retrieved.
+      DryRunSuccessAccountInfo {drsAccountInfo :: !AccountInfo}
+    | -- | The smart contract instance info was successfully retrieved.
+      DryRunSuccessInstanceInfo {drsInstanceInfo :: !Wasm.InstanceInfo}
+    | -- | The current timestamp was successfully set.
+      DryRunSuccessTimestampSet
+    | -- | The requested amount was minted to the account.
+      DryRunSuccessMintedToAccount
```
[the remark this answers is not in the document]
td202: I don't think this would be a significant improvement.
- It is redundant, and there's not any real reason that the client should forget what it just asked for.
- Should the client be expected to deal with the situation where the response does not match the request?
- It opens the gate for parroting more information back to the client (e.g. for `BlockNotFound`, which block was not found), and there does not seem to be a clear line between when that would be useful and when it would not.

abizjak (approved):

PR #456, opened 2023-10-06
Fix policy serialization and mark methods as `const`

Bargsteen:
## Purpose

Fix a bug in the serialization of the policies. The bug caused issues when using policies in your tests with the `concordium-smart-contract-testing` library.
Also marks some methods as const so they can be used to make constants.

## Changes

- Serialize the number of policies using two bytes rather than one, which matches what the node does.
- Mark `Timestamp::from_timestamp_millis` and `Timestamp::timestamp_millis` as const

## Checklist

      hard-to-understand areas.


abizjak (approved):
Good catch.

PR #460, opened 2023-10-17
Add helpers to extract build information from Wasm modules.

abizjak:
## Purpose

For the reproducible builds work we need to embed build information into the Wasm modules. This PR adds helpers for extract this information.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/src/hashes.rs
+++ b/rust-src/concordium_base/src/hashes.rs
@@ -27,11 +27,6 @@ pub enum UpdateSignMarker {}
 pub enum StateMarker {}
 
 #[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
-/// Used as a phantom type to indicate a hash is a hash with no specific
-/// meaning.
-pub enum PureHashMarker {}
-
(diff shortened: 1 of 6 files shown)


rust-src/concordium_base/src/hashes.rs
```diff
@@ -26,11 +26,6 @@ pub enum UpdateSignMarker {}
 /// Used as a phantom type to indicate a hash is a block state hash.
 pub enum StateMarker {}
```
limemloh, line 28: Since these are public, either reexport them from `concordium_contract_common` here or update the changelog of `concordium-base` with these being moved

limemloh (approved):

rust-src/concordium_base/src/hashes.rs
abizjak: I think they are re-exported already.
```
pub use concordium_contracts_common::hashes::*;
```

abizjak: So there is no change in base if I'm not mistaken.

Bargsteen (approved):
LGTM 👍

PR #464, opened 2023-10-20
Add helpers to WasmModule.

abizjak:
## Purpose

Helper functions for reading modules.

It is often necessary to read a versioned module from a file, both in deploy scripts and in tools we have. This simplifies the repetitive task.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/src/smart_contracts.rs
+++ b/rust-src/concordium_base/src/smart_contracts.rs
@@ -6,6 +6,7 @@ use crate::{
     },
     constants::*,
 };
+use anyhow::Context;
 /// Re-export of common helper functionality for smart contract, such as types
 /// and serialization specific for smart contracts.
 pub use concordium_contracts_common::{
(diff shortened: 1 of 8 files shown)


rust-src/concordium_base/src/smart_contracts.rs
```diff
@@ -143,6 +144,25 @@ impl WasmModule {
         self.serial(&mut hasher);
         ModuleReference::from(<[u8; 32]>::from(hasher.finalize()))
     }
+
+    /// Attempt to read a [`WasmModule`] from a file.
+    pub fn from_file(path: &std::path::Path) -> anyhow::Result<Self> {
```
limemloh, line 149: Since this will be in the Rust SDK, could we avoid exposing `anyhow`? Don't remember how much we are doing this already.

limemloh (approved):

rust-src/concordium_base/src/smart_contracts.rs
abizjak: I was thinking about that, but then decided not to do that since ParseResult = anyhow::Result already and we have IOErrors.

But I guess in some cases it might be useful to distinguish "read file error" vs. "parse error" even if the latter is just anyhow::Error for now.

PR #468, opened 2023-10-31
Fix `Display` on `Timestamp`

schwartz-concordium:
## Purpose

Currently deserialization of an `Type` in `to_json` fails if a Timestamp is too far in future.  This is because it calls `to_string` which used implementation of `Display`. The possible error in this implementation is never catched which makes the program panic.

## Changes
- Bugfix `Display` implementation on `Timestamp` such that an error isn’t returned. If parsing to RFC3339 fails then a u64 as string is returned.
- Updated `FromStr` on `Timestamp` to be able to parse u64 string. 

## Checklist

      hard-to-understand areas.

### DIFF
--- a/smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
+++ b/smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased changes
 
-- Fix `Display` implementation on `Timestamp` such that errors isn't returned but instead it default to a `u64` string. Also updated `FromStr` on `Timestamp` such that it can be mapped from a `u64`.
+- Fix `Display` implementation on `Timestamp` such that errors isn't returned but instead it defaults to an `u64` string. Also updated `FromStr` on `Timestamp` such that it can be mapped from an `u64`.
 
 ## concordium-contracts-common 8.1.0 (2023-10-18)
 
--- a/smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
+++ b/smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased changes
 
-- Add `TryFrom<Timestamp> for String` such that possible errors can be matched.
+- Fix `Display` implementation on `Timestamp` such that errors isn't returned but instead it default to a `u64` string. Also updated `FromStr` on `Timestamp` such that it can be mapped from a `u64`.
 
 ## concordium-contracts-common 8.1.0 (2023-10-18)
 
--- a/smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
+++ b/smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+- Add `TryFrom<Timestamp> for String` such that possible errors can be matched.
+
 ## concordium-contracts-common 8.1.0 (2023-10-18)
 
 - Add contract event schema getter on `VersionedModuleSchema`.
--- a/smart-contracts/contracts-common/concordium-contracts-common/src/types.rs
+++ b/smart-contracts/contracts-common/concordium-contracts-common/src/types.rs
@@ -778,8 +778,8 @@ impl fmt::Display for ParseTimestampError {
 impl std::error::Error for ParseTimestampError {}
 
 #[cfg(feature = "derive-serde")]
-/// The FromStr parses a string representing either an [`u64`] of milliseconds or
-/// the time according to RFC3339.
+/// The FromStr parses a string representing either an [`u64`] of milliseconds
+/// or the time according to RFC3339.
 impl str::FromStr for Timestamp {
     type Err = ParseTimestampError;
 
--- a/smart-contracts/contracts-common/concordium-contracts-common/src/types.rs
+++ b/smart-contracts/contracts-common/concordium-contracts-common/src/types.rs
@@ -747,7 +747,7 @@ impl TryFrom<Timestamp> for String {
     type Error = TimestampOverflow;
 
     fn try_from(value: Timestamp) -> Result<Self, Self::Error> {
-        if let Ok(utc)=  <chrono::DateTime<chrono::Utc>>::try_from(value) {
+        if let Ok(utc) =  <chrono::DateTime<chrono::Utc>>::try_from(value) {
             Ok(utc.to_rfc3339())
         } else {
             Err(TimestampOverflow)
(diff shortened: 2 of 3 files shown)


smart-contracts/contracts-common/concordium-contracts-common/src/types.rs
```diff
@@ -798,10 +811,9 @@ impl str::FromStr for Timestamp {
 /// format in the UTC time zone.
 impl fmt::Display for Timestamp {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        use chrono::offset::TimeZone;
-        let time = self.timestamp_millis() as i64;
-        let date = chrono::Utc.timestamp_millis_opt(time).single().ok_or(fmt::Error)?;
-        write!(f, "{}", date.to_rfc3339())
+        let date: String = String::try_from(*self)
+            .map_err(|_| fmt::Error)?;
+        write!(f, "{}", date)
```
abizjak: I know this was incorrect before, but it is still incorrect since Display implementations are not meant to fail for "spurious" reasons
```
Additionally, the return value of this function is [fmt::Result](https://doc.rust-lang.org/std/fmt/type.Result.html) which is a type alias of [Result](https://doc.rust-lang.org/std/result/enum.Result.html)<(), [std::fmt::Error](https://doc.rust-lang.org/std/fmt/struct.Error.html)>. Formatting implementations should ensure that they propagate errors from the [Formatter](https://doc.rust-lang.org/std/fmt/struct.Formatter.html) (e.g., when calling [write!](https://doc.rust-lang.org/std/macro.write.html)). However, they should never return errors spuriously. That is, a formatting implementation must and may only return an error if the passed-in [Formatter](https://doc.rust-lang.org/std/fmt/struct.Formatter.html) returns an error. This is because, contrary to what the function signature might suggest, string formatting is an infallible operation. This function only returns a result because writing to the underlying stream might fail and it must provide a way to propagate the fact that an error has occurred back up the stack.
```

limemloh: Right so I can see three options: 
- Find library (or implement) conversion with to RFC3339 from `u64`.
- Deprecate/remove this function in favor of `String::try_from`.
- Instead of the error, return a string with something like `"Unsupported date representation"`. (Don't like this, but included it anyway)

abizjak: We discussed this in the office. 

Since we assume that JSON serialization never fails, and it would be annoying not to have a Display for timestamp, the solution we'll go with is that the Display implementation (and JSON implementation) will try to render it as a RFC3393 string, and if that fails fall back to just showing the timestamp.

The FromStr implementation will be updated to support both format as well.

abizjak: I also recommend not having the `TryFrom` for String implementation. It is too ambiguous.

We should instead add a method `to_rfc_3393` that can fail.

schwartz-concordium: I have updated it now.

I didn’t add `to_rfc_3393`. Looking at it, and thinking of it as a user, I would either was always like to end up with a string so IMHO `Display` is enough.

smart-contracts/contracts-common/concordium-contracts-common/src/types.rs
```diff
@@ -778,11 +778,19 @@ impl fmt::Display for ParseTimestampError {
 impl std::error::Error for ParseTimestampError {}
 
 #[cfg(feature = "derive-serde")]
-/// The FromStr parses the time according to RFC3339.
+/// The FromStr parses a string representing either an [`u64`] or
```
limemloh, line 781:
```suggestion
/// The FromStr parses a string representing either an [`u64`] of milliseconds or
```

smart-contracts/contracts-common/concordium-contracts-common/src/types.rs
```diff
@@ -794,14 +802,19 @@ impl str::FromStr for Timestamp {
 }
 
 #[cfg(feature = "derive-serde")]
-/// The display implementation displays the timestamp according to RFC3339
-/// format in the UTC time zone.
+/// The display implementation tries to display the timestamp according to
+/// RFC3339 format in the UTC time zone.
+/// If parsing to a [`chrono::DateTime<Utc>`] fails [`Timestamp`] milliseconds,
+/// [`u64`], is returned directly as a string.
```
limemloh, line 808: I don't understand this sentence, something seems off about it.

smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased changes
 
+- Fix `Display` implementation on `Timestamp` such that errors isn't returned but instead it defaults to an `u64` string. Also updated `FromStr` on `Timestamp` such that it can be mapped from an `u64`.
```
limemloh, line 5:
```suggestion
- Fix `Display` implementation on `Timestamp` such that errors isn't returned but instead it falls back to an `u64` string.
- Change `FromStr` for `Timestamp` such that it can also be mapped from a string with an unsigned integer (`u64`).
```

limemloh (approved):
LGTM

smart-contracts/contracts-common/concordium-contracts-common/src/types.rs
schwartz-concordium: Better now?

smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased changes
 
+- Fix `Display` implementation on `Timestamp` such that errors isn't returned but instead it falls back to an `u64` string.
```
abizjak, line 5:
```suggestion
- Fix `Display` implementation on `Timestamp` such that it does not return an error if the timestamp is not representable in RFC3339 format, but instead it falls back to formatting as the number of milliseconds since the unix epoch.
```

smart-contracts/contracts-common/concordium-contracts-common/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
...
+- Change `FromStr` for `Timestamp` such that it can also be mapped from a string with an unsigned integer (`u64`).
```
abizjak, line 6:
```suggestion
- Change `FromStr` for `Timestamp` such that it also supports parsing from a number of milliseconds since the unix epoch.
```

smart-contracts/contracts-common/concordium-contracts-common/src/types.rs
```diff
@@ -778,11 +778,19 @@ impl fmt::Display for ParseTimestampError {
...
 #[cfg(feature = "derive-serde")]
-/// The FromStr parses the time according to RFC3339.
+/// The FromStr parses a string representing either an [`u64`] of milliseconds
+/// or the time according to RFC3339.
 impl str::FromStr for Timestamp {
     type Err = ParseTimestampError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let try_parse_u64 =
+            s.parse::<u64>().map(|milliseconds| Timestamp::from_timestamp_millis(milliseconds));
```
abizjak, line 788:
```suggestion
            s.parse::<u64>().map(Timestamp::from_timestamp_millis);
```

abizjak (approved):

PR #470, opened 2023-11-07
Add sign_message function to AccountKeys

DOBEN:
## Purpose

I am in the process of adding new test cases to the sponsored transaction smart contract. These new test cases generate  `AccountKeys` and `Signatures` within the test cases. To avoid having to convert signature map types, this PR adds a function `sign_message` that returns an `AccountSignatures` type. This simplifies the test cases as follows:

```
let account_signatures: AccountSignatures = keypairs
        .sign_message(&to_bytes(&message_hash));
```               
 
## Changes

- Add `sign_message` function to `AccountKeys`

### DIFF
--- a/rust-src/concordium_base/CHANGELOG.md
+++ b/rust-src/concordium_base/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased changes
 
+## 3.1.2
+
+- Add `sign_message` function to sign a message with all `AccountKeys`. The return type is `AccountSignatures`.
+
 ## 3.1.1 (2023-10-27)
 
 - Add helpers `from_file` and `from_slice` to construct a `WasmModule`.
--- a/rust-src/concordium_base/CHANGELOG.md
+++ b/rust-src/concordium_base/CHANGELOG.md
@@ -1,7 +1,5 @@
 ## Unreleased changes
 
-## 3.1.2
-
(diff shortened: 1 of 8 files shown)


rust-src/concordium_base/src/id/types.rs
```diff
@@ -2100,6 +2102,34 @@ impl AccountKeys {
         }
         signatures
     }
+
+    /// Sign the provided data with all of the keys in [`AccountKeys`].
+    /// The thresholds are ignored. The return type is [`AccountSignatures`].
```
abizjak, line 2107: I don't think it's necessary to say that the return type is that. It's clear from the type signature :)

rust-src/concordium_base/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## Unreleased changes
 
+## 3.1.2
```
abizjak, line 3: Can you please keep the Unreleased changes and not change the version number in this PR.

We do this when releasing.

abizjak (approved):

rust-src/concordium_base/src/id/types.rs
DOBEN: I added it explicitly because the `sign_data()` function has the same description without it. the last sentence gives a hint what the difference between `sign_data()` and `sign_message()` function is.

PR #471, opened 2023-11-07
Bump GRPC API to remove V1 API.

abizjak:
## Purpose

Remove V1 API.
- bump submodule
- remove blocksummary type since it is not used either in the node or in the client.

No other functionality is removed from the base libraries. While some JSON serialization might in principle be removable, most of it is still used in the client and might also come in handy for testing, so is kept.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Types/Queries.hs
+++ b/haskell-src/Concordium/Types/Queries.hs
@@ -14,10 +14,8 @@ module Concordium.Types.Queries where
 
 import Data.Aeson
 import Data.Aeson.TH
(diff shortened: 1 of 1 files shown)


td202:
The `BlockSummary` type is no longer needed, and should be removed.

abizjak:
> The `BlockSummary` type is no longer needed, and should be removed.

Yes, I already removed it.

EmilLa1 (approved):

td202 (approved):

PR #474, opened 2023-11-13
Add protocol version 7 and genesis data

limemloh:
## Purpose

Add protocol version 7 boilerplate, at this point it should behave the same as P6.

Part of https://github.com/Concordium/concordium-node/issues/1066.
Should be merged after https://github.com/Concordium/concordium-grpc-api/pull/49.
See also https://github.com/Concordium/concordium-node/pull/1075.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Constants.hs
+++ b/haskell-src/Concordium/Constants.hs
@@ -19,7 +19,10 @@ maxPayloadSize :: SProtocolVersion pv -> Word32
 maxPayloadSize SP1 = 100 * 1024 -- 100kB
 maxPayloadSize SP2 = 100 * 1024 -- 100kB
 maxPayloadSize SP3 = 100 * 1024 -- 100kB
-maxPayloadSize _ = maxWasmModuleSizeV1 + 1 + 4 + 4 -- +1 for the payload tag, +4 for the length, +4 for the module version
+maxPayloadSize SP4 = maxWasmModuleSizeV1 + 1 + 4 + 4 -- +1 for the payload tag, +4 for the length, +4 for the module version
+maxPayloadSize SP5 = maxPayloadSize SP4
+maxPayloadSize SP6 = maxPayloadSize SP4
+maxPayloadSize SP7 = maxPayloadSize SP4
 
 -- * Web assembly related constants
 
@@ -30,7 +33,9 @@ maxParameterLen SP1 = 1024
 maxParameterLen SP2 = 1024
 maxParameterLen SP3 = 1024
 maxParameterLen SP4 = 1024
-maxParameterLen _ = 65535
+maxParameterLen SP5 = 65535
+maxParameterLen SP6 = maxParameterLen SP5
+maxParameterLen SP7 = maxParameterLen SP5
 
 -- | Whether the number of logs and size of return values should be limited.
 --  The limits have been removed in P5 and onward.
@@ -39,7 +44,9 @@ limitLogsAndReturnValues SP1 = True
 limitLogsAndReturnValues SP2 = True
 limitLogsAndReturnValues SP3 = True
 limitLogsAndReturnValues SP4 = True
-limitLogsAndReturnValues _ = False
+limitLogsAndReturnValues SP5 = False
+limitLogsAndReturnValues SP6 = False
+limitLogsAndReturnValues SP7 = False
 
 -- | Maximum module size of a V0 module.
 maxWasmModuleSizeV0 :: Word32
--- a/haskell-src/Concordium/Constants.hs
+++ b/haskell-src/Concordium/Constants.hs
@@ -19,9 +19,7 @@ maxPayloadSize :: SProtocolVersion pv -> Word32
 maxPayloadSize SP1 = 100 * 1024 -- 100kB
 maxPayloadSize SP2 = 100 * 1024 -- 100kB
 maxPayloadSize SP3 = 100 * 1024 -- 100kB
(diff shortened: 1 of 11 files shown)


haskell-src/Concordium/Constants.hs
```diff
@@ -19,9 +19,7 @@ maxPayloadSize :: SProtocolVersion pv -> Word32
 maxPayloadSize SP1 = 100 * 1024 -- 100kB
 maxPayloadSize SP2 = 100 * 1024 -- 100kB
 maxPayloadSize SP3 = 100 * 1024 -- 100kB
-maxPayloadSize SP4 = maxWasmModuleSizeV1 + 1 + 4 + 4 -- +1 for the payload tag, +4 for the length, +4 for the module version
-maxPayloadSize SP5 = maxPayloadSize SP4
-maxPayloadSize SP6 = maxPayloadSize SP4
+maxPayloadSize _ = maxWasmModuleSizeV1 + 1 + 4 + 4 -- +1 for the payload tag, +4 for the length, +4 for the module version
```
abizjak, line 22: I don't really like this. I prefer that it is explicit and warned when new protocols are defined. So we don't miss something like this when we need to change it.

haskell-src/Concordium/Constants.hs
```diff
@@ -32,8 +30,7 @@ maxParameterLen SP1 = 1024
 maxParameterLen SP2 = 1024
 maxParameterLen SP3 = 1024
 maxParameterLen SP4 = 1024
-maxParameterLen SP5 = 65535
-maxParameterLen SP6 = 65535
+maxParameterLen _ = 65535
```
abizjak, line 33: Same here.

haskell-src/Concordium/Constants.hs
```diff
@@ -42,8 +39,7 @@ limitLogsAndReturnValues SP1 = True
 limitLogsAndReturnValues SP2 = True
 limitLogsAndReturnValues SP3 = True
 limitLogsAndReturnValues SP4 = True
-limitLogsAndReturnValues SP5 = False
-limitLogsAndReturnValues SP6 = False
+limitLogsAndReturnValues _ = False
```
abizjak, line 42: And here. I don't think wildcards are a good idea.

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -471,9 +479,7 @@ supportsV1Contracts :: SProtocolVersion pv -> Bool
 supportsV1Contracts SP1 = False
 supportsV1Contracts SP2 = False
 supportsV1Contracts SP3 = False
-supportsV1Contracts SP4 = True
-supportsV1Contracts SP5 = True
-supportsV1Contracts SP6 = True
+supportsV1Contracts _ = True
```
abizjak, line 482: Please don't use wildcard here. I think these configurations should be very explicit to avoid accidents.

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -542,4 +546,4 @@ supportsAccountSignatureChecks spv = case spv of
     SP3 -> False
     SP4 -> False
     SP5 -> False
-    SP6 -> True
+    _ -> True
```
abizjak, line 549: Same in all of the above.

abizjak (approved):

haskell-src/Concordium/Genesis/Data/P7.hs
```diff
@@ -0,0 +1,144 @@
...
+-- | Serialize genesis data with a version tag.
+--  This will use the V9 format.
+putVersionedGenesisData :: Putter GenesisDataP7
+putVersionedGenesisData gd = do
+    putVersion 9
+    putGenesisDataV9 gd
+
+-- | Compute the block hash of the genesis block with the given genesis data.
+--  Every block hash is derived from a message that begins with the block slot,
+--  which is 0 for genesis blocks.
```
td202, line 91: We probably should have fixed this in P6 - it's too late now, but we can fix it in P7. We should make sure the hashing schemes for (re)genesis blocks are consistent with the general hashing scheme. In particular, we shouldn't be using slot numbers. This could wait until we revise the hashing scheme, but we should at least make a note to do it.

td202 (approved):

haskell-src/Concordium/Genesis/Data/P7.hs
limemloh: I have added this issue for it https://github.com/Concordium/concordium-node/issues/1083

EmilLa1 (approved):

haskell-src/Concordium/Genesis/Data/P7.hs
```diff
@@ -0,0 +1,144 @@
...
+import qualified Concordium.Genesis.Data.Base as Base
+import qualified Concordium.Genesis.Data.BaseV1 as BaseV1
+import Concordium.Types
+
+-- | Initial genesis data for the P7 protocol version.
+data GenesisDataP7 = GDP7Initial
+    { -- | The immutable genesis parameters.
+      genesisCore :: !BaseV1.CoreGenesisParametersV1,
+      -- | Serialized initial block state.
+      --  NB: This block state contains some of the same values as 'genesisCore', and they should match.
```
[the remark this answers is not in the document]
limemloh: It seems that there are no overlapping values.
I'll remove the comment

PR #476, opened 2023-11-19
Improve the multiscalar multiplication precomputation step.

abizjak:
## Purpose

Avoid field operations since they can be slow.

While it does not appear to have a significant effect for the BLS curve, but still has non-trivial effect, it might matter more with other curves.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/src/curve_arithmetic/mod.rs
+++ b/rust-src/concordium_base/src/curve_arithmetic/mod.rs
@@ -215,10 +215,10 @@ pub fn multiexp_worker_given_table<C: Curve>(
     window: usize,
 ) -> C {
     // Compute the wnaf
-    let window_size = window + 1;
+    let window_plus1 = window + 1;
     let k = exps.len();
-    // assert_eq!(gs.len(), k);
-    assert!(window_size < 62);
+    assert!(window_plus1 >= 2);
+    assert!(window_plus1 < 63);
 
     let mut wnaf = Vec::with_capacity(k);
 
@@ -228,7 +228,7 @@ pub fn multiexp_worker_given_table<C: Curve>(
     // this avoids any field arithmetic and operates directly on the bit
     // representation of the scalars, leading to a substantial performance
     // improvement.
-    let width = 1u64 << window_size;
+    let width = 1u64 << window_plus1;
     let window_mask = width - 1;
 
     for c in exps.iter() {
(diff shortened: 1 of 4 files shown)


rust-src/concordium_base/src/curve_arithmetic/mod.rs
```diff
@@ -212,46 +212,70 @@ pub fn multiexp_worker<C: Curve, X: Borrow<C>>(
 pub fn multiexp_worker_given_table<C: Curve>(
     exps: &[C::Scalar],
     table: &[Vec<C>],
-    window_size: usize,
+    window: usize,
 ) -> C {
     // Compute the wnaf
-
+    let window_size = window + 1;
```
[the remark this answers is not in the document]
hamiidreza: The code here makes sense to me. What we implement here is the NAF method with width "window+1" which has a different width parameter compared to dalek NAF with width "window". Calling it "window_size" looks a bit bit confusing and inconsistent with the rest of the code, though! I would call it something that reflects that this is 1 plus the window parameter, e.g., "window_plus1".

rust-src/concordium_base/src/curve_arithmetic/mod.rs
```diff
@@ -212,46 +212,70 @@ pub fn multiexp_worker<C: Curve, X: Borrow<C>>(
...
     let k = exps.len();
     // assert_eq!(gs.len(), k);
-    assert!(window_size >= 1);
```
hamiidreza, line 221: Why is this commented? Should it not be `assert!(window_size >= 2);` now?

rust-src/concordium_base/src/curve_arithmetic/mod.rs
```diff
@@ -212,46 +212,70 @@ pub fn multiexp_worker<C: Curve, X: Borrow<C>>(
...
     assert!(window_size < 62);
```
hamiidreza, line 221: I think for consistency with the rest of the code, this should also be `assert!(window_size < 63);`.

abizjak: Ok.

rust-src/concordium_base/src/curve_arithmetic/mod.rs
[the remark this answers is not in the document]
abizjak: Ok.

rust-src/concordium_base/src/curve_arithmetic/mod.rs
abizjak: I reinstated it.

PR #477, opened 2023-11-20
Re-export ed25519_dalek crate

DOBEN:
## Purpose

It would be nice to re-export the type `ed25519_dalek::PublicKey` in the smart contract integration testing library so that the `ed25519_dalek` crate does not have to be added as a dev-dependency in smart contract projects.

## Changes

Re-export `ed25519_dalek` crate.

### DIFF
--- a/rust-src/concordium_base/src/lib.rs
+++ b/rust-src/concordium_base/src/lib.rs
@@ -32,6 +32,8 @@ pub mod ps_sig;
 
 pub mod dodis_yampolskiy_prf;
 
+pub use ed25519_dalek::{PublicKey, SecretKey, Signature};
+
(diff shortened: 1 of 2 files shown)


rust-src/concordium_base/src/lib.rs
```diff
@@ -32,6 +32,8 @@ pub mod ps_sig;
 
 pub mod dodis_yampolskiy_prf;
 
+pub use ed25519_dalek;
```
abizjak, line 35: I don't think we want to re-export the entire crate. That's too much.

We should only export the public and secret keys and signatures since they appear in our own API.

abizjak (approved):

PR #478, opened 2023-11-20
Add singleton account access structure

DOBEN:
## Purpose

- Add `singleton` function to `accountAccessSructure` that generates the most basic `account access structure` with one public key at index 0 in both maps.
- Add `new` function to `accountAccessStructure` that given the indices, public_keys, and thresholds generates corresponding  `account access structure`.  

## Changes
-  Add `singleton` function
- Add `new` function

Note: The PR contains the two line changes from https://github.com/Concordium/concordium-base/pull/477.

### DIFF
--- a/rust-src/concordium_base/CHANGELOG.md
+++ b/rust-src/concordium_base/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Unreleased changes
 
+- Add `singleton` and `generate` function to `AccountAccessStructure`.
 - Export `PublicKey`, `SecretKey`, and `Signature` type from `ed25519_dalek` crate.
 - Add `sign_message` function to sign a message with all `AccountKeys`. The return type is `AccountSignatures`.
 
--- a/rust-src/concordium_base/CHANGELOG.md
+++ b/rust-src/concordium_base/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Unreleased changes
 
-- Add `singleton` and `generate` function to `AccountAccessStructure`.
+- Add `singleton` and `new` function to `AccountAccessStructure`.
 - Export `PublicKey`, `SecretKey`, and `Signature` type from `ed25519_dalek` crate.
 - Add `sign_message` function to sign a message with all `AccountKeys`. The return type is `AccountSignatures`.
 
--- a/rust-src/concordium_base/src/lib.rs
+++ b/rust-src/concordium_base/src/lib.rs
@@ -32,7 +32,11 @@ pub mod ps_sig;
 
 pub mod dodis_yampolskiy_prf;
 
-pub use ed25519_dalek::{PublicKey, SecretKey, Signature};
+/// We expose the `PublicKey`, `SecretKey`, and `Signature` from the third-party
+/// `ed25519_dalek` crate here because these types appear in Concordium's API.
+pub mod ed25519 {
+    pub use ed25519_dalek::{PublicKey, SecretKey, Signature};
+}
 
 #[cfg(feature = "ffi")]
 mod ffi_helpers;
--- a/rust-src/concordium_base/src/transactions.rs
+++ b/rust-src/concordium_base/src/transactions.rs
@@ -24,6 +24,7 @@ use crate::{
     smart_contracts, updates,
 };
 use concordium_contracts_common as concordium_std;
+use concordium_std::SignatureThreshold;
 use derive_more::*;
 use rand::{CryptoRng, Rng};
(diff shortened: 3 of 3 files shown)


rust-src/concordium_base/src/transactions.rs
```diff
@@ -1464,6 +1465,52 @@ pub struct AccountAccessStructure {
...
+    CredentialIndex,
+    SignatureThreshold,
+    &'a [(KeyIndex, ed25519_dalek::PublicKey)],
+)];
+
+impl AccountAccessStructure {
+    /// Generate a new [`AccountAccessStructure`] for the thresholds, public
+    /// keys, and key indices specified in the input. If there are duplicate
+    /// indices then later ones override the previous ones.
+    pub fn generate(account_threshold: AccountThreshold, structure: AccountStructure) -> Self {
```
abizjak, line 1478: I think it'd be better to call this `new` since we usually use `generate` for values that are somehow generated from either randomness or seed.

rust-src/concordium_base/src/transactions.rs
```diff
@@ -1464,6 +1465,52 @@ pub struct AccountAccessStructure {
...
+        }
+
+        AccountAccessStructure {
+            keys:      map,
+            threshold: account_threshold,
+        }
+    }
+
+    /// Generate a new [`AccountAccessStructure`] with a single credential and
+    /// public key, at credential and key indices 0.
```
abizjak, line 1504:
```suggestion
    /// public key, at credential and key index 0.
```

abizjak (approved):

rust-src/concordium_base/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased changes
 
+- Add `singleton` and `generate` function to `AccountAccessStructure`.
```
abizjak, line 3: Make sure to also update the name here.

abizjak (approved):

rust-src/concordium_base/src/transactions.rs
```diff
@@ -1464,6 +1465,52 @@ pub struct AccountAccessStructure {
     pub threshold: AccountThreshold,
 }
 
+type AccountStructure<'a> = &'a [(
```
limemloh, line 1468: I think this should be `pub` and have a doc comment

rust-src/concordium_base/src/lib.rs
```diff
@@ -32,6 +32,8 @@ pub mod ps_sig;
 
 pub mod dodis_yampolskiy_prf;
 
+pub use ed25519_dalek::{PublicKey, SecretKey, Signature};
```
limemloh, line 35: I don't think we should have these exported at the top level of the crate, since we will now export both a `PublicKey` and a `VerifyKey` where the type `VerifyKey` is what we want people to use.
Also, `SignatureEd25519` and this `Signature` are sort of 'overlapping' in meaning.

However, it does make sense to expose these as they are a peer dependency, so maybe do:
```suggestion
pub use ed25519_dalek;
```

limemloh (approved):

rust-src/concordium_base/src/lib.rs
DOBEN: There was a discussion in https://github.com/Concordium/concordium-base/pull/477 to export these three types.

rust-src/concordium_base/src/transactions.rs
DOBEN: I feel it should be plural because we refer to both indices `credential index 0` and the `key index 0`.

rust-src/concordium_base/src/lib.rs
abizjak: I don't want to expose the entire dalek crate here. Exposing third party crates complicates our updates and versioning of our own crate. I prefer to expose types that appear in our API but not everything else.

But it's a good point that exposing these at the top-level under these names is not ideal, so we should rename them.

Perhaps
```rust
pub mod ed25519 {
          pub use ed25519_dalek::{PublicKey, SecretKey, Signature};
}
```
with some documentation would be most understandable.

DOBEN: I added an explanation and changed it:

```
/// We expose the `PublicKey`, `SecretKey`, and `Signature` from the third-party `ed25519_dalek` crate here because these types appear in Concordium's API.
pub mod ed25519 {
    pub use ed25519_dalek::{PublicKey, SecretKey, Signature};
}
```

PR #479, opened 2023-11-21
Add from trait accountKeys to accountPublicKeys

DOBEN:
## Purpose

Add from trait `AccountKeys` to `AccountPublicKeys`. This will simplify the cis3 test case since the generated `keyPairs` can be converted into `AccountPublicKeys`  as follows:

```
    let account_public_keys: AccountPublicKeys = (keypairs).into();
```

## Changes
Add from trait `AccountKeys` to `AccountPublicKeys`.

### DIFF
--- a/smart-contracts/contracts-common/concordium-contracts-common/src/types.rs
+++ b/smart-contracts/contracts-common/concordium-contracts-common/src/types.rs
@@ -457,6 +457,11 @@ pub struct ZeroSignatureThreshold;
 #[repr(transparent)]
 pub struct PublicKeyEd25519(pub [u8; 32]);
 
+impl PublicKeyEd25519 {
+    /// Create a new [`PublicKeyEd25519`] from bytes.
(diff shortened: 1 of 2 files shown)


limemloh (approved):
Don't forget to update the changelogs

smart-contracts/contracts-common/concordium-contracts-common/src/types.rs
```diff
@@ -457,6 +457,11 @@ pub struct ZeroSignatureThreshold;
 #[repr(transparent)]
 pub struct PublicKeyEd25519(pub [u8; 32]);
 
+impl PublicKeyEd25519 {
+    /// Create a new [`PublicKeyEd25519`] from bytes.
+    pub fn from_bytes(bytes: [u8; 32]) -> Self { PublicKeyEd25519(bytes) }
+}
```
abizjak, line 463: Why do we need this method?

I'd prefer not to have it since we have public fields already and this method does not verification compared to a similar method for the ed25519::PublicKey.

abizjak (approved):

abizjak (approved):

PR #480, opened 2023-11-21
Attempt to fix CI.

abizjak:
## Purpose

Fix actions so that the contracts-common actions actually run.

Also avoid using obsolete actions-rs actions and update checkout action to avoid warnings about obsolete versions.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/.github/workflows/build-test-smart-contracts.yaml
+++ b/.github/workflows/build-test-smart-contracts.yaml
@@ -113,6 +113,7 @@ jobs:
       - name: Clippy
         working-directory: smart-contracts/${{ matrix.build-dir }}
         run: |
+          git config --global url."https://github.com/".insteadOf "git@github.com:"
           rustup default ${{ env.RUST_CLIPPY }}
           rustup component add clippy
           cargo clippy --color=always --tests --benches -- -Dclippy::all
@@ -138,6 +139,7 @@ jobs:
       - name: Clippy
         working-directory: smart-contracts/${{ matrix.build-dir }}
         run: |
+          git config --global url."https://github.com/".insteadOf "git@github.com:"
           rustup default ${{ env.RUST_CLIPPY }}
           rustup component add clippy
           cargo clippy --locked --features ${{ matrix.features }} --no-default-features --color=always --tests --benches -- -Dclippy::all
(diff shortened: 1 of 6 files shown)


.github/workflows/build-test-smart-contracts.yaml
```diff
@@ -187,19 +161,14 @@ jobs:
...
-          target: wasm32-unknown-unknown
-          components: clippy
+        uses: actions/checkout@v4
       - name: Clippy
         working-directory: smart-contracts/rust-contracts/example-contracts/${{ matrix.example-contract }}
         run: |
           git config --global url."https://github.com/".insteadOf "git@github.com:"
+          rustup default ${{ env.RUST_CLIPPY }}
+          rustup component add clippy
+          rustup target add wasm32-unknown-unknown
```
DOBEN, line 171: Do we need the `target` here?

.github/workflows/build-test-smart-contracts.yaml
```diff
@@ -243,35 +208,25 @@ jobs:
...
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
-      - name: Install Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: ${{ env.RUST_CLIPPY }}
-          override: true
+        uses: actions/checkout@v4
```
DOBEN, line 211: There is this comment `#Remaining example contracts have not yet been added to ci`. Just wondering if we could add the remaining tests.

.github/workflows/build-test-sources.yaml
```diff
@@ -66,16 +66,10 @@ jobs:
...
-      with:
-        profile: minimal
-        toolchain: ${{ matrix.plan.rust }}
-        override: true
-        components: rustfmt
-
     - name: Run rustfmt
       run: |
+        rustup default ${{ matrix.plan.rust }}
+        rustup component add rustfmt
```
DOBEN, line 72: The lines below, e.g. `cargo fmt --manifest-path rust-bins/Cargo.toml -- --check` can be written via a matrix in my opinion.

DOBEN (approved):

.github/workflows/build-test-sources.yaml
abizjak: Yes.

.github/workflows/build-test-smart-contracts.yaml
abizjak: Yes, because some example contracts have a `.cargo/config.toml` which specifies the wasm target.

.github/workflows/build-test-smart-contracts.yaml
abizjak: Yes, but I think it's not entirely trivial so I would like to postpone this to a separate PR.

I added an issue https://github.com/Concordium/concordium-base/issues/481

PR #482, opened 2023-11-22
Fix idiss memory leak

schwartz-concordium:
## Purpose

The pointer returns to C# from FFI calls to rust are never deallocated.
https://github.com/Concordium/concordium-base/blob/d23e634de670b8917e226ed8545f56abe6ccc484/idiss-csharp/IdissLib/Idiss.cs#L98
https://github.com/Concordium/concordium-base/blob/d23e634de670b8917e226ed8545f56abe6ccc484/idiss/src/cs_exports.rs#L80

## Changes

* Add a callback which frees memory of array.
* Added missing `json` file for test to run green

## Checklist

      hard-to-understand areas.


idiss/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.6.1
+  - Add dependency to feature `ffi` from `concordium_base` such that `ffi_helpers` module can be used.
```
abizjak, line 4: It is better to be more direct by stating that the `free_array_len_cap` is now exported from the shared library.

idiss/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "idiss"
 build = "build.rs"
-version = "0.6.0"
+version = "0.6.1"
```
abizjak, line 4: There is no need to change this version. This functionality is not affected.

rust-src/concordium_base/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased changes
 
+- Add extern function `free_array_len_cap` such vector pointers can be released and deprecated `free_array_len`
+  since it didn't account for capacity size.
```
abizjak, line 4: `ffi` stuff is internal and we make no guarantees about it. It should also not go into the public changelog.

rust-src/concordium_base/src/ffi_helpers/common.rs
```diff
@@ -1,8 +1,20 @@
 #[no_mangle]
 /// Free an array that was converted to a pointer from a vector.
 /// This assumes the vector's capacity and length were the same.
+#[deprecated(
+    note = "use [`free_array_len_cap`] instead since it correctly frees the whole capacity size \
+            of the vector."
+)]
```
abizjak, line 7: Don't put the deprecation notice on this. It has no effect since this function is not to be used from Rust. None of the `ffi` module is meant to be used from Rust.

idiss/src/cs_exports.rs
```diff
@@ -355,6 +364,7 @@ pub unsafe extern "C" fn validate_recovery_request_cs(
         Err(e) => {
             let mut bytes = format!("{}", e).into_bytes();
             *out_length = bytes.len() as i32;
+            *out_capacity = bytes.capacity() as i32;
             let ptr = bytes.as_mut_ptr();
             std::mem::forget(bytes);
             ptr
```
abizjak, line 370: While this works, it would be better to use the newer ManuallyDrop since it is more clear
```
let wrapper = ManuallyDrop::new(bytes);
wrapper.as_ptr()
```

rust-src/concordium_base/src/ffi_helpers/common.rs
abizjak: Also, I think the `free_array_len` is technically correct the way we use it. Ultimately to free the pointer it only needs the pointer, the memory allocator will know what chunk of data has been allocated with that pointer.

So we don't want to make it look like the pervasive use of free_array_len in the node is incorrect.

But free_array_len_cap is simpler to understand so it's cleaner to use for the idiss case.

idiss-csharp/IdissLib/Idiss.cs
```diff
@@ -202,25 +247,46 @@ public static IdentityCreationV1 CreateIdentityObjectV1(Versioned<IpInfo> ipInfo
...
         }
+        
+        private static void FreeNonZeroPtr(IntPtr outputPtr, int outLength, int outCapacity)
+        {
+            if (outputPtr == IntPtr.Zero)
+            {
+                return;
+            }
+
+            free_array_len_cap(outputPtr, outLength, outCapacity);
```
abizjak, line 289: It's a bit odd that here `outLength` and `outCapacity` are also `int` but the method takes `u64` in Rust. Whereas other functions that are imported also use `int` in C# but in Rust take i32.

idiss/Cargo.toml
schwartz-concordium: The `cs_export.rs` has changed and that is actually a breaking change which I didn’t mention in the changelog. I have updated the changelog now and bump the major.

Please let me now if procedure is wrong.

rust-src/concordium_base/CHANGELOG.md
schwartz-concordium: removed

rust-src/concordium_base/src/ffi_helpers/common.rs
schwartz-concordium: removed

idiss-csharp/IdissLib/Idiss.cs
```diff
@@ -22,25 +23,28 @@ public static class Idiss
...
+        out int out_length, out int out_capacity, out int out_success);
 
         [DllImport("idiss.dll")]
         private static extern IntPtr validate_recovery_request_cs([MarshalAs(UnmanagedType.LPArray)] byte[] ctx, int ctx_len,
         [MarshalAs(UnmanagedType.LPArray)] byte[] ip_info, int ip_info_len,
-        [MarshalAs(UnmanagedType.LPArray)] byte[] request, int request_len, out int out_length);
+        [MarshalAs(UnmanagedType.LPArray)] byte[] request, int request_len, out int out_length, out int out_capacity);
+        
+        [DllImport("idiss.dll")]
+        private static extern IntPtr free_array_len_cap(IntPtr array_ptr,  uint out_length, uint out_capacity);
```
limemloh, line 47: The `out_length` and `out_capacity` needs to be `ulong` to [match](https://learn.microsoft.com/en-us/dotnet/standard/native-interop/type-marshalling#default-rules-for-marshalling-common-types) the rust side

```suggestion
        private static extern IntPtr free_array_len_cap(IntPtr array_ptr, ulong out_length, ulong out_capacity);
```

rust-src/concordium_base/src/ffi_helpers/common.rs
```diff
@@ -6,3 +6,11 @@ extern "C" fn free_array_len(ptr: *mut u8, len: u64) {
         Vec::from_raw_parts(ptr, len as usize, len as usize);
     }
 }
+
+#[no_mangle]
+/// Free an array that was converted to a pointer from a vector.
```
limemloh, line 11: **Request change:** I find the comment and function name a bit misleading, since the function is for freeing a `Vec<u8>` specifically, not an array and not just any `Vec<T>`.

limemloh (changes requested):

rust-src/concordium_base/src/ffi_helpers/common.rs
limemloh: At least it is not an array in Rust or C# terms, I guess it is an array in some other languages, like javascript.

idiss-csharp/IdissLib/Idiss.cs
abizjak: My point was more about the fact that int seems  to be 32-bit and the size of the value should match on both sides of the FFI boundary. Otherwise we might have problems, e.g., it could be that in some platforms a 64-bit value is passed as two 32-bit values in registers.

The signed/unsigned distinction I don't think is a problem since all platforms just use two's complement representation.

But I see you changed the type of the C# function to take a ulong so that resolves this issue.

limemloh (approved):

abizjak (approved):

PR #491, opened 2023-12-07
Validate mnemonic word length

orhoj:
## Purpose
According to the documentation for BIP39 (https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#generating-the-mnemonic) the mnemonic must encode entropy in multiples of 32 bits, and the allowed size is 128-256 bits. This corresponds to allowing for 12, 15, 18, 21 and 24 word mnemonics. The current implementation allowed for any type of input, but assumed it was always 24 words. This change updates it so that it restricts the input as described by the standard, and also adds the missing tests to cover the test vectors for 12 and 18 words. There are no "official" test vectors for 15 and 21, and therefore a single test has been added to ensure that those work.

Note that this is a breaking change to the interface, hence the major bump.

## Changes
- Restrict number of words to be in `[12, 15, 18, 21, 24]`.
- Added unit tests to cover test vectors for 12 and 18 words input.
- Added single unit tests for 15 and 21 words (no "official" test vectors available for these).

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/key_derivation/src/lib.rs
+++ b/rust-src/key_derivation/src/lib.rs
@@ -16,6 +16,7 @@ use hmac::Hmac;
 use keygen_bls::keygen_bls;
 use sha2::Sha512;
 use std::fmt;
+use thiserror::Error;
 
 #[derive(Copy, Clone, Debug, SerdeSerialize, SerdeDeserialize)]
 pub enum Net {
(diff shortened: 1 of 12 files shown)


rust-src/key_derivation/src/lib.rs
```diff
@@ -37,25 +38,42 @@ impl fmt::Display for Net {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result { write!(f, "{}", self.net_code()) }
 }
 
+#[derive(Debug, Error)]
+pub enum MnemonicLengthError {
+    #[error("Invalid mnemonic length. The length must be 12, 15, 18, 21 or 24.")]
+    InvalidLength,
+}
```
abizjak, line 45: Why is this an enum? If there's only one variant it'd be better to just have
```
#[derive(Debug, Error)]
#[error("Invalid mnemonic length. The length must be 12, 15, 18, 21 or 24.")]
pub struct MnemonicLengthError;
```

PR #495, opened 2023-12-19
Fix credential common input struct

orhoj:
## Purpose
Move `attribute_randomness` and `credential_public_keys` out of common struct for credential deployment as they should be derived from the seed phrase when creating the transaction from seed.

## Changes
- Moved `credential_public_keys` and `attribute_randomness` to the keys specific struct from the common struct.
- I have not bumped versions or updated CHANGELOG as this code is not in use yet.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/wallet_library/src/credential.rs
+++ b/rust-src/wallet_library/src/credential.rs
@@ -14,7 +14,7 @@ use concordium_base::{
 use key_derivation::Net;
 use serde::{Deserialize as SerdeDeserialize, Serialize as SerdeSerialize};
 use serde_json::json;
-use std::{collections::BTreeMap, str::FromStr};
+use std::collections::BTreeMap;
 
 use crate::wallet::get_wallet;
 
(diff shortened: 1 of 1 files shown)


rust-src/wallet_library/src/credential.rs
```diff
@@ -82,11 +83,37 @@ pub fn create_unsigned_credential_with_seed_v1_aux(
...
+    let mut attribute_randomness = BTreeMap::new();
+    for attribute_name in ATTRIBUTE_NAMES.iter() {
+        let tag = AttributeTag::from_str(attribute_name).unwrap();
+        let randomness: PedersenRandomness<ArCurve> = wallet.get_attribute_commitment_randomness(
+            identity_provider_index,
+            identity_index,
+            input.common.cred_number.into(),
+            tag,
+        )?;
+        attribute_randomness.insert(tag, randomness);
```
abizjak, line 95: I don't understand why we are generating all randomness here. In the `UnsignedCredentialInput` we have the tags in the identity object. Should we not be generating randomness only for those (or rather only for those that are not revealed)?

orhoj: I copied what we had in the JS SDK that already did this. Looking at the code further down I can see that we indeed only use the randomness for the non-revealed attributes. I will fix it.

rust-src/wallet_library/src/credential.rs
```diff
@@ -82,11 +83,42 @@ pub fn create_unsigned_credential_with_seed_v1_aux(
...
+    let key = wallet.get_account_public_key(
+        identity_provider_index,
+        identity_index,
+        input.common.cred_number.into(),
+    )?;
+    let mut key_map: BTreeMap<KeyIndex, VerifyKey> = BTreeMap::new();
+    key_map.insert(KeyIndex(0), VerifyKey::Ed25519VerifyKey(key));
+    let credential_public_keys: CredentialPublicKeys = CredentialPublicKeys {
+        threshold: NonZeroThresholdU8::ONE,
+        keys:      key_map,
```
abizjak, line 112: Maybe it would be useful to explicitly state this in the documentation, the fact that this hardcodes the choice of threshold and number of keys.

abizjak (approved):

PR #496, opened 2024-01-04
Arkworks integration and ed25519 upgrade

annenkov:
## Purpose

This PR adds generic adapters (as blanket implementation) to the ArkWorks interfaces. The adapters (with some additional configuration traits) are used to replace the BLS12-381 curve with ArkWorks implementation.  In addition to that, `ed25519-dalek` is upgraded to `v2.0`, which was previously blocked by the `pairing` crate and the corresponding BLS implementation (new versions don't expose the target group element type `Fp12`, preventing from serializing it, [see details here](https://github.com/zkcrypto/bls12_381/issues/96)). 

## Changes

- implement `ArkGroup<G>` and `ArkField<F>` wrappers;
- blanket implementations of the `Field` and `PrimeField` traits for `ArkField<F>`;
- blanket implementation of  the `Curve` trait `ArkGroup<G>`;
- all required trait implementations for the ArkWorks `BLS12-381` curve, that replaces the current BLS curve implementation from the `pairing` crate.
- upgrade `ed25519-dalek` signature scheme; the `v2.0` API is a bit different, it bundles a secret and verifying (public) key into a signing key and maintains an invariant that the verifying key is always the public key of  secret_key; signing does not require passing the verifying key, so it was removed from parameters of the corresponding functions.
- update FFI function types on the Haskell side to match the changes in the signing API (removed `verifing_key` parameter)

## Checklist

      hard-to-understand areas.
- [x] I have updated the CHANGELOG.


rust-src/concordium_base/src/curve_arithmetic/arkworks_instances.rs
```diff
@@ -0,0 +1,207 @@
...
+            let mut v = [0u8; 8];
+            v[..chunk.len()].copy_from_slice(chunk);
+            fr.push(u64::from_le_bytes(v));
+        }
+        // unset two topmost bits in the last read u64.
+        *fr.last_mut().expect("Non empty vector expected") &= !(1u64 << 63 | 1u64 << 62);
+        // fr[3] &= !(1u64 << 63 | 1u64 << 62);
+        <Self::Scalar>::from_repr(&fr)
+            .expect("The scalar with top two bits erased should be valid.")
+    }
```
abizjak, line 193: So this is problematic. This `scalar_from_bytes` only makes sense for the BLS curve, but it will not make sense in general. We need to either have this implementation more general, or fix it a specific curve.

Is there no serialization for scalars in ark? I guess there is https://docs.rs/ark-ff/latest/ark_ff/fields/trait.PrimeField.html#method.from_le_bytes_mod_order but that won't be quite the same.

I can't do it now, but we should look into how this method is actually used to see what the best thing to do is.

rust-src/concordium_base/src/curve_arithmetic/arkworks_instances.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    fn hash_to_group(m: &[u8]) -> Self {
+        let hasher = G::Hasher::new(G::DOMAIN_STRING.as_ref())
+            .expect("Expected valid domain separation string");
+        let res = G::Hasher::hash(&hasher, m).expect("Expected successful hashing to curve");
```
abizjak, line 198: It's unclear to me in which cases this fails?

Can it happen? 

It's not ideal to have all of these unwraps here since it makes the implemenation not generic.

Would it complicate a lot to propagate the failure?

rust-src/concordium_base/src/curve_arithmetic/arkworks_instances.rs
```diff
@@ -0,0 +1,207 @@
...
+        ArkGroup(res.into())
+    }
+}
+
+impl<F: FromStr> FromStr for ArkField<F> {
+    type Err = F::Err;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> { F::from_str(s).map(|x| x.into()) }
+}
```
abizjak, line 207: Can't you do `derive_more::FromStr`?

rust-src/concordium_base/src/curve_arithmetic/bls12_381_arkworks.rs
```diff
@@ -0,0 +1,293 @@
...
+        let res = Fr::from_bigint(big_int)
+            .ok_or(anyhow!("Cannot convert from bigint to a field element"))?;
+        Ok(res)
+    }
+}
+
+impl Serial for Fr {
+    fn serial<B: Buffer>(&self, out: &mut B) {
+        // Note that it is crucial to use `into_bigint()` here.
+        // The internal representation is accessible direclty, but it's NOT the same
```
abizjak, line 65:
```suggestion
        // The internal representation is accessible directly, but it's NOT the same
```

rust-src/concordium_base/src/curve_arithmetic/ed25519_instance.rs
```diff
@@ -318,7 +275,7 @@ pub(crate) mod tests {
             1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 254, 255, 255, 255, 255, 255, 255, 255,
             0, 0, 0, 0, 0, 0, 0, 0,
         ])
-        .expect("Expected a valid scalar")
+        .unwrap()
```
abizjak, line 278: This is strictly worse :) I wonder why it was changed?

rust-src/concordium_base/src/curve_arithmetic/field_adapters.rs
```diff
@@ -0,0 +1,57 @@
+//! Wrapper types and blanket implementations serving as adapters from
+//! the `ff` crate `Field`.
+use ff;
```
abizjak, line 3: Could you also explain why we need this?

Is it because curve25519 and ristretto use `ff`?

Does ark not have ristretto?

rust-src/concordium_base/src/curve_arithmetic/secret_value.rs
```diff
@@ -117,7 +120,17 @@ mod tests {
...
-
-    macro_test_value_to_byte_conversion!(value_to_byte_conversion_bls12_381_g2_affine, G2Affine);
+    // TODO: the code used to be diefined using the affine representation, but
+    // ArkWorks' BLS does not implement `CurveGroup` for the affine representation.
+    // Is it important that it's the affine representation? For now, the BLS
+    // projective representation will be used here.
+    macro_test_value_to_byte_conversion!(
+        value_to_byte_conversion_bls12_381_g1_affine,
+        ArkGroup<G1Projective>
+    );
```
abizjak, line 130: Just rename it to use projective. This is just a test, it is not important that it was for affine, since we don't have the affine groups anymore.

rust-src/concordium_base/src/ecvrf/mod.rs
```diff
@@ -262,7 +262,12 @@ mod tests {
 
             // Test serialization of generated secret scalar
             let x = expanded_sk.key;
-            assert_eq!(x, Scalar::from_bits(x_bytes));
+            // TODO: we use deprecated `from_bits` here, this also requires enabling the
+            // `legacy_compatibility` feature for the `ed25519-dalek` dependency. Maybe ther
+            // is a different way of implementing this.
+            #[allow(deprecated)]
+            let scalar = Scalar::from_bits(x_bytes);
+            assert_eq!(x, scalar);
```
abizjak, line 270: So this is just in a test, so we can only enable this legacy feature as part of a `dev-dependency`.

But instead, could we not just check `x.to_bytes() == x_bytes`?

rust-src/concordium_base/src/ecvrf/proof.rs
```diff
@@ -34,14 +34,16 @@ pub struct Proof(pub EdwardsPoint, pub Scalar, pub Scalar);
...
+        // so we just convert it to bytes.
+        let c = &self.1.to_bytes();
         // assert c is within range
         assert_eq!(c[16..32], [0u8; 16]);
         x.write_all(&self.0.compress().to_bytes()[..])
             .expect("Writing to buffer should succeed.");
         x.write_all(&c[..16])
             .expect("Writing to buffer should succeed.");
-        x.write_all(&self.2.reduce().to_bytes()[..])
+        x.write_all(&self.2.to_bytes()[..])
```
abizjak, line 46: So this is safe because to_bytes always returns canonical representatives/reduced scalars?

rust-src/concordium_base/src/ecvrf/public.rs
```diff
@@ -94,11 +94,16 @@ impl PublicKey {
...
-        bits[31] |= 0b_0100_0000;
+        // It seems like `from_bits` is the only way to construct an unreduced scalar
+        // that can potentially be grater than the field's order.
 
-        let point = &Scalar::from_bits(*bits) * &constants::ED25519_BASEPOINT_TABLE;
+        // TODO: we use deprecated `from_bits` here, this also requires enabling the
+        // `legacy_compatibility` feature for the `ed25519-dalek` dependency. Maybe ther
+        // is a different way of implementing this.
+        #[allow(deprecated)]
+        let scalar = Scalar::from_bits(clamp_integer(*bits));
```
abizjak, line 104: So doesn't `from_bits` do nothing here since https://docs.rs/curve25519-dalek/latest/curve25519_dalek/scalar/fn.clamp_integer.html always produces a reduced integer (that is reduced)?

So we could use other construction functions and just unwrap? Commenting why it's safe of course.

rust-src/concordium_base/src/ecvrf/secret.rs
```diff
@@ -123,12 +126,18 @@ impl From<&SecretKey> for ExpandedSecretKey {
...
-        lower[31] |= 0b_0100_0000;
+        // It seems like `from_bits` is the only way to construct an unreduced scalar
+        // that can potentially be grater than the field's order.
+
+        // TODO: we use a
+        // deprecated `from_bits` here, this also requires enabling the
+        // `legacy_compatibility` feature for the `ed25519-dalek` dependency. Maybe ther
+        // is a different way of implementing this.
+        #[allow(deprecated)]
+        let scalar = Scalar::from_bits(clamp_integer(lower));
```
abizjak, line 137: Same as above.

rust-src/concordium_base/src/eddsa_ed25519/dlog_ed25519.rs
```diff
@@ -61,20 +61,28 @@ pub enum PointDecodingError {
...
+
+    // It seems like `from_bits` is the only way to construct an unreduced scalar
+    // that can potentially be grater than the field's order.
+
+    // TODO: we use a
+    // deprecated `from_bits` here, this also requires enabling the
+    // `legacy_compatibility` feature for the `ed25519-dalek` dependency. Maybe ther
+    // is a different way of implementing this.
+    #[allow(deprecated)]
+    Scalar::from_bits(clamp_integer(scalar_bytes))
```
abizjak, line 79: As above.

rust-src/concordium_base/src/pedersen_commitment/key.rs
```diff
@@ -217,15 +223,15 @@ mod tests {
         };
     }
 
-    macro_test_commit_open!(commit_open_bls12_381_g1_affine, G1Affine);
+    // macro_test_commit_open!(commit_open_bls12_381_g1_affine, G1Affine);
```
abizjak, line 226: Just remove these (and below).

rust-src/concordium_base/src/pedersen_commitment/key.rs
```diff
@@ -166,9 +170,11 @@ mod tests {
         };
     }
 
-    macro_test_key_byte_conversion!(key_byte_conversion_bls12_381_g1_affine, G1Affine);
+    // NOTE: ArkWorks doesn't provide `CurveGroup` instances for the affine
```
abizjak, line 173: And remove this note.

rust-src/concordium_base/src/pedersen_commitment/randomness.rs
```diff
@@ -97,13 +101,16 @@ mod tests {
         };
     }
 
+    // NOTE: ArkWorks doesn't provide `CurveGroup` instances for the affine
```
abizjak, line 104: Remove note and tests that don't make sense anymore.

rust-src/key_derivation/src/lib.rs
```diff
@@ -257,6 +258,10 @@ impl ConcordiumHdWallet {
         ])?;
         let attribute_commitment_randomness_seed =
             derive_from_parsed_path(&path, &self.seed)?.private_key;
+        println!(
+            "attribute_commitment_randomness_seed: {:?}",
+            attribute_commitment_randomness_seed
+        );
```
abizjak, line 264: Not sure where this got in, but I don't think it should stay.

rust-src/key_derivation/src/lib.rs
```diff
@@ -570,6 +574,10 @@ mod tests {
         let attribute_commitment_randomness = create_wallet(Net::Testnet, TEST_SEED_1)
             .get_attribute_commitment_randomness(5, 0, 4, AttributeTag(0))
             .unwrap();
+        println!(
+            "attribute_commitment_randomness: {:?}",
+            attribute_commitment_randomness
+        );
```
abizjak, line 580: We should remove this debug print.

rust-src/keygen_bls/src/lib.rs
```diff
@@ -20,11 +24,12 @@ pub fn keygen_bls(ikm: &[u8], key_info: &[u8]) -> Result<Fr, hkdf::InvalidLength
...
     l_bytes.push(l);
     let salt = b"BLS-SIG-KEYGEN-SALT-";
-    let mut sk = Fr::zero();
+    let mut sk = <ArkField<Fr>>::zero();
     // shift with
     // 452312848583266388373324160190187140051835877600158453279131187530910662656 =
     // 2^248 = 2^(31*8)
-    let shift = Fr::from_repr(FrRepr([0, 0, 0, 72057594037927936])).unwrap();
+    let shift = <ArkField<Fr>>::from_repr(&[0, 0, 0, 72057594037927936]).unwrap();
+    println!("Shift {:}", shift);
```
abizjak, line 32: That probably should not stay.

abizjak (changes requested):

rust-bins/Cargo.toml
```diff
@@ -13,12 +13,11 @@ vendored-ssl = ["openssl-sys/vendored"]
...
 clap = "2.33"
 dialoguer = "0.10"
-pairing = "0.15"
-rand = "=0.7"
+rand = "=0.8"
 serde = "1.0"
 serde_json = "1.0"
-ed25519-dalek = "=1.0"
-curve25519-dalek = "3.0"
+ed25519-dalek = { version = "2.0", features = ["serde"]}
```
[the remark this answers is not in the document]
annenkov: `serde` is used here https://github.com/Concordium/concordium-base/blob/arkworks-integration/rust-bins/src/bin/keygen.rs#L309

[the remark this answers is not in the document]
abizjak: That should be rendered in hex though. Which is different from this instance.

So we should do
```
base16_encode_string(&sig)
```
if the type is the dalek Signature.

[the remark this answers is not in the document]
annenkov: Is my understanding correct, that in this fragment
```Rust
let sig = keypair.sign(&public_bytes);
    let level_str = match kgup.level {
        Level::Root => "root",
        Level::One => "level1",
        Level::Two => "level2",
    };
let public_data = serde_json::json!({
        "key": {
            "verifyKey": base16_encode_string(&signing_key.verifying_key()),
            "scheme": "Ed25519",
        },
        "signature": sig,
        "type": level_str,
    });
```
I change ` "signature": sig`  to `"signature": base16_encode_string(&sig)`. Then there is no need for a `SerdeSerialize` instance for the signature (since it's a hex string now).
,

idiss/src/lib.rs
```diff
@@ -14,11 +15,13 @@ use concordium_base::{
     },
     ps_sig,
 };
-use pairing::bls12_381::{Bls12, G1};
 use serde::{Deserialize as SerdeDeserialize, Serialize as SerdeSerialize};
 #[cfg(feature = "nodejs")]
 use serde_json::ser::to_string;
 
+type Bls12 = ark_ec::bls12::Bls12<ark_bls12_381::Config>;
+type G1 = ArkGroup<G1Projective>;
```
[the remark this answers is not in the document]
annenkov: Fixed. I use `id::constants::ArCurve` and `id::constants::IpPairing` everywhere where `concordium_base` is imported. I also defined `id::constants::ArCurveG2` to avoid `arkworks` dependency where `G2` is used

mobile_wallet/Cargo.toml
```diff
@@ -8,21 +8,21 @@ license-file = "../../LICENSE-APACHE"
...
-ed25519-dalek = "=1.0.0"
+ed25519-dalek = "2.0"
 byteorder = "1.3"
 either = "1.6"
 sha2 = "0.10"
 libc = "0.2"
 thiserror = "1.0"
 base64 = "0.21"
+ark-bls12-381 = { version = "0.4"}
+ark-ec = { version = "0.4"}
```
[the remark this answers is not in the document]
annenkov: Removed now. Use constants from `id::constants` instead.

mobile_wallet/src/lib.rs
```diff
@@ -46,6 +45,8 @@ use std::{
     str::FromStr,
 };
 
+type Bls12 = ark_ec::bls12::Bls12<ark_bls12_381::Config>;
```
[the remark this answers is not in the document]
annenkov: Done. Here in in other similar places.

rust-bins/Cargo.toml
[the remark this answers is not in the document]
annenkov: Fixed as it is described above.

rust-src/concordium_base/CHANGELOG.md
```diff
@@ -1,8 +1,15 @@
...
 
-- Improve performance of `multiexp*` family of functions.
-- Add traits `Field` and `PrimeField` with implementations for the underlying field of the `BLS12-381`` curve.
 - Add `MultiExp` trait that allows to have different `multiexp` algorithm implementations for different curves.
+- Improve performance of the generic `multiexp` algorithm.
+- Add an instance of `MultiExp` that is specific to `curve25519`.
+- Add traits `Field` and `PrimeField` with implementations for the underlying field of the `BLS12-381` curve.
+- Add integration with the `arkworks` library interfaces for fields and elliptic curves (wrapper types and blanket trait implementations).
+- Add the `BLS12-381`implementation from the `arkworks` ecosystem.
+- Remove the `pairing` crate from dependencies along with the corresponding `BLS12-381` code (replaced with the `arkworks` implementation).
```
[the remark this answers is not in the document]
annenkov: Fixed.

rust-src/concordium_base/CHANGELOG.md
```diff
@@ -1,8 +1,15 @@
...
+- Upgrade `ed25519-dalek` to `v2.0`. This leads to changes in FFI:
```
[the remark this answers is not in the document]
annenkov: Fixed. Removed mentions of FFI.

rust-src/concordium_base/Cargo.toml
```diff
@@ -12,28 +12,32 @@ repository = "https://github.com/Concordium/concordium-base"
...
 [dependencies]
-ff = "0.5"
+ff = "0.13"
+ark-ff = { version = "0.4"}
+ark-ec = { version = "0.4"}
+ark-serialize = { version = "0.4"}
+ark-std = { version = "0.4"}
+ark-bls12-381 = { version = "0.4"}
 sha2 = "0.10"
+sha2-old = { version = "0.9", package = "sha2" }
```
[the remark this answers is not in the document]
annenkov: There were some outdated libraries, but we don't need the `sha2-old` now, you're right.
Removed.

rust-src/concordium_base/Cargo.toml
```diff
@@ -12,28 +12,32 @@ repository = "https://github.com/Concordium/concordium-base"
...
 either = "1.6"
-pairing = "0.15"
 derive_more = "0.99"
 thiserror = "1.0"
-rand = "0.7"
+rand = "0.8"
 num = "0.4"
 group = "0.2"
-curve25519-dalek = "3"
+curve25519-dalek = { version = "4.1.1", features = ["rand_core", "group"]}
```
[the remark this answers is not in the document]
annenkov: No, no specific reason, `4.1` is fine.

rust-src/concordium_base/src/ecvrf/proof.rs
annenkov: Yes, indeed. The documentation says "The Scalar struct holds an element of Z/ℓZ.". The only way to break the invariant that the scalar is reduced is to use `from_bits()`, which we don't use now. Moreover, `reduce()`  is private now.

rust-src/key_derivation/src/lib.rs
annenkov: Removed here in in other places, where it should not be.

rust-src/concordium_base/src/curve_arithmetic/arkworks_instances.rs
annenkov: Done.

rust-src/concordium_base/src/curve_arithmetic/arkworks_instances.rs
annenkov: I made the implementation generic. From the documentation of the trait, it is clear what is expected form the implementation.

rust-src/concordium_base/src/curve_arithmetic/arkworks_instances.rs
```diff
@@ -0,0 +1,207 @@
...
+    fn generate<R: rand::prelude::Rng>(rng: &mut R) -> Self { ArkGroup(G::rand(rng)) }
+
+    fn generate_scalar<R: rand::prelude::Rng>(rng: &mut R) -> Self::Scalar {
+        <G::ScalarField as ark_ff::UniformRand>::rand(rng).into()
+    }
+
+    fn scalar_from_u64(n: u64) -> Self::Scalar { ArkField(G::ScalarField::from(n)) }
+
+    fn scalar_from_bytes<A: AsRef<[u8]>>(bs: A) -> Self::Scalar {
+        // Traverse at most 4 8-byte chunks, for a total of 256 bits.
```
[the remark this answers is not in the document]
annenkov: Fixed.

rust-src/concordium_base/src/pedersen_commitment/key.rs
annenkov: Done.

rust-src/concordium_base/src/ecvrf/mod.rs
annenkov: Changed to use `from_bytes_mod_order()`

rust-src/concordium_base/src/eddsa_ed25519/dlog_ed25519.rs
annenkov: Changed to use `from_bytes_mod_order()`

rust-src/concordium_base/src/curve_arithmetic/secret_value.rs
annenkov: Done.

rust-src/concordium_base/src/ecvrf/secret.rs
annenkov: CHanged to use `from_bytes_mod_order()`

rust-src/concordium_base/src/ecvrf/public.rs
annenkov: `clamp_integer()` produces something that could be bigger than the order.
This function is removed, no mangling is required at this point. The scalar is properly computed during the creation of an `ExpandedSecretKey`.

rust-src/concordium_base/src/curve_arithmetic/field_adapters.rs
annenkov: The Ristretto representation from the `curve25519-dalek` implements the `ff` traits. The crate is relatively widely used, so it could be useful for integrating with other elliptic curve implementations.

I'll add this to the documentation.

annenkov: > Does ark not have ristretto?

They have curve25519 implementation, but not specifically the Ristretto representation, as far as I remember. That implementation is less used than the dalek one. Maybe this is something to look at in the future.

rust-src/concordium_base/src/curve_arithmetic/ed25519_instance.rs
annenkov: Yeah, they use constant-time option type `CtOption`, which does not have `expect()`, conversion it to `Option` requires some type annotations :)
I'll do this properly :)

annenkov: Fixed now.

rust-src/concordium_base/src/curve_arithmetic/arkworks_instances.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+/// Curve configuration.
+///
+/// These parameters cannot be taken from the `arkworks` traits. Each `arkworks`
+/// curve should come with an implementation of this configuration trait.
+pub(crate) trait ArkCurveConfig<G: ark_ec::CurveGroup> {
+    const SCALAR_LENGTH: usize;
+    const GROUP_ELEMENT_LENGTH: usize;
+    const DOMAIN_STRING: &'static str;
```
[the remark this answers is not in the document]
annenkov: Fixed.

rust-src/concordium_base/src/curve_arithmetic/arkworks_instances.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    fn double_point(&self) -> Self { ArkGroup(self.0.double()) }
+
+    fn plus_point(&self, other: &Self) -> Self { ArkGroup(self.0 + other.0) }
+
+    fn minus_point(&self, other: &Self) -> Self { ArkGroup(self.0 - other.0) }
+
+    fn mul_by_scalar(&self, scalar: &Self::Scalar) -> Self { ArkGroup(self.0 * scalar.0) }
+
+    fn bytes_to_curve_unchecked<R: byteorder::ReadBytesExt>(b: &mut R) -> anyhow::Result<Self> {
```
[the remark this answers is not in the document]
annenkov: Yes, there is only one place in ElGamal
https://github.com/Concordium/concordium-base/blob/arkworks-integration/rust-src/concordium_base/src/elgamal/cipher.rs#L59

And it already returns `Result`, so using the checked version is fine, but it will be a bit slower.
Also in this case `Cipher::from_bytes_unchecked()` should be renamed into `Cipher::from_bytes()`, but it's a public method, maybe this is not desirable.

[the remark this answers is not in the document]
annenkov: I removed `Curve::bytes_to_curve_unchecked()` and  renamed `Cipher::from_bytes_unchecked()` to `Cipher::from_bytes()`; the method uses `deserial()` now instead of `Curve::bytes_to_curve_unchecked()`.

rust-src/concordium_base/benches/eddsa_benchmarks.rs
```diff
@@ -32,13 +28,10 @@ pub fn bench_sign(c: &mut Criterion) {
...
     let d = a.clone();
 
-    let sk = SecretKey::generate(&mut csprng);
-    let pk = PublicKey::from(&sk);
-    let expanded_sk = ExpandedSecretKey::from(&sk);
-    let sig = expanded_sk.sign(a.as_slice(), &pk);
-    c.bench_function("sign {}", move |b| {
-        b.iter(|| expanded_sk.sign(a.as_slice(), &pk))
-    });
+    let signing = SigningKey::generate(&mut csprng);
```
hamiidreza, line 31: minor comment: I would call this something else like signing_key or sk

rust-src/ed25519_hd_key_derivation/src/lib.rs
```diff
@@ -187,8 +189,8 @@ mod tests {
     fn assert_keys(seed: &str, path: &str, chain_code: &str, private_key: &str, public_key: &str) {
         let seed = hex::decode(seed).unwrap();
         let keys = derive(path, &seed).unwrap();
-        let secret_key = ed25519_dalek::SecretKey::from_bytes(&keys.private_key).unwrap();
-        let public_key_derived = ed25519_dalek::PublicKey::from(&secret_key).to_bytes();
+        let signing = SigningKey::from_bytes(&keys.private_key);
```
hamiidreza, line 192: same comment: I would call it signing_key or sk.

abizjak:
Just a reminder, one thing we must check is that all of these new dependencies actually work fine in Wasm as well. Since that's what the browsre wallet needs.

rust-src/concordium_base/src/ecvrf/proof.rs
```diff
@@ -29,24 +29,26 @@ pub fn hash_points(pts: &[CompressedEdwardsY]) -> Scalar {
 /// a given public key.
 pub struct Proof(pub EdwardsPoint, pub Scalar, pub Scalar);
 
-/// Implements step 8 of <https://tools.ietf.org/id/draft-irtf-cfrg-vrf-07.html#rfc.section.5.1>
```
hamiidreza, line 32: As we discussed, we may just correct the link without transitioning to the stable version!

rust-src/concordium_base/src/ecvrf/proof.rs
```diff
@@ -29,24 +29,26 @@ pub fn hash_points(pts: &[CompressedEdwardsY]) -> Scalar {
...
+/// Implements step 8 of <https://www.rfc-editor.org/rfc/rfc9381.html#name-ecvrf-proving>
```
hamiidreza, line 32:
```suggestion
/// Implements step 8 of 
<https://tools.ietf.org/html/draft-irtf-cfrg-vrf-07.html#section-5.1>
```

rust-src/concordium_base/src/ecvrf/proof.rs
```diff
@@ -29,24 +29,26 @@ pub fn hash_points(pts: &[CompressedEdwardsY]) -> Scalar {
...
         x.write_all(&c[..16])
             .expect("Writing to buffer should succeed.");
-        x.write_all(&self.2.reduce().to_bytes()[..])
+        x.write_all(&self.2.to_bytes()[..])
             .expect("Writing to buffer should succeed.");
     }
 }
 
-/// Implements <https://tools.ietf.org/id/draft-irtf-cfrg-vrf-07.html#rfc.section.5.4.4>
+/// Implements <https://www.rfc-editor.org/rfc/rfc9381.html#name-ecvrf-decode-proof>
```
hamiidreza, line 51:
```suggestion
/// Implements <https://tools.ietf.org/html/draft-irtf-cfrg-vrf-07.html#section-5.4.4>
```

rust-src/concordium_base/src/ecvrf/proof.rs
```diff
@@ -78,7 +80,7 @@ impl Debug for Proof {
     }
 }
 
-/// Implements <https://tools.ietf.org/id/draft-irtf-cfrg-vrf-07.html#rfc.section.5.2>
+/// Implements <https://www.rfc-editor.org/rfc/rfc9381.html#name-ecvrf-proof-to-hash>
```
hamiidreza, line 83:
```suggestion
/// Implements <https://tools.ietf.org/html/draft-irtf-cfrg-vrf-07.html#section-5.2>
```

rust-src/concordium_base/src/ecvrf/public.rs
```diff
@@ -88,25 +82,11 @@ impl PublicKey {
...
-        bits[31] |= 0b_0100_0000;
-
-        let point = &Scalar::from_bits(*bits) * &constants::ED25519_BASEPOINT_TABLE;
-        let compressed = point.compress();
-
-        PublicKey(compressed, point)
-    }
-
-    /// Implements <https://tools.ietf.org/id/draft-irtf-cfrg-vrf-07.html#rfc.section.5.4.1.1>
+    /// Implements <https://https://www.rfc-editor.org/rfc/rfc9381.html#name-ecvrf_encode_to_curve_try_a>
```
hamiidreza, line 85:
```suggestion
    /// Implements <https://tools.ietf.org/html/draft-irtf-cfrg-vrf-07.html#section-5.4.1.1>
```

rust-src/concordium_base/src/ecvrf/public.rs
```diff
@@ -137,15 +117,15 @@ impl PublicKey {
 
     pub fn verify_key(&self) -> bool { !self.1.is_small_order() }
 
-    /// Implements <https://tools.ietf.org/id/draft-irtf-cfrg-vrf-07.html#rfc.section.5.3>
+    /// Implements <https://www.rfc-editor.org/rfc/rfc9381.html#name-ecvrf-verifying>
```
hamiidreza, line 120:
```suggestion
    /// Implements <https://tools.ietf.org/html/draft-irtf-cfrg-vrf-07.html#section-5.3>
```

rust-src/concordium_base/src/ecvrf/secret.rs
```diff
@@ -137,7 +136,7 @@ use super::proof::*;
 
 impl ExpandedSecretKey {
     /// VRF proof with expanded secret key
-    /// Implements <https://tools.ietf.org/id/draft-irtf-cfrg-vrf-07.html#rfc.section.5.1>
+    /// Implements <https://www.rfc-editor.org/rfc/rfc9381.html#name-ecvrf-proving>
```
hamiidreza, line 139:
```suggestion
    /// Implements <https://tools.ietf.org/html/draft-irtf-cfrg-vrf-07.html#section-5.1>
```

rust-src/concordium_base/src/ecvrf/secret.rs
```diff
@@ -160,7 +159,7 @@ impl ExpandedSecretKey {
         Proof(gamma, c, k_plus_cx)
     }
 
-    /// Implements <https://tools.ietf.org/id/draft-irtf-cfrg-vrf-07.html#rfc.section.5.4.2.2>
+    /// Implements <https://www.rfc-editor.org/rfc/rfc9381.html#name-ecvrf-nonce-generation-from->
```
hamiidreza, line 162:
```suggestion
    /// Implements <https://tools.ietf.org/html/draft-irtf-cfrg-vrf-07.html#section-5.4.2.2>
```

hamiidreza (commented):
Looks good. I added some suggestions regarding the broken link for the VRF RFC draft version.

rust-src/concordium_base/src/curve_arithmetic/arkworks_instances.rs
annenkov: It is changed now to return `Result`. It the usage points, the return values are unwrapped.

rust-src/concordium_base/src/ecvrf/public.rs
```diff
@@ -88,25 +82,11 @@ impl PublicKey {
...
-        let compressed = point.compress();
-
-        PublicKey(compressed, point)
-    }
-
-    /// Implements <https://tools.ietf.org/id/draft-irtf-cfrg-vrf-07.html#rfc.section.5.4.1.1>
+    /// Implements <https://https://www.rfc-editor.org/rfc/rfc9381.html#name-ecvrf_encode_to_curve_try_a>
     /// The failure should not happen in practice, expected number of iterations
     /// is 2.
+    // TODO: Check whether it still implements the algorithm after the RFC draft was
```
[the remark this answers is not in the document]
annenkov: Removed now; I accepted changes with the correct URL pointing to the relevant draft.

rust-src/concordium_base/src/ecvrf/proof.rs
annenkov: Done; we use the correct URL pointing to the draft version now.

rust-src/concordium_base/benches/eddsa_benchmarks.rs
annenkov: Fixed.

rust-src/ed25519_hd_key_derivation/src/lib.rs
annenkov: Fixed.

abizjak:
@annenkov did you propagate this PR to concordium-client & the node?

rust-src/concordium_base/src/id/constants.rs
```diff
@@ -18,12 +19,15 @@ use std::{fmt, io::Cursor, str::FromStr};
...
+pub type BlsG2 = ArkGroup<G2Projective>;
 /// Pairing used by the identity provider.
-pub type IpPairing = pairing::bls12_381::Bls12;
+pub type IpPairing = Bls12<ark_bls12_381::Config>;
 /// Field used by the identity provider and anonymity revoker.
-/// This isthe base field of both the ArCurve and the IpPairing.
-pub type BaseField = <pairing::bls12_381::Bls12 as Pairing>::ScalarField;
+/// This is the scalar field of both the ArCurve and the IpPairing.
+pub type BaseField = <IpPairing as Pairing>::ScalarField;
+// pub type BaseField = ArkField<Fr>;
```
abizjak, line 30: Why is this still here?

abizjak (approved):

rust-src/concordium_base/src/id/constants.rs
annenkov: Good point, I'll remove it.

annenkov:
> @annenkov did you propagate this PR to concordium-client & the node?

Yes, I checked `concordium-client` and `concordium-node` can be built with this branch. There are only changes in the signatures of functions used internally.
However, it would make sense to change signatures of the `sign` function(s). I created an issue for this https://github.com/Concordium/concordium-base/issues/501

rust-src/concordium_base/src/curve_arithmetic/arkworks_instances.rs
```diff
@@ -0,0 +1,205 @@
...
+        <G::ScalarField as ark_ff::UniformRand>::rand(rng).into()
+    }
+
+    fn scalar_from_u64(n: u64) -> Self::Scalar { ArkField(G::ScalarField::from(n)) }
+
+    fn scalar_from_bytes<A: AsRef<[u8]>>(bs: A) -> Self::Scalar {
+        // Traverse at most `ceil(CAPACITY / 8)` 8-byte chunks.
+        let s = num::integer::div_ceil(Self::Scalar::CAPACITY, 8);
+        let mut fr = Vec::with_capacity(s as usize);
+        for chunk in bs.as_ref().chunks(8).take(s as usize) {
```
abizjak, line 182:
```suggestion
        for (chunk, place) in bs.as_ref().chunks(8).take(s as usize).zip(&mut fr) {
```

rust-src/concordium_base/src/curve_arithmetic/bls12_381_arkworks.rs
```diff
@@ -0,0 +1,597 @@
...
+            let m = m.into_repr();
+            let mask = !(1u64 << 63 | 1u64 << 62);
+            assert_eq!(n[0], m[0], "First limb.");
+            assert_eq!(n[1], m[1], "Second limb.");
+            assert_eq!(n[2], m[2], "Third limb.");
+            assert_eq!(n[3] & mask, m[3] & mask, "Fourth limb with top bit masked.");
+        }
+    }
+
+    /// Test that `into_repr()` correclty converts a scalar constructed from a
```
eb-concordium, line 192:
```suggestion
    /// Test that `into_repr()` correctly converts a scalar constructed from a
```

eb-concordium (approved):

eb-concordium (approved):

PR #498, opened 2024-01-10
Remove seed phrase input types

orhoj:
## Purpose
While implementing the Java SDK parts that utilizes the wallet_library we found some inconveniences that this PR is trying to clean up.
1. We don't want to encourage using the seed phrase directly when generating identity and credential requests. Therefore this code is removed from the wallet_library, and is instead only implemented in the web-sdk to ensure backwards compatiblity there. The Java SDK will only supply functions for using the keys directly.
2. Removed the shared `common` types that no longer make any sense as there is now only a single input type and function. This in particular makes the types in the other SDK's nicer as they don't have to contain this `common` field only used for the JSON serialization.
3. Tried to align the naming of types and functions in `identity.rs` a little.

## Changes
- Removed methods for generating identity object, identity recovery and credential deployment requests from a seed phrase.
- Changed some naming.

Note: I haven't bumped the version as the library has still not been used yet in anything outside of base.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/wallet_library/src/identity.rs
+++ b/rust-src/wallet_library/src/identity.rs
@@ -18,7 +18,7 @@ use std::collections::BTreeMap;
 type JsonString = String;
 
 /// Defines the JSON structure that matches the output of the function
-/// generating the identity objcect request.
+/// generating the identity object request.
 #[derive(SerdeSerialize, SerdeDeserialize)]
 #[serde(rename_all = "camelCase")]
 struct IdentityObjectRequestV1 {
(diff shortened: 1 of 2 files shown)


rust-src/wallet_library/src/identity.rs
```diff
@@ -116,88 +84,31 @@ pub fn create_id_request_with_keys_v1_aux(input: IdRequestInputWithKeys) -> Resu
...
-    seed_as_hex:    String,
-    net:            Net,
-    identity_index: u32,
+    id_cred_sec:    PedersenValue<ArCurve>,
 }
 
 #[derive(SerdeSerialize, SerdeDeserialize)]
 #[serde(rename_all = "camelCase")]
-pub struct IdRecoveryRequestOut {
+pub struct IdentityRecoveryRequestOut {
```
abizjak, line 99: If I wanted to be nitpicky, I would suggest to call this `Output` with analogy to `Input`. :)

abizjak (approved):

PR #499, opened 2024-01-12
Unify Timestamp in base and contracts_common

### LINKED ISSUE #429: There are two Timestamp types in rust-src/concordium-base
**Task description**

There are two timestamp types (for timestamps in milliseconds). One defined in contracts-common and one in base.

This is confusing to users of the sc-testing library and the rust SDK.

They are almost the same, the only difference being the JSON serialization, which is based on u64 for the base one, and rfc3393 for the one in contracts-common.

The type in base is mostly not used, except in a few places in the Rust SDK where the specific serialization still does matter.

We should unify those types, and just re-export the one from contracts-common from base.

Where we need the specific JSON instances we should migrate to doing the parsing in a different way.

### LINKED ISSUE #450: Implement SchemaType trait for &'static str
**Task description**

Add trait `SchemaType` for `&'static str`. This will allow us to return a reference to a static string in a smart contract.
 
 ```
 #[receive(
    contract = "blablabla",
    name = "blabla",
    return_value = "&'static str"
)]
fn blabla<S: HasStateApi>(
    _ctx: &impl HasReceiveContext,
    _host: &impl HasHost<State, StateApiType = S>,
) -> ReceiveResult<&'static str> {
      "abc"
}
 ```

limemloh:
## Purpose

Closes #429 and closes #450.

## Changes

- Unify `Timestamp` from `concordium_base` and `concordium_contracts_common` in `concordium_contracts_common`.
- Add `SchemaType` implementation for `&str`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/CHANGELOG.md
+++ b/rust-src/concordium_base/CHANGELOG.md
@@ -9,9 +9,6 @@
   The alternate formatter (using `#`) displays it as a list of bytes.
 - Add `FromStr` and `Display` instances to `dodis_yampolskiy_prf::SecretKey`.
 - Change `Debug` instance of `dodis_yampolskiy_prf::SecretKey` to hide the value.
-
-### Breaking changes
-
 - Remove `Timestamp` to instead reexport the similar type from `concordium_contracts_common`.
   This adds several new methods, but results in a breaking change in the `serde::Serialize` implementation, which is now using string containing RFC3393 representation instead the underlying milliseconds.
 
--- a/rust-src/concordium_base/CHANGELOG.md
+++ b/rust-src/concordium_base/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Unreleased changes
 
 - Improve performance of `multiexp*` family of functions.
-- Add traits `Field` and `PrimeField` with implementations for the underlying field of the `BLS12-381`` curve.
+- Add traits `Field` and `PrimeField` with implementations for the underlying field of the `BLS12-381` curve.
 - Add `MultiExp` trait that allows to have different `multiexp` algorithm implementations for different curves.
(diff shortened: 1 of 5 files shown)


rust-src/concordium_base/CHANGELOG.md
```diff
@@ -10,6 +10,11 @@
 - Add `FromStr` and `Display` instances to `dodis_yampolskiy_prf::SecretKey`.
 - Change `Debug` instance of `dodis_yampolskiy_prf::SecretKey` to hide the value.
 
+### Breaking changes
```
abizjak, line 13: You've introduced this section now, but a lot of breaking changes have not been included in it.

Since we have not been grouping changes before in the changelog don't just start it like this.

smart-contracts/contracts-common/concordium-contracts-common/src/types.rs
```diff
@@ -653,7 +653,15 @@ pub struct CredentialSignatures {
 #[cfg_attr(feature = "fuzz", derive(Arbitrary))]
 pub struct Timestamp {
     /// Milliseconds since unix epoch.
-    pub(crate) milliseconds: u64,
```
abizjak, line 656: I can't find the FromStr implementation for this and the base timestamp has it.

I think ti should support parsing both from a u64 as milliseconds and the RFC... format (Dual to display).

limemloh: The implementation can be found here https://github.com/Concordium/concordium-base/pull/499/files#diff-5b114d32401abf835eb7e1165d42c0d36970eab7c96eda1d0e121601fe7e105dR794
And it already supports both u64 strings and the RFC... format.

abizjak (approved):

PR #500, opened 2024-01-15
Unify WasmVersion into concordium_contracts_common

### LINKED ISSUE #465: Unify WasmVersions
**Task description**

We define a WasmVersion type both in concordium_base and in wasm-chain-integration.

This is annoying at least, and causes usability issues.

It's a bit unclear what the best solution is.

One option is to add this type to `contracts_common` and re-export it from the two libraries. It looks like the trait implementations are compatible already so this should not be any breaking change.

limemloh:
## Purpose

Closes #465.
This PR unifies `WasmVersion` in `concordium-base` and `wasm-chain-integration` into one type in `concordium-contracts-common`.
See also the PR updating `cargo-concordium` to use the unified type https://github.com/Concordium/concordium-smart-contract-tools/pull/151

## Changes

- Removes `WasmVersion::read` in favor of `common::Deserial` in `concordium-base`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/smart-contracts/wasm-chain-integration/src/utils.rs
+++ b/smart-contracts/wasm-chain-integration/src/utils.rs
@@ -17,39 +17,6 @@ use concordium_wasm::{
 use rand::{prelude::*, RngCore};
 use std::{collections::BTreeMap, default::Default};
 
-#[derive(Debug, Clone, Copy)]
-pub enum WasmVersion {
-    V0,
-    V1,
-}
-
-impl std::str::FromStr for WasmVersion {
-    type Err = anyhow::Error;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s {
-            "V0" | "v0" => Ok(WasmVersion::V0),
-            "V1" | "v1" => Ok(WasmVersion::V1),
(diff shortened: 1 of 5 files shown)


smart-contracts/wasm-chain-integration/src/utils.rs
```diff
@@ -17,39 +18,6 @@ use concordium_wasm::{
...
-            "V1" | "v1" => Ok(WasmVersion::V1),
-            _ => anyhow::bail!("Unsupported version: '{}'. Only 'V0' and 'V1' are supported.", s),
-        }
-    }
-}
-
-impl WasmVersion {
-    /// Get the version from the cursor. This is not a Serial implementation
-    /// since it uses big-endian.
-    pub fn read(source: &mut std::io::Cursor<&[u8]>) -> anyhow::Result<WasmVersion> {
```
abizjak, line 41: Is this not needed anywhere? Cause I don't see it moved.

limemloh: The place where I found it used was in `cargo-concordium`, but there we have access to `Deserial::deserial` (the BE one) which behaves the same, so I removed it.

abizjak: Ok.

abizjak (approved):

PR #502, opened 2024-01-17
Credential deployment serialization

orhoj:
## Purpose
To reduce serialization code duplication the code for serializing a new credential for signing and for submission to the node is added to `wallet_library`.

## Changes
- Added methods for serializing the credential deployment details (that which must be serialized, hashed and signed) and the credential deployment payload (that which is sent as a raw payload to the node).

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/wallet_library/src/credential.rs
+++ b/rust-src/wallet_library/src/credential.rs
@@ -151,18 +151,16 @@ pub struct CredentialDeploymentPayload {
 /// Serializes the credential deployment payload. The result of this
 /// serialization should be sent as a raw payload to the node. The serialized
 /// bytes are returned hex encoded.
-pub fn serialize_credential_deployment_payload(
-    payload: CredentialDeploymentPayload,
-) -> Result<String> {
-    let cdi = get_credential_deployment_info(payload.signatures, payload.unsigned_cdi)?;
+pub fn serialize_credential_deployment_payload(payload: CredentialDeploymentPayload) -> String {
+    let cdi = get_credential_deployment_info(payload.signatures, payload.unsigned_cdi);
     let acc_cred = AccountCredential::Normal { cdi };
-    Ok(base16_encode_string(&acc_cred))
+    base16_encode_string(&acc_cred)
 }
 
 fn get_credential_deployment_info(
     signatures: BTreeMap<KeyIndex, AccountOwnershipSignature>,
     unsigned_cdi: UnsignedCredentialDeploymentInfo<IpPairing, ArCurve, AttributeKind>,
-) -> Result<CredentialDeploymentInfo<IpPairing, ArCurve, AttributeKind>, anyhow::Error> {
+) -> CredentialDeploymentInfo<IpPairing, ArCurve, AttributeKind> {
     let proof_acc_sk = AccountOwnershipProof { sigs: signatures };
 
     let cdp = CredDeploymentProofs {
@@ -170,10 +168,10 @@ fn get_credential_deployment_info(
         proof_acc_sk,
     };
 
-    Ok(CredentialDeploymentInfo {
+    CredentialDeploymentInfo {
         values: unsigned_cdi.values,
         proofs: cdp,
-    })
+    }
 }
 
 #[cfg(test)]
@@ -298,7 +296,7 @@ mod tests {
             signatures,
         };
 
-        let serialized_payload = serialize_credential_deployment_payload(payload).unwrap();
+        let serialized_payload = serialize_credential_deployment_payload(payload);
 
         assert!(serialized_payload.contains("0101000029723ec9a0b4ca16d5d548b676a1a0adbecdedc5446894151acb7699293d69b101b317d3fea7de56f8c96f6e72820c5cd502cc0eef8454016ee548913255897c6b52156cc60df965d3efb3f160eff6ced40000000001000300000001"));
         assert!(serialized_payload.contains(signature_hex));
--- a/rust-src/wallet_library/src/credential.rs
+++ b/rust-src/wallet_library/src/credential.rs
@@ -1,6 +1,10 @@
 use anyhow::{bail, Result};
 use concordium_base::{
-    common::{types::TransactionTime, Buffer, Serial},
+    common::{
+        base16_decode_string,
+        types::{KeyIndex, TransactionTime},
+        Buffer, Serial,
+    },
     id::{
         account_holder::create_unsigned_credential,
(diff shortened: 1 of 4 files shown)


rust-src/wallet_library/src/credential.rs
```diff
@@ -111,6 +117,102 @@ fn build_policy(
...
+pub fn serialize_credential_deployment_payload(
+    payload: CredentialDeploymentPayload,
+) -> Result<String> {
+    let cdi = get_credential_deployment_info(payload.signatures, payload.unsigned_cdi)?;
+    let acc_cred = AccountCredential::Normal { cdi };
+
+    let mut acc_cred_ser = Vec::<u8>::new();
+    acc_cred.serial(&mut acc_cred_ser);
+
+    Ok(hex::encode(acc_cred_ser))
```
abizjak, line 171:
```suggestion
    Ok(base16_encode_string(&acc_cred))
```

rust-src/wallet_library/src/credential.rs
```diff
@@ -111,6 +117,102 @@ fn build_policy(
...
+/// Serializes the credential deployment details. It is this serialization that
+/// should be signed when creating a new credential for deployment. The
+/// serialized bytes are returned hex encoded.
+pub fn serialize_credential_deployment_details(
+    credential_deployment_details: CredentialDeploymentDetails,
+) -> String {
+    let mut serialized_details = Vec::<u8>::new();
+    credential_deployment_details.serial(&mut serialized_details);
+    hex::encode(serialized_details)
+}
```
abizjak, line 148: Do we need the actual data, or do we just need the hash?

There is this helper function
```rust
|pub fn credential_hash_to_sign<                                                                                                                                
            }                                                                                                                                                 |    P: Pairing,                                                                                                                                                
            return Ok(());                                                                                                                                    |    C: Curve<Scalar = P::ScalarField>,                                                                                                                         
        }                                                                                                                                                     |    AttributeType: Attribute<C::Scalar>,                                                                                                                       
    }                                                                                                                                                         |>(                                                                                                                                                             
                                                                                                                                                              |    values: &CredentialDeploymentValues<C, AttributeType>,                                                                                                     
    if let Some(wallet_path) = wallet_path {                                                                                                                  |    proofs: &IdOwnershipProofs<P, C>,                                                                                                                          
        let wallet = WalletAccount::from_json_file(wallet_path)?;                                                                                             |    new_or_existing: &Either<TransactionTime, AccountAddress>,                                                                                                 
        let param = OwnedParameter::from_serial(data: &weights)?;                                                                                             |) -> Vec<u8>
```

which would be ideal to use instead of redefining.

rust-src/wallet_library/src/credential.rs
```diff
@@ -111,6 +117,102 @@ fn build_policy(
...
+
+    Ok(CredentialDeploymentInfo {
+        values: unsigned_cdi.values,
+        proofs: cdp,
+    })
+}
+
+fn build_signature_map(
+    signatures: &[String],
+) -> Result<BTreeMap<KeyIndex, AccountOwnershipSignature>, anyhow::Error> {
```
abizjak, line 196: Can you elaborate on this function. It seems there are some implicit assumptions about indices and order for the input which would be good to document.

rust-src/wallet_library/src/credential.rs
```diff
@@ -111,6 +117,102 @@ fn build_policy(
...
+    signatures
+        .iter()
+        .enumerate()
+        .map(|(index, key)| {
+            let index_u8: u8 = match index.try_into() {
+                Ok(idx) => idx,
+                Err(err) => return Err(anyhow!(err)),
+            };
```
abizjak, line 204:
```suggestion
            let index_u8: u8 =index.try_into()?;
```

rust-src/wallet_library/src/credential.rs
```diff
@@ -111,6 +117,102 @@ fn build_policy(
...
+
+            let decoded_key = match base16_decode_string(key) {
+                Ok(decoded) => decoded,
+                Err(e) => return Err(anyhow!(e)),
+            };
```
abizjak, line 209:
```suggestion
            let decoded_key = base16_decode_string(key)?;
```

rust-src/wallet_library/src/credential.rs
```diff
@@ -111,6 +117,102 @@ fn build_policy(
...
+        self.unsigned_cdi.serial(out);
+        0u8.serial(out);
+        self.expiry.seconds.serial(out);
+    }
+}
+
+/// Serializes the credential deployment details. It is this serialization that
+/// should be signed when creating a new credential for deployment. The
+/// serialized bytes are returned hex encoded.
+pub fn serialize_credential_deployment_details(
```
abizjak, line 142: I'm wondering whether we need this specifically, or just the hash of this which I think is the thing that is actually signed.

rust-src/wallet_library/src/credential.rs
orhoj: We just need the hash. I am not sure why I didn't just have it return that. I tried looking through concordium-base for an existing function like `credential_hash_to_sign` but failed to find it, so it is great that you pointed out that one already exists. I will refactor to use that.

rust-src/wallet_library/src/credential.rs
orhoj: We just need the hash. I am refactoring it.

rust-src/wallet_library/src/credential.rs
```diff
@@ -195,7 +296,47 @@ mod tests {
...
+
+        let details = CredentialDeploymentDetails {
+            expiry,
+            unsigned_cdi,
+        };
+
+        let serialized_details = serialize_credential_deployment_details(details);
+
+        assert!(serialized_details.contains("01000029723ec9a0b4ca16d5d548b676a1a0adbecdedc5446894151acb7699293d69b101b317d3fea7de56f8c96f6e72820c5cd502cc0eef8454016ee548913255897c6b52156cc60df965d3efb3f160eff6ced40000000001000300000001"));
+        assert!(serialized_details.contains("0685810406858104"));
```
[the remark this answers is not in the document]
orhoj: I have changed this as it now tests on the hash instead. The hash is different on each run due to the randomness generated in the function, so I just check that the hash length is correct.

rust-src/wallet_library/src/credential.rs
orhoj: I am removing this function and we will provide the map with keys directly instead.

abizjak (approved):

PR #504, opened 2024-01-21
Update bs58 and ed25519-zebra dependencies.

abizjak:
## Purpose

Update bs58 and ed25519-zebra dependencies to avoid duplicate transitive dependencies after the ed25519-dalek upgrade.

## Changes

- The zebra upgrade has no API changes, simply an upgrade of curve25519-dalek.
- The bs58 has a few API changes, the only one relevant for us is renaming into to onto.
- Use ChaCha20Rng instead of StdRng. In the update from 0.7 to 0.8 of rand the StdRng was changed from ChaCha20 to ChaCha12 leading to some test failures due to hardcoded constants.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/identity-provider-service/Cargo.lock
+++ b/identity-provider-service/Cargo.lock
@@ -490,7 +490,6 @@ dependencies = [
  "ed25519-dalek",
  "either",
  "ff 0.13.0",
- "group 0.2.0",
  "hex",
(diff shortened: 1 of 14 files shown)


annenkov (approved):
Looks good.

PR #506, opened 2024-01-24
Extend the lifetime in generated code.

abizjak:
## Purpose

Extend the lifetime in the generated code so that if the return value of the call has references that refer to host the generated code still compiles.

In particular this comes up in advanced cases where we have an entrypoint of a contract of the form
```rust
fn receive(...) -> Result<impl Serial, SomeError>
``` 
where the return value `impl Serial` would typically refer to either the host (via state if it is retrieving the state) or to the parameter.

This was discovered while prototyping https://github.com/Concordium/concordium-rust-smart-contracts/pull/386

## Checklist

      hard-to-understand areas.


smart-contracts/contracts-common/concordium-contracts-common-derive/src/attribute.rs
```diff
@@ -659,9 +662,11 @@ pub fn receive_worker(attr: TokenStream, item: TokenStream) -> syn::Result<Token
                 if let Ok(state) = DeserialWithState::deserial_with_state(&state_api, &mut state_api.lookup_entry(&[]).unwrap_abort()) {
                     let mut state_builder = StateBuilder::open(state_api);
                     let mut host = ExternHost { state, state_builder };
-                    match #fn_name(&ctx, #host_ref, #(#fn_optional_args, )*) {
+                    let result = #fn_name(&ctx, #host_ref, #(#fn_optional_args, )*);
```
limemloh, line 665: **Suggestion (non-blocking):** Add a comment or some test case, ensuring we don't inline this again in the future.

limemloh (approved):

smart-contracts/contracts-common/concordium-contracts-common-derive/src/attribute.rs
abizjak: The problem is that the test for this would have to be in concordium-rust-smart-contracts repository which is not ideal.

I'll add a comment though.

abizjak: Added.

PR #507, opened 2024-01-24
Lts 22.9

td202:
## Purpose

Addresses https://github.com/Concordium/concordium-node/issues/1093

This upgrades the stack resolver to LTS-22.5, and the corresponding GHC version to 9.6.3.

## Changes

- Resolver is update to lts-22.5 in stack.yaml files.
- The package proto-lens-setup and its dependencies proto-lens-protoc and ghc-source-gen are not included in the LTS distribution, so specific versions from Hackage are pinned in the stack.yaml files.
- TypeOperators is added to the default extensions. In future, this will be needed for using ~, and currently will generate a warning if not enabled.
- Some imports are changed due to changes in library structure (notably, Control.Monad is explicitly imported, whereas it previously was implicitly imported through other monad modules).
- Some code changes make types more explicit, e.g. in [type-changing record updates](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.6#type-changing-record-updates-involving-type-families).

## Checklist

      hard-to-understand areas.


stack.yaml
```diff
@@ -41,6 +41,11 @@ packages:
 extra-lib-dirs:
     - ./lib/
 
+extra-deps:
+  - proto-lens-setup-0.4.0.7@sha256:acca0b04e033ea0a017f809d91a7dbc942e025ec6bc275fa21647352722c74cc,3122
+  - proto-lens-protoc-0.8.0.0@sha256:a146ee8c9af9e445ab05651e688deb0ff849357d320657d6cea5be33cb54b960,2235
+  - ghc-source-gen-0.4.4.0@sha256:8499f23c5989c295f3b002ad92784ca5fed5260fd4891dc816f17d30c5ba9cd9,4236
```
abizjak, line 47: I think this needs to be propagated both to the client and wallet-proxy as well.

Have you checked that the client builds fine with the updated versions?

EmilLa1 (approved):

limemloh (approved):

stack.yaml
td202: It doesn't yet, but I'm working on it.

PR #508, opened 2024-01-24
Make VersionedSchemaError public in concordium-contracts-common

schwartz-concordium:
## Purpose

Make public such that `VersionedSchemaError` can be matched in applications which depend on `concordium-contracts-common`.

## Changes

Made `VersionedSchemaError ` public in `concordium-contracts-common`

## Checklist

      hard-to-understand areas.

### DIFF
--- a/smart-contracts/contracts-common/concordium-contracts-common/src/schema.rs
+++ b/smart-contracts/contracts-common/concordium-contracts-common/src/schema.rs
@@ -1020,7 +1020,7 @@ pub fn deserial_length(source: &mut impl Read, size_len: SizeLength) -> ParseRes
 
 // Versioned schema helpers
 #[cfg(feature = "derive-serde")]
-mod impls {
+pub mod impls {
     use crate::{from_bytes, schema::*};
     use base64::{engine::general_purpose, Engine};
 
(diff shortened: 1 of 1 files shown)


smart-contracts/contracts-common/concordium-contracts-common/src/schema.rs
```diff
@@ -1020,7 +1020,7 @@ pub fn deserial_length(source: &mut impl Read, size_len: SizeLength) -> ParseRes
 
 // Versioned schema helpers
 #[cfg(feature = "derive-serde")]
-mod impls {
+pub mod impls {
```
abizjak, line 1023: We don't want this. We don't want to expose the fact that there is an `impls` module publicly.
If specific functions are needed then they should be publicly exported from the library using `pub use`.

schwartz-concordium: Have added 
```
#[cfg(feature = "derive-serde")]
pub use impls::VersionedSchemaError;
```

in top of file `schema.rs`. Let me know if it is the wrong place to put it.

abizjak: It's fine.

limemloh (approved):

PR #509, opened 2024-02-06
Add function for creating web3Id proofs

shjortConcordium:
## Changes

Added function to create web3Id proofs to wallet-library.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/wallet_library/src/proofs.rs
+++ b/rust-src/wallet_library/src/proofs.rs
@@ -1,0 +1,134 @@
+use anyhow::Result;
+use concordium_base::{
+    common::base16_decode,
+    id::{constants, types::*},
+    web3id::{OwnedCommitmentInputs, Presentation, Request, Web3IdAttribute, Web3IdSigner},
+};
+use serde::Deserialize as SerdeDeserialize;
+
+/// Serializeable wrapper for a SecretKey.
+#[derive(SerdeDeserialize)]
+pub struct Web3SecretKey(#[serde(deserialize_with = "base16_decode")] ed25519_dalek::SecretKey);
+
+impl Web3IdSigner for Web3SecretKey {
+    fn id(&self) -> ed25519_dalek::VerifyingKey { self.0.id() }
+
+    fn sign(&self, msg: &impl AsRef<[u8]>) -> ed25519_dalek::Signature { self.0.sign(msg) }
+}
+
+/// The input used for creating a web3Id proof. It requires the request itself,
+/// the global context of the chain, and the secret commitments for the
+/// attributes.
+#[derive(SerdeDeserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Web3IdProofInput {
+    request:           Request<constants::ArCurve, Web3IdAttribute>,
+    global_context:    GlobalContext<constants::ArCurve>,
+    commitment_inputs:
+        Vec<OwnedCommitmentInputs<constants::ArCurve, Web3IdAttribute, Web3SecretKey>>,
+}
+
+/// creates a web3Id proof.
+pub fn create_web3_id_proof(
+    input: Web3IdProofInput,
+) -> Result<Presentation<constants::ArCurve, Web3IdAttribute>> {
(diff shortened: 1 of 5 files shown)


rust-src/wallet_library/src/proofs.rs
```diff
@@ -0,0 +1,134 @@
...
+use concordium_base::{
+    common::base16_decode,
+    id::{constants, types::*},
+    web3id::{OwnedCommitmentInputs, Presentation, Request, Web3IdAttribute, Web3IdSigner},
+};
+use serde::Deserialize as SerdeDeserialize;
+
+/// Serializeable wrapper for a SecretKey.
+#[derive(SerdeDeserialize)]
+pub struct Web3SecretKey(#[serde(deserialize_with = "base16_decode")] ed25519_dalek::SecretKey);
```
orhoj, line 11:
```suggestion
pub struct Web3IdSecretKey(#[serde(deserialize_with = "base16_decode")] ed25519_dalek::SecretKey);
```

rust-src/wallet_library/src/proofs.rs
```diff
@@ -0,0 +1,134 @@
+use anyhow::Result;
```
abizjak, line 1: Could you pleas not import this. It's ambiguous with Result in std.

Please use anyhow::Result

rust-src/wallet_library/src/proofs.rs
```diff
@@ -0,0 +1,134 @@
...
+
+/// creates a web3Id proof.
+pub fn create_web3_id_proof(
+    input: Web3IdProofInput,
+) -> Result<Presentation<constants::ArCurve, Web3IdAttribute>> {
+    let presentation = input.request.prove(
+        &input.global_context,
+        input.commitment_inputs.iter().map(Into::into),
+    );
+    presentation.map_err(|e| e.into())
```
abizjak, line 39: Why do we map the error here? What's wrong with keeping the error that was already there. You can print it just the same.

rust-src/wallet_library/src/proofs.rs
```diff
@@ -0,0 +1,134 @@
...
+        });
+
+        let data = serde_json::to_string_pretty(&proof?)?;
+        assert!(
+            serde_json::from_str::<Presentation<constants::ArCurve, Web3IdAttribute>>(&data)
+                .is_ok(),
+            "Cannot deserialize proof correctly."
+        );
+        Ok(())
+    }
```
abizjak, line 133: I feel that these tests don't really add value.

The functions are so trivial that this is just duplicating the tests that are there in the web3id modules.

rust-src/wallet_library/src/proofs.rs
shjortConcordium: Hmm, I did it to make it an anyhow::Error, but that is not necessary if I don't use the anyhow::Result.

rust-src/wallet_library/src/proofs.rs
shjortConcordium: Yes, I guess I kinda just added it to have a test :/
I have changed it to parse from json, to have an example json input :thinking: 
If you still think it's "worthless" then I'll remove it :upside_down_face:

abizjak: No, the parsing test is good.

rust-src/wallet_library/src/proofs.rs
```diff
@@ -0,0 +1,62 @@
...
+
+/// Creates a web3Id proof.
+pub fn create_web3_id_proof(
+    input: Web3IdProofInput,
+) -> Result<Presentation<constants::ArCurve, Web3IdAttribute>, ProofError> {
+    input.request.prove(
+        &input.global_context,
+        input.commitment_inputs.iter().map(Into::into),
+    )
+}
```
abizjak, line 40: So in Rust this would be usually better as a method.
```rust
impl Web3IdProofInput {
   pub fn create_proof(self) -> Result<...> {
     ...
   }
}`
```

abizjak (approved):

PR #510, opened 2024-02-07
Include the protocol version in block hash Merkle proofs.

td202:
## Purpose

Support the inclusion of the protocol version in the calculation of the block hash.

## Changes

- Include the protocol version in the block hash schema.
- Update the test accordingly.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/MerkleProofs.hs
+++ b/haskell-src/Concordium/MerkleProofs.hs
@@ -25,6 +25,7 @@ import Data.Word
 import Lens.Micro.Platform
 
 import qualified Concordium.Crypto.SHA256 as SHA256
+import Concordium.Types.ProtocolVersion
 import Control.Monad.Reader
 import qualified Data.FixedByteString as FBS
 import Data.Foldable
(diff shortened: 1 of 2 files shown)


haskell-src/Concordium/MerkleProofs.hs
```diff
@@ -596,7 +597,10 @@ blockSchema = runState builder emptySchemaBuilderState & _2 %~ _builderSchema
                   Hashed "result" resultHash
                 ]
         blockQuasiHash <- setFresh . Sequence $ [Hashed "meta" metaHash, Hashed "data" dataHash]
-        schemaAt blockHash ?= Sequence [Hashed "header" blockHeaderHash, Hashed "quasi" blockQuasiHash]
+        rootHash <-
+            setFresh . Sequence $
+                [Hashed "header" blockHeaderHash, Hashed "quasi" blockQuasiHash]
+        schemaAt blockHash ?= Sequence [LiteralBytes (S.encode P7), Hashed "root" rootHash]
```
EmilLa1, line 603: Maybe `blockSchema` should be named such that it reflects that it's only useful for P7? Could the function be parameterized by some "BlockHashVersion" that is determined from the protocol version?

EmilLa1 (approved):

abizjak (approved):

haskell-src/Concordium/MerkleProofs.hs
td202: I think in future the schema could/should support parsing for multiple protocol versions. At the moment, it's only well-defined for protocol version 7.

PR #513, opened 2024-02-21
Request satisfies restriction

shjortConcordium:
## Purpose

Add method to check whether statement violates general rules for the wallets. 

## Checklist

      hard-to-understand areas.


rust-src/wallet_library/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Added trait `WalletCheck` with method `satifies_wallet_restrictions`, and implemented it web3IdProofInput, request and statements.
```
abizjak, line 7: "and implemented it ..."

Something is missing. Also quote `Web3IdProofInput` and capitalize correctly please.

rust-src/wallet_library/Cargo.toml
```diff
@@ -13,6 +13,8 @@ serde_json = "1.0"
 thiserror = "1.0"
 either = "1.6"
 ed25519-dalek = "2.0"
+chrono = "0.4.24" # the patch version to match base
```
abizjak, line 16: Why do you need the patch version here? Base already sets it, so this won't really do anything.

rust-src/wallet_library/Cargo.toml
```diff
@@ -13,6 +13,8 @@ serde_json = "1.0"
...
+rust_iso3166 = "0.1.11"
```
abizjak, line 17: Do we need patch version?

rust-src/wallet_library/src/proofs.rs
```diff
@@ -7,6 +7,8 @@ use concordium_base::{
 };
 use serde::Deserialize as SerdeDeserialize;
```
abizjak, line 9:
```suggestion
```

rust-src/wallet_library/src/statement.rs
```diff
@@ -0,0 +1,274 @@
+use concordium_base::{
+    curve_arithmetic::Curve,
+    id::{id_proof_types::AtomicStatement, types::*},
+    web3id::{CredentialStatement, Request},
+};
+
+const IDENTITY_RANGE_TAGS: &[AttributeTag] = &[AttributeTag(3), AttributeTag(9), AttributeTag(10)]; // dob, idDocIssuedAt, idDocExpiresAt
+const IDENTITY_SET_TAGS: &[AttributeTag] = &[
```
abizjak, line 8: Document. The names are hard to parse and understand without a bit of context.

abizjak: Are these deliberately not public?

rust-src/wallet_library/src/statement.rs
```diff
@@ -0,0 +1,274 @@
...
+    AttributeTag(4),
+    AttributeTag(5),
+    AttributeTag(6),
+    AttributeTag(8),
+]; // Country of residence, Nationality, IdDocType, IdDocIssuer
+
+fn is_iso8601(date: &str) -> bool {
+    println!("{}", date);
```
abizjak, line 16: What purpose does this serve?

rust-src/wallet_library/src/statement.rs
```diff
@@ -0,0 +1,274 @@
...
+    chrono::NaiveDate::parse_from_str(date, "%Y%m%d").is_ok()
+}
+
+fn is_iso3166_alpha_2(code: &str) -> bool { rust_iso3166::from_alpha2(code).is_some() }
+
+fn is_iso3166_2(code: &str) -> bool { rust_iso3166::iso3166_2::from_code(code).is_some() }
+
+pub trait WalletCheck {
+    fn satisfies_wallet_restrictions(&self) -> bool;
+}
```
abizjak, line 26: Please document what this is for.

abizjak: Also why is this called "WalletCheck" when it's implemented for requests and statements.

It seems there is no "wallet" parameter anywhere. And how will we handle that some wallets, such as the browser wallet, have looser restrictions?

rust-src/wallet_library/src/statement.rs
```diff
@@ -0,0 +1,274 @@
...
+            } => {
+                if statement.is_empty() {
+                    return false;
+                }
+                let mut used_tags = Vec::<AttributeTag>::new();
+                for atomic_statement in statement {
+                    if used_tags.contains(&atomic_statement.attribute()) {
+                        return false;
+                    }
+                    used_tags.push(atomic_statement.attribute());
```
abizjak, line 54: Please use a set here. Don't be [accidentally quadratic](https://accidentallyquadratic.tumblr.com/)

rust-src/wallet_library/src/statement.rs
```diff
@@ -0,0 +1,274 @@
...
+                    }
+                }
+            }
+            CredentialStatement::Web3Id {
+                ty: _,
+                network: _,
+                contract: _,
+                credential: _,
+                statement: _,
+            } => todo!(),
```
abizjak, line 67: todo?

rust-src/wallet_library/src/statement.rs
```diff
@@ -0,0 +1,274 @@
...
+            3 | 9 | 10 => is_iso8601(&value.to_string()),
+            _ => true,
+        };
+
+        match self {
+            AtomicStatement::RevealAttribute { statement: _ } => true,
+            AtomicStatement::AttributeInRange { statement } => {
+                check_attribute_value(&statement.lower)
+                    && check_attribute_value(&statement.upper)
+                    && statement.lower < statement.upper
```
abizjak, line 94: WHy not <=?

rust-src/wallet_library/src/statement.rs
abizjak, line 26: I would have expected a trait more in line with
```rust
pub trait AcceptableRequest {
    type Error;
    fn acceptable_request(&self, wallet: &WalletConfig) -> Result<(), Error>;
}
```

and you'd have an enum WalletConfig.

The `Error` would be an enum as well that could be used to display something meaningful to either the user or dapp.

I'm imagining the scenario where a dapp asks for a web3id proof of the mobile wallet. What will the experience be?

rust-src/wallet_library/Cargo.toml
shjortConcordium: Ahh, I thought it might use the latest here if I didn't specify. I've removed the patch.

rust-src/wallet_library/Cargo.toml
shjortConcordium: Nope, I've removed it.

rust-src/wallet_library/src/statement.rs
shjortConcordium: None, should have been, and now is, removed.

rust-src/wallet_library/src/statement.rs
shjortConcordium: They are public now.

rust-src/wallet_library/src/statement.rs
shjortConcordium: Yeah, I've have changed it to use Result, and added a `WalletConfig` that allows the user to specify tags that are allowed for range and set, and to provide a function for "custom" attribute verification.

rust-src/wallet_library/src/statement.rs
shjortConcordium: Hmm I thought that lower = upper doesn't make sense, when upper bound is exclusive? :thinking: 
We don't allow it in javascript. :upside_down_face:

rust-src/wallet_library/src/statement.rs
shjortConcordium: It didn't make sense without having the schema. I added handling of it with the WalletConfig now.

rust-src/wallet_library/src/statement.rs
shjortConcordium: I have renamed it as suggested (And split it into two traits)
(And added a WalletConfig parameter, as suggested)

rust-src/wallet_library/src/statement.rs
abizjak: Yeah you are right, upper bound is exclusive.

rust-src/wallet_library/src/statement.rs
```diff
@@ -0,0 +1,375 @@
...
+    web3id::{CredentialStatement, Request},
+};
+use std::{
+    collections::{BTreeMap, HashSet},
+    fmt::Display,
+    hash::Hash,
+    marker::PhantomData,
+};
+
+pub trait AcceptableRequest<C: Curve, AttributeType: Attribute<C::Scalar>> {
```
abizjak, line 15: Please document the purpose of the trait

rust-src/wallet_library/src/statement.rs
```diff
@@ -0,0 +1,375 @@
...
+    /// A function to check attributes using custom behaviour.
+    pub attribute_check: AttributeCheck<'a, TagType, AttributeType>,
+    pub _marker:         PhantomData<C>,
+}
+
+pub trait AttributeTagType: Serialize + Ord + Hash + Display {}
+impl<T: Serialize + Ord + Hash + Display> AttributeTagType for T {}
+
+type AttributeCheck<'a, TagType, AttributeType> =
+    Box<dyn Fn(&TagType, &AttributeType) -> AcceptableRequestResult + 'a>;
```
abizjak, line 69: Why this complicated lifetime when it always seems to be `'static` at the end?

rust-src/wallet_library/src/statement.rs
```diff
@@ -0,0 +1,375 @@
...
+        }
+        // checks based on wallet config
+        if let Some(rules) = config_rules {
+            let check = &rules.attribute_check;
+            match self {
+                AtomicStatement::RevealAttribute { statement: _ } => (),
+                AtomicStatement::AttributeInRange { statement } => {
+                    let tag = &statement.attribute_tag;
+                    if !rules.range_tags.contains(tag) {
+                        return Err(RequestCheckError::IllegalRangeTag(format!("{}", tag)));
```
abizjak, line 180:
```suggestion
                        return Err(RequestCheckError::IllegalRangeTag(tag.to_string()));
```
I'm surprised clippy does not complain :/

rust-src/wallet_library/src/statement.rs
```diff
@@ -0,0 +1,375 @@
...
+    #[error("Credential statement may not reuse attribute tags across atomic statements")]
+    DuplicateTag,
+    #[error("Web3Id statement support have not been added yet")]
+    Web3IdStatementNotSupported,
+    #[error("`{0}`")]
+    InvalidValue(String),
+    #[error("Range statement min must be less than max")]
+    RangeMinMaxError,
+    #[error("The tag `{0}` is not allowed to be used for range statements")]
+    IllegalRangeTag(String),
```
abizjak, line 84: I don't quite get why this is a String since this loses information. Why not just put the tag here directly?

rust-src/wallet_library/src/statement.rs
```diff
@@ -0,0 +1,375 @@
...
+                }
+                Ok(())
+            }
+            CredentialStatement::Web3Id {
+                ty: _,
+                network: _,
+                contract: _,
+                credential: _,
+                statement: _,
+            } => Err(RequestCheckError::Web3IdStatementNotSupported),
```
abizjak, line 138: I'm a bit confused with this. There is a web3_rules in the walletconfig but you don't use them here at all.

shjortConcordium: Yeah, I meant to implement this (and remove that error). I've added it.

rust-src/wallet_library/src/statement.rs
shjortConcordium: The tag can be a string (for web3Id), so this would need to be generic, so I thought making this a string was preferable.

rust-src/wallet_library/src/statement.rs
shjortConcordium: Because formatting is always cool :sunglasses: (Fixed it)

rust-src/wallet_library/src/statement.rs
shjortConcordium: For the default config it is just static, but when others use it (and add web3Id rules) I think it's better to have this. :thinking:

rust-src/wallet_library/src/statement.rs
shjortConcordium: I've tried to do so. :upside_down_face:

rust-src/wallet_library/src/default_wallet_config.rs
```diff
@@ -0,0 +1,84 @@
...
+pub const ALLOWED_IDENTITY_SET_TAGS: [AttributeTag; 4] = [
+    AttributeTag(4),
+    AttributeTag(5),
+    AttributeTag(6),
+    AttributeTag(8),
+];
+
+/// Returns the `WalletConfig` that is used by our wallet implementations for
+/// identity proofs. Note that it does not contain any rules for web3
+/// statements.
```
abizjak, line 73: I suppose what you mean with the last sentence is that this config does not allow statements about web3id credentials. If that's the case, state it like so.

rust-src/wallet_library/src/default_wallet_config.rs
```diff
@@ -0,0 +1,84 @@
...
+pub fn get_default_wallet_config() -> WalletConfig<'static, constants::ArCurve, AttributeKind> {
```
abizjak, line 74:
```suggestion
pub fn default_wallet_config() -> WalletConfig<'static, constants::ArCurve, AttributeKind> {
```

rust-src/wallet_library/src/statement.rs
abizjak: Ok. I'm generally of the opinion that complexity should be there if it's needed, and abstractions should be introduced when they are needed, not when they are thought that they might be needed.

But OK :)

rust-src/wallet_library/src/statement.rs
```diff
@@ -0,0 +1,403 @@
...
+                    atomic_statement.acceptable_atomic_statement(config.web3_rules.get(contract))?
+                }
+                Ok(())
+            }
+        }
+    }
+}
+
+// Note that this is an implementation only for the account statement, the tag
+// is AttributeTag.
```
abizjak, line 168: This does not appear accurate.

rust-src/wallet_library/src/statement.rs
```diff
@@ -0,0 +1,403 @@
...
+        > = AtomicStatement::AttributeInRange {
+            statement: AttributeInRangeStatement {
+                attribute_tag: AttributeTag(3),
+                lower:         constants::AttributeKind::from(5),
+                upper:         constants::AttributeKind::from(10),
+                _phantom:      PhantomData::<constants::ArCurve>,
+            },
+        };
+
+        let statement: CredentialStatement<constants::ArCurve, constants::AttributeKind> = CredentialStatement::Account { network: Network::Testnet, cred_id: CredentialRegistrationID::from_str("8a3a87f3f38a7a507d1e85dc02a92b8bcaa859f5cf56accb3c1bc7c40e1789b4933875a38dd4c0646ca3e940a02c42d8")?, statement: vec![statement1, statement2]};
```
abizjak, line 395: That's a long line. Is there no formatting for this crate?

abizjak (approved):
If formatting is not set up for this crate yet, please add it in a separate PR.

rust-src/wallet_library/src/default_wallet_config.rs
shjortConcordium: I mean that it has no specific checks, so for web3Id statements it will only check "basic rules", but it will still allow them. I've made that more clear.

rust-src/wallet_library/src/statement.rs
shjortConcordium: When I run fmt, it doesn't care :thinking:
I thought it uses the rules that are specified in the root folder.
I don't think it is checked on the CI, so I'll add that in another PR.

PR #520, opened 2024-03-11
Support state-related host calls in `TestHost`

limemloh:
## Purpose

Extend the `TestHost` to allow unit tests to use the state-related host function calls.

Needed to do allow unit tests for the new `StateBTreeMap` and `StateBTreeSet` introduced in 
https://github.com/Concordium/concordium-rust-smart-contracts/pull/405 using cargo-concordium.

Related cargo-concordium PR: https://github.com/Concordium/concordium-smart-contract-tools/pull/158

## Changes

- Define a new constructor for `TestHost` taking the instance state to make available.
- `TestHost` support host function calls related to state.
- `run_module_tests` provide an empty instance state for every test case.

### DIFF
--- a/smart-contracts/wasm-chain-integration/src/utils.rs
+++ b/smart-contracts/wasm-chain-integration/src/utils.rs
@@ -1,7 +1,10 @@
 //! Various utilities for testing and extraction of schemas and build
 //! information.
 
-use crate::{v1::EmittedDebugStatement, ExecResult};
+use crate::{
+    v1::{host, trie, EmittedDebugStatement, InstanceState},
+    ExecResult,
+};
 use anyhow::{anyhow, bail, ensure, Context};
 pub use concordium_contracts_common::WasmVersion;
 use concordium_contracts_common::{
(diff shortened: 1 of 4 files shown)


smart-contracts/wasm-chain-integration/src/utils.rs
```diff
@@ -163,6 +189,7 @@ impl<R: RngCore> machine::Host<ArtifactNamedImport> for TestHost<R> {
         memory: &mut Vec<u8>,
         stack: &mut machine::RuntimeStack,
     ) -> machine::RunResult<Option<NoInterrupt>> {
+        let mut energy = crate::InterpreterEnergy::new(u64::MAX);
```
abizjak, line 192: This at least warrants some explanation.

smart-contracts/wasm-chain-integration/src/utils.rs
```diff
@@ -39,39 +42,60 @@ impl<I> machine::Host<I> for TrapHost {
...
-pub struct TestHost<R> {
+pub struct TestHost<'a, R, BackingStore> {
     /// A RNG for randomised testing.
     rng:              Option<R>,
     /// A flag set to `true` if the RNG was used.
     rng_used:         bool,
     /// Debug statements in the order they were emitted.
     pub debug_events: Vec<EmittedDebugStatement>,
+    /// In-memory instance state used for state-related host calls.
+    state:            Option<InstanceState<'a, BackingStore>>,
```
abizjak, line 53: Why exactly is this an option?

Can't we always just construct an empty one?

limemloh: I did try making it construct this empty instance state as part of constructing `TestHost`, but I ran into issues with it needing to be self-referential, but I'll give it another attempt.

limemloh: The issue is that `TestHost` needs to own the backing store, the `MutableState` and then the runtime state (`InstanceState`) which holds a reference to both of the before mentioned. It might be possible to get it working, but I think it involves using unsafe.

abizjak: Hmm, I could be wrong, but it seems to me that we never really need to construct the testhost without a state.

We seem to use it in `uninitialized` as a "import export" validator, the same as the the one constructed with ::new. So could we not have these two separate, and then remove the Option here seeing it's only constructed with the actual state ever?

limemloh: I have updated the PR with this last suggestion

abizjak (approved):

smart-contracts/wasm-chain-integration/src/utils.rs
abizjak: Looks good.

abizjak: Needs some fixes in benchmarks though.

limemloh: Yes, I'll use `TrapHost` instead of `TestHost` for these.

PR #527, opened 2024-04-17
Bumb IdissLib to NET6.0 and support Mac and Linux

limemloh:
## Purpose

Bumb IdissLib to NET6.0 and support Mac and Linux

## Changes

- IdissLib now builds the `idiss` Rust library and includes the binaries, removing the need for manually copying a shared library.
- Support Linux and Mac for IdissLib, by conditionally copying `idiss` shared-library based on platform.
- Bump to NET6.0, since NET5.0 no longer have official support anymore, and therefore causes warnings.
- Update Readme

Additional changes:
- Update `concordium-base` and `haskell-src` with company attributes.


idiss-csharp/IdissExample/IdissExample.csproj
```diff
@@ -6,7 +6,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
abizjak, line 9: This sort of change is not ideal without checking with the consumers first.

This idiss library was developed for DTS specifically first so we should make sure ideally that they can still use it.

If NET 5 is no longer supported I assume that is the case.

abizjak (approved):

eb-concordium (approved):
Looks good to me.

idiss-csharp/IdissExample/IdissExample.csproj
limemloh: Yes, that makes sense.
I have brought this to their attention and there should be no issue.

PR #529, opened 2024-05-02
Add SignedTransaction type

DOBEN:
## Purpose

Addresses https://github.com/Concordium/concordium-client/issues/296
Related https://github.com/Concordium/concordium-client/pull/300

## Changes

Implement `toJSON`/`fromJSON` instances of the `Payload` types.


haskell-src/Concordium/Crypto/EncryptedTransfers.hs
```diff
@@ -297,6 +297,9 @@ instance FromJSON EncryptedAmountTransferData where
         eatdProof <- v .: "proof"
         return EncryptedAmountTransferData{..}
 
+instance ToJSON EncryptedAmountTransferData where
+    toJSON = error "Will not be implemented since feature will be deprecated soon"
```
limemloh, line 301: We don't want to update this text again, so I think we should just say that it is already deprecated.

haskell-src/Concordium/Types.hs
```diff
@@ -711,7 +711,23 @@ instance Show Address where
 
 -- | Time in seconds since the unix epoch
 newtype TransactionTime = TransactionTime {ttsSeconds :: Word64}
-    deriving (Show, Read, Eq, Num, Ord, FromJSON, ToJSON, Real, Enum, Integral) via Word64
+    deriving (Show, Read, Eq, Num, Ord, Real, Enum, Integral) via Word64
+
+-- Implement `ToJSON` instance for `TransactionTime`.
+instance ToJSON TransactionTime where
+    toJSON (TransactionTime seconds) =
+        String $ T.pack $ formatTime defaultTimeLocale "%FT%T%QZ+00:00" (posixSecondsToUTCTime (fromIntegral seconds))
```
limemloh, line 719: **Suggestion:** Avoid formatting dates manually, this is not the first time we depend on this format, so I expect we have something already.
Maybe this is good enough https://hackage.haskell.org/package/time-1.14/docs/Data-Time-Format-ISO8601.html

haskell-src/Concordium/Types.hs
```diff
@@ -711,7 +711,23 @@ instance Show Address where
 
 -- | Time in seconds since the unix epoch
 newtype TransactionTime = TransactionTime {ttsSeconds :: Word64}
-    deriving (Show, Read, Eq, Num, Ord, FromJSON, ToJSON, Real, Enum, Integral) via Word64
+    deriving (Show, Read, Eq, Num, Ord, Real, Enum, Integral) via Word64
```
limemloh, line 714: **Potential issue:** This library is used by both the `concordium-node` and the `concordium-wallet-proxy`, so we need to ensure this is not breaking anything. 
Don't think anything in the node depends on the JSON format anymore, but I imagine this is exposed by the wallet-proxy.

stack.yaml
```diff
@@ -41,6 +41,8 @@ packages:
 extra-lib-dirs:
     - ./lib/
 
+system-ghc: true
```
limemloh, line 44: From my understanding, whether this is needed depends on your setup, so I don't think we should have this as default behavior for everyone.
It seems to work fine for me, so maybe we should figure out the difference of our setup instead

haskell-src/Concordium/Wasm.hs
```diff
@@ -311,6 +334,20 @@ data WasmModule
     | WasmModuleV1 (WasmModuleV V1)
     deriving (Eq, Show)
 
+-- Implement `FromJSON` and `ToJSON` instances for `WasmModule`.
+$( deriveJSON
```
limemloh, line 338: How does the resulting JSON look for this type, I find it hard to derive in my head? :sweat_smile:

DOBEN:
```
{
    "energy": 200000,
    "expiryTime": "2024-05-14T14:23:18Z+00:00",
    "nonce": 38,
    "payload": {
        "mod": {
            "content": {
                "source": "0061736d010000000...................................616e736665724572726f720200"
            },
            "version": "wasmModuleV1"
        },
        "transactionType": "deployModule"
    },
    "signature": {
        "0": {
            "0": "a4ad6d482b1744c81c6223478c50ad2053d5add62c415948ef2f000275b7446eda183373e8a113dac8781b7d346a088873080957c3c5b9715a7b56929a49bf07"
        }
    },
    "signer": "4jxvYasaPncfmCFCLZCvuL5cZuvR5HAQezCHZH7ZA7AGsRYpix",
    "version": 1
}
```

haskell-src/Concordium/Crypto/EncryptedTransfers.hs
td202: I really think we shouldn't have instances like this (i.e. that simply error). `error` should be used with care, and only in places where we can be pretty sure it won't happen, but if it does it's OK to crash.

```suggestion
    toJSON EncryptedAmountTransferData{..} =
        object
            [ "remainingAmount" .= eatdRemainingAmount,
              "transferAmount" .= eatdTransferAmount,
              "index" .= eatdIndex,
              "proof" .= eatdProof
            ]
```

haskell-src/Concordium/Crypto/EncryptedTransfers.hs
```diff
@@ -491,6 +494,9 @@ instance FromJSON SecToPubAmountTransferData where
         stpatdProof <- v .: "proof"
         return SecToPubAmountTransferData{..}
 
+instance ToJSON SecToPubAmountTransferData where
+    toJSON = error "Will not be implemented since feature will be deprecated soon"
```
td202, line 498:
```suggestion
instance ToJSON SecToPubAmountTransferData where
    toJSON SecToPubAmountTransferData{..} =
        object
            [ "remainingAmount" .= stpatdRemainingAmount,
              "transferAmount" .= stpatdTransferAmount,
              "index" .= stpatdIndex,
              "proof" .= stpatdProof
            ]
```

haskell-src/Concordium/ID/Types.hs
```diff
@@ -797,6 +798,10 @@ instance FromJSON CredentialDeploymentInformation where
                         }
             Left _ -> fail "\"proofs\" is not a valid base16 string."
 
+-- Implement `ToJSON` instance for `CredentialDeploymentInformation`.
+instance ToJSON CredentialDeploymentInformation where
+    toJSON = error "Not yet implemented"
```
td202, line 803:
```suggestion
instance ToJSON CredentialDeploymentInformation where
    toJSON CredentialDeploymentInformation{..} =
        object $
            ("proofs" .= cdiProofs)
                : credentialDeploymentValuesList cdiValues
```

haskell-src/Concordium/Types/Transactions.hs
```diff
@@ -139,6 +141,60 @@ instance S.Serialize TransactionSignature where
...
+--  The chosen representation is the minimal necessary data needed to construct the
+-- 'TransactionSignHash' which is the value that is signed by the signer. The
+-- 'TransactionSignHash' and 'payloadSize' should be re-computed when processing a
+-- 'SignedTransaction' (e.g. when adding signatures or sending the transaction on-chain).
+--
+-- The representation has a `version` field.
+data SignedTransaction = SignedTransaction
+    { -- | A version to distinguish between future formats of signed/partially-signed transactions.
+      -- The initial version is 1 and will be incremented for every new format.
+      stVersion :: !Int,
```
td202, line 168: This is an implementation detail of the JSON encoding, so it shouldn't actually be included in the data structure. I recommend defining the `ToJSON` and `FromJSON` instances explicitly, and including the version handling as part of that.

haskell-src/Concordium/Types/Transactions.hs
```diff
@@ -139,6 +141,60 @@ instance S.Serialize TransactionSignature where
...
+-- representation ready to be printed into a JSON file. This file can be shared among
+-- different tools of the Concordium ecosystem for adding additional signatures.
+--
+--  The chosen representation is the minimal necessary data needed to construct the
+-- 'TransactionSignHash' which is the value that is signed by the signer. The
+-- 'TransactionSignHash' and 'payloadSize' should be re-computed when processing a
+-- 'SignedTransaction' (e.g. when adding signatures or sending the transaction on-chain).
+--
+-- The representation has a `version` field.
+data SignedTransaction = SignedTransaction
```
td202, line 165: I am not sure why this should be here (in base) rather than in the client.

haskell-src/Concordium/Types.hs
```diff
@@ -996,6 +1012,7 @@ instance S.Serialize PayloadSize where
 -- | Serialized payload of the transaction
 newtype EncodedPayload = EncodedPayload {_spayload :: BSS.ShortByteString}
     deriving (Eq, Show)
+    deriving (AE.ToJSON, AE.FromJSON) via BSH.ByteStringHex
```
td202, line 1015: What is this needed for?

haskell-src/Concordium/Types.hs
```diff
@@ -711,7 +711,23 @@ instance Show Address where
...
+-- Implement `FromJSON` instance for `TransactionTime`.
+instance FromJSON TransactionTime where
+    parseJSON (String v) =
+        case parseTimeM True defaultTimeLocale "%FT%T%QZ+00:00" (T.unpack v) of
+            Just time -> return $ TransactionTime (convertToWord64 $ utcTimeToPOSIXSeconds time)
+            Nothing -> fail "Invalid RFC 3339 timestamp format for TransactionTime (expect %FT%T%QZ+00:00)"
+    parseJSON _ = fail "Expect JSON string for TransactionTime"
+
+convertToWord64 :: NominalDiffTime -> Word64
+convertToWord64 = floor . nominalDiffTimeToSeconds
```
td202, line 730: This should not be a top-level definition.

haskell-src/Concordium/Crypto/EncryptedTransfers.hs
DOBEN: we don't want to spend time on implementing this feature since the decision was made to deprecate (even remove) the whole code related to shielded transactions. In that case, I think it is ok to throw an `error` but coping the `toJSON` and not spending time on testing is also fine with me.

DOBEN: we don't want to spend time on implementing this feature since the decision was made to deprecate (even remove) the whole code related to shielded transactions. In that case, I think it is ok to throw an error but coping the toJSON and not spending time on testing is also fine with me.

haskell-src/Concordium/Types.hs
DOBEN: I kept the old `fromJSON`/`toJSON` implementation now.

stack.yaml
DOBEN: The language server crashes with the error `GHC ABIs don't match!`. Adding the above line has fixed it:

https://www.reddit.com/r/haskell/comments/uk5wg7/comment/i7n990t/?utm_source=reddit&utm_medium=web2x&context=3

haskell-src/Concordium/Types.hs
DOBEN: Initially, the encoded payload was part of a partially signed transaction in the JSON file. Now we have changed it to use the decoded (human-readable) payload instead. So you are right that this line is currently not needed. I could remove it but since it was already implemented, I just kept it in case we need it at some point.

haskell-src/Concordium/Types/Transactions.hs
DOBEN: The main reason to move it to base was that this type should be used across Concordium components (meaning the multi-sig JSON format should be compatible). E.g. maybe we add a service to the `wallet-proxy` for wallets to display pending partially-signed transactions at some point and this type can be re-used.

eb-concordium (approved):
Looks good modulo the remaining unresolved comments :)

haskell-src/Concordium/Wasm.hs
limemloh: Not sure how much energy we should put into the readability of the JSON, but it would be nice if the expiryTime was readable.
I also think it would be nice to clean up the JSON representation, maybe:
```
"payload": {
  "module": {
    "version": "v1",
    "source": "..."
  }
  "transactionType": "deployModule"
}
```

stack.yaml
limemloh: I will help you fix this, but this workaround could cause others setup to break, so please remove this again.

DOBEN: Some further digging got me to these workarounds:

https://github.com/haskell/vscode-haskell?tab=readme-ov-file#ghc-abis-dont-match

One workaround is to fix it in your local setup to force stack to always use the global ghc version.
https://www.haskell.org/ghcup/guide/#strategy-2-system-ghc-works-on-all-stack-versions

haskell-src/Concordium/Types/Transactions.hs
td202: `concordium-client` is a dependency of the `wallet-proxy` anyway, so this is not a very strong reason. (`concordium-client` is sort of the Haskell SDK + CLI.)

DOBEN: `Concordium-client` should be a CLI and not be treated as a `Haskell SDK` for other components to depend on it.

DOBEN: ok, that makes sense, if we argue that `base` should be kept minimal and only include a necessary dependency for the node especially since the `base` is at the end of some of our sub-modules link chains.
That `concordium-client` is treated as a Haskell SDK is something we should fix at some point so :).

haskell-src/Concordium/Types/Execution.hs
```diff
@@ -404,6 +429,139 @@ instance S.Serialize TransactionType where
...
+    toJSON AddBaker{..} = AE.object ["electionVerifyKey" AE..= abElectionVerifyKey, "signatureVerifyKey" AE..= abSignatureVerifyKey, "aggregationVerifyKey" AE..= abAggregationVerifyKey, "proofSig" AE..= abProofSig, "proofElection" AE..= abProofElection, "proofAggregation" AE..= abProofAggregation, "bakingStake" AE..= abBakingStake, "restakeEarnings" AE..= abRestakeEarnings, "transactionType" AE..= AE.String "addBaker"]
+    toJSON RemoveBaker = AE.object ["transactionType" AE..= AE.String "removeBaker"]
+    toJSON UpdateBakerStake{..} = AE.object ["stake" AE..= ubsStake, "transactionType" AE..= AE.String "updateBakerStake"]
+    toJSON UpdateBakerRestakeEarnings{..} = AE.object ["restakeEarnings" AE..= ubreRestakeEarnings, "transactionType" AE..= AE.String "updateBakerRestakeEarnings"]
+    toJSON UpdateBakerKeys{..} = AE.object ["electionVerifyKey" AE..= ubkElectionVerifyKey, "signatureVerifyKey" AE..= ubkSignatureVerifyKey, "aggregationVerifyKey" AE..= ubkAggregationVerifyKey, "proofSig" AE..= ubkProofSig, "proofElection" AE..= ubkProofElection, "proofAggregation" AE..= ubkProofAggregation, "transactionType" AE..= AE.String "updateBakerKeys"]
```
td202, line 455: Please format on multiple lines for readability.

haskell-src/Concordium/Wasm.hs
```diff
@@ -290,6 +304,15 @@ moduleSourceLength = fromIntegral . BS.length . moduleSource
...
 
+-- Implement `FromJSON` and `ToJSON` instances for `WasmModuleV`.
+$( deriveJSON
+    defaultOptions
+        { AE.constructorTagModifier = firstLower,
+          AE.fieldLabelModifier = firstLower . dropWhile isLower
+        }
+    ''WasmModuleV
+ )
+
```
td202, line 315: I think giving the instance explicitly would be preferable, because it's short and it will be clearer. Also, I would suggest including the version in the serialization to be in line with the Serialize instance:

```suggestion
instance (IsWasmVersion v) => AE.ToJSON (WasmModuleV v) where
    toJSON (WasmModuleV ws) =
        AE.object
            [ "version" AE..= wasmVersionToWord (demoteWasmVersion (getWasmVersion @v)),
              "source" AE..= moduleSource ws
            ]

instance (IsWasmVersion v) => AE.FromJSON (WasmModuleV v) where
    parseJSON = AE.withObject "WasmModuleV" $ \obj -> do
        version <- obj AE..: "version"
        if wordToWasmVersion version == Just (demoteWasmVersion (getWasmVersion @v))
            then WasmModuleV . ModuleSource <$> obj AE..: "source"
            else fail $ "Expecting a " ++ show (getWasmVersion @v) ++ " module."

```

haskell-src/Concordium/Wasm.hs
```diff
@@ -311,6 +334,21 @@ data WasmModule
...
+-- Custom implementation of FromJSON for WasmModule
+instance AE.FromJSON WasmModule where
+    parseJSON = AE.withObject "WasmModule" $ \obj -> do
+        version <- obj AE..: "version"
+        content <- obj AE..: "content"
+        case version :: Int of
+            0 -> WasmModuleV0 <$> AE.parseJSON content
+            1 -> WasmModuleV1 <$> AE.parseJSON content
+            _ -> fail "Invalid version number"
+
```
td202, line 351: I think the format `{"version"=0, "content"={"source"="..."}}` is a bit redundant, and something like `{"version"=0, "source"="..."}` would be cleaner. Something like:

```suggestion
instance AE.ToJSON WasmModule where
    toJSON = \case
        WasmModuleV0 wm -> AE.toJSON wm
        WasmModuleV1 wm -> AE.toJSON wm

instance AE.FromJSON WasmModule where
    parseJSON = AE.withObject "WasmModule" $ \obj -> do
        version <- obj AE..: "version"
        case wordToWasmVersion version of
            Just V0 -> WasmModuleV0 . WasmModuleV <$> obj AE..: "source"
            Just V1 -> WasmModuleV1 . WasmModuleV <$> obj AE..: "source"
            Nothing -> fail $ "Unsupported Wasm version " ++ show version
```

td202 (approved):
Looks good. I would suggest adding QuickCheck tests for the JSON serializations (in the same manner as for the binary serializations) as these can potentially catch more bugs.

td202 (approved):

DOBEN:
For now, I created an issue to improve on the tests:

https://github.com/Concordium/concordium-base/issues/535

PR #537, opened 2024-06-20
Deprecate encrypted transfers

Victor-N-Suadicani:
## Purpose

Encrypted transfers will no longer be possible in protocol version 7. The functionality must still be kept for backwards compatibility when looking at old blocks on the chain however, so we opt to deprecate the functionality indefinitely.

## Changes

Functions and enum variants related to encrypted transfers are marked with the `#[deprecated]` attribute.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/identity-provider-service/Cargo.lock
+++ b/identity-provider-service/Cargo.lock
@@ -472,7 +472,7 @@ dependencies = [
 
 [[package]]
 name = "concordium_base"
-version = "5.0.1"
+version = "5.0.0"
 dependencies = [
  "anyhow",
  "ark-bls12-381",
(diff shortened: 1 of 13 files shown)


td202:
I feel if we're doing this, we should probably also deprecate `AddBaker`, `RemoveBaker`, `UpdateBakerStake`, `UpdateBakerRestakeEarnings` and `UpdateBakerKeys`, which were removed in protocol version 4.

Victor-N-Suadicani:
Done :)

limemloh (approved):
Looks good to me :+1:

PR #549, opened 2024-08-19
Suspend in ConfigureBaker

drsk0:
## Purpose

This implements the new suspend/resume field and events for the `ConfigureBaker` transaction in base.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/haskell-src/Concordium/Types/Accounts.hs
+++ b/haskell-src/Concordium/Types/Accounts.hs
@@ -732,7 +732,10 @@ data AccountInfo = AccountInfo
       aiAccountCooldowns :: ![Cooldown],
       -- | The balance of the account that is available for transactions.
       aiAccountAvailableAmount :: !Amount,
-      -- | Flag indicating whether the account is currently suspended.
+      -- | Flag indicating whether the account is currently suspended. A
+      -- suspended account is in effect not participating in the consensus
+      -- protocol. The aiAccountIsSuspended flag does not have any effect on
+      -- stake or delegators of a validator.
       aiAccountIsSuspended :: !Bool
     }
     deriving (Eq, Show)
--- a/haskell-src/Concordium/Types/Execution.hs
+++ b/haskell-src/Concordium/Types/Execution.hs
@@ -969,7 +969,10 @@ getPayload spv size = S.isolate (fromIntegral size) (S.bytesRead >>= go)
     supportMemo = supportsMemo spv
     supportDelegation = protocolSupportsDelegation spv
     supportEncrypted = supportsEncryptedTransfers spv
-    configureBakerBitMask = 0b0000000111111111
+    supportSuspend = protocolSupportsSuspend spv
+    configureBakerBitMask
+        | supportSuspend = 0b0000000111111111
+        | otherwise = 0b0000000011111111
     configureDelegationBitMask = 0b0000000000000111
 
 -- | Builds a set from a list of ascending elements.
--- a/haskell-src/Concordium/Types/Execution.hs
+++ b/haskell-src/Concordium/Types/Execution.hs
@@ -357,8 +357,9 @@ data Payload
           cbBakingRewardCommission :: !(Maybe AmountFraction),
           -- | The commission the pool owner takes on finalization rewards.
           cbFinalizationRewardCommission :: !(Maybe AmountFraction),
-          -- | Whether the account should be suspended or not.
-          cbSuspended :: !(Maybe Bool)
+          -- | Whether the account should be suspended (`Just True`) or resumed
+          --   (`Just False`). `Nothing` has no effect.
+          cbSuspend :: !(Maybe Bool)
         }
     | -- | Configure an account's stake delegation.
       -- As with 'ConfigureBaker', the serialization uses a 16-bit bitmap.
@@ -406,6 +407,7 @@ instance S.Serialize TransactionType where
         TTTransferWithScheduleAndMemo -> S.putWord8 18
         TTConfigureBaker -> S.putWord8 19
         TTConfigureDelegation -> S.putWord8 20
+
     get =
         S.getWord8 >>= \case
             0 -> return TTDeployModule
@@ -652,7 +654,7 @@ instance AE.FromJSON Payload where
                 cbTransactionFeeCommission <- obj AE..: "transactionFeeCommission"
                 cbBakingRewardCommission <- obj AE..: "bakingRewardCommission"
                 cbFinalizationRewardCommission <- obj AE..: "finalizationRewardCommission"
-                cbSuspended <- obj AE..: "bakerSuspended"
+                cbSuspend <- obj AE..: "bakerSuspend"
                 return ConfigureBaker{..}
             "configureDelegation" -> do
                 cdCapital <- obj AE..: "capital"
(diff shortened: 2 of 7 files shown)


haskell-src/Concordium/Types/Execution.hs
```diff
@@ -356,7 +356,9 @@ data Payload
           -- | The commission the pool owner takes on baking rewards.
           cbBakingRewardCommission :: !(Maybe AmountFraction),
           -- | The commission the pool owner takes on finalization rewards.
-          cbFinalizationRewardCommission :: !(Maybe AmountFraction)
+          cbFinalizationRewardCommission :: !(Maybe AmountFraction),
+          -- | Whether the account should be suspended (`Just True`) or resumed (`Just False`). `Nothing` has no effect.
```
drsk0, line 360: I guess in general, a `ConfigureBaker` payload with fields set to `Nothing` will not change the configuration?

Victor-N-Suadicani (commented):
I'm not sure where it would fit best but it might be good to expand in a comment somewhere in the code what "suspend" and "resume" means or maybe link/point to some documentation somewhere that explains the terms, if we have that :)

haskell-src/Concordium/Types/Execution.hs
```diff
@@ -945,6 +950,7 @@ getPayload spv size = S.isolate (fromIntegral size) (S.bytesRead >>= go)
                 cbTransactionFeeCommission <- maybeGet 5
                 cbBakingRewardCommission <- maybeGet 6
                 cbFinalizationRewardCommission <- maybeGet 7
+                cbSuspend <- maybeGet 8
```
td202, line 953: For backwards compatibility, deserialization must fail if the suspend field is present for protocol versions that don't support suspending validators. (cf. how `supportDelegation` is used to guard deserialization of `ConfigureBaker`/`ConfiguredDelegation`.)

drsk0:
@Victor-N-Suadicani That's a good suggestion. I'll add this to this PR.

haskell-tests/Generators.hs
```diff
@@ -297,6 +297,7 @@ genPayloadConfigureBaker = do
     cbTransactionFeeCommission <- liftArbitrary genAmountFraction
     cbBakingRewardCommission <- liftArbitrary genAmountFraction
     cbFinalizationRewardCommission <- liftArbitrary genAmountFraction
+    cbSuspend <- arbitrary
```
td202, line 300: Whether this is (possibly) present should be guarded by passing a `ProtocolVersion` or a `Bool`, since otherwise this *should* break the serialization testing for older protocol versions.

Possibly the `genInvalidPayload...` functions should also be updated.

td202 (changes requested):

haskell-src/Concordium/Types/Execution.hs
drsk0: I'll open a separate PR to add the necessary type level machinery that implements the guard.

drsk0: https://github.com/Concordium/concordium-base/pull/551

haskell-tests/Generators.hs
drsk0: It's now guarded behind `protocolSupportsSuspend` (as introduced in #551). But I'm not 100% I got this right. The serialization is unchanged, as older protocols will just not set the corresponding bits in the bitmap.

haskell-tests/Generators.hs
```diff
@@ -180,9 +180,9 @@ genPayloadUnsafe =
           genPayloadUpdateBakerStake,
           genPayloadUpdateBakerRestateEarnings,
           genPayloadUpdateBakerKeys,
-          genPayloadConfigureBaker,
           genPayloadConfigureDelegation
         ]
+            ++ [genPayloadConfigureBaker pv | pv <- [P4, P5, P6, P7, P8]]
```
drsk0, line 185: Is there a reason there is no `Enum` instance for `ProtocolVersion`?

haskell-src/Concordium/Types/Execution.hs
```diff
@@ -929,6 +934,24 @@ getPayload spv size = S.isolate (fromIntegral size) (S.bytesRead >>= go)
...
+                        | otherwise = return Nothing
+                cbCapital <- maybeGet 0
+                cbRestakeEarnings <- maybeGet 1
+                cbOpenForDelegation <- maybeGet 2
+                cbKeysWithProofs <- maybeGet 3
+                cbMetadataURL <- maybeGet 4
+                cbTransactionFeeCommission <- maybeGet 5
+                cbBakingRewardCommission <- maybeGet 6
+                cbFinalizationRewardCommission <- maybeGet 7
+                cbSuspend <- maybeGet 8
```
td202, line 953: There is no reason to have an entire duplicate case. Just have one case, and we don't need to test `supportSuspend` here because the definition of the bitmask determines whether `cdSuspend` can be present.

haskell-tests/Generators.hs
td202: Probably not a particularly good one. It might have to do with the singletons stuff, at least historically.

haskell-tests/Generators.hs
```diff
@@ -297,6 +297,7 @@ genPayloadConfigureBaker = do
     cbTransactionFeeCommission <- liftArbitrary genAmountFraction
     cbBakingRewardCommission <- liftArbitrary genAmountFraction
     cbFinalizationRewardCommission <- liftArbitrary genAmountFraction
+    cbSuspend <- if pv < P8 then return Nothing else arbitrary
```
td202, line 300: As a general rule, I'd suggest preferring to test `supportsAccountSuspension (accountVersionFor pv)` which is clearer semantically.

td202 (approved):
I suggest having a test case that specifically ensures that `ConfigureBaker` payloads with the suspend field cannot be deserialized at old protocol versions.

haskell-src/Concordium/Types/Accounts.hs
```diff
@@ -732,7 +732,7 @@ data AccountInfo = AccountInfo
       aiAccountCooldowns :: ![Cooldown],
       -- | The balance of the account that is available for transactions.
       aiAccountAvailableAmount :: !Amount,
-      -- | Flag indicating whether the account is currently suspended.
+      -- | Flag indicating whether the account is currently suspended. A suspended account is in effect not participating in the consensus protocol. The aiAccountIsSuspended flag does not have any effect on stake or delegators of a validator. The full design proposal can be found on confluence under [Suspend Inactive Validators].
```
td202, line 735: As a style point, try to avoid lines longer than 100 characters. Also, I don't think it's good to reference confluence (which is not public) from the codebase (which is).

td202 (commented):
I suggest having a test case that specifically ensures that `ConfigureBaker` payloads with the suspend field cannot be deserialized at old protocol versions.

haskell-src/Concordium/Types/Accounts.hs
drsk0: Thanks for flagging this, I overlooked the soft-wrap my editor was doing there. I removed the confluence reference, it would be nice to point to design documents though somehow.

haskell-tests/Generators.hs
drsk0: done

PR #559, opened 2024-09-18
Move compilation target from `concordium_base/cargo.toml` to `Setup.hs`

soerenbf:
## Purpose

To avoid conflicts when compiling dependants on `concordium_base`, this proposal moves the compilation target specification from the `concordium_base` crate to the compilation done by the haskell dependant.

One instance where a conflict occurs is, if `concordium_base` is used as both a dependency in `build.rs` and `src/*.rs`. In `./mobile_wallet` this was worked around by aligning the profile used for `build.rs` with the one used for `src/lib.rs`, however this was not a feasible solution in another project and is hard for other developers to debug if they use this in a similar fashion.

I don't know if this should be marked as a change in the changelog or not?

## Checklist

      hard-to-understand areas.


Setup.hs
```diff
@@ -29,67 +25,78 @@ linuxBuild :: Bool -> WithEnvAndVerbosity
...
     noticeNoWrap verbosity "Dynamic linking."
-    rawSystemExitWithEnv verbosity "cargo" (addFeatures ["build", "--release", "--manifest-path", "rust-src/Cargo.toml"]) env
-    let copyLib lib = do
-            let source = "../rust-src/target/release/lib" ++ lib
-                target = "./lib/lib" ++ lib
+    let makeLib lib = do
+            let libName = fst lib
+                libFeatures = snd lib
+            rawSystemExitWithEnv verbosity "cargo" (["rustc", "--release", "--manifest-path", "rust-src/" ++ libName ++ "/Cargo.toml", "--crate-type", "cdylib"] ++ libFeatures) env
+            notice verbosity "Linking libraries to ./lib"
```
td202, line 48: I think here (and in the other places like this) it would be better to log "Linking library <libname> to ./lib", since this occurs for each library.

PR #560, opened 2024-09-24
Remove check on padding in mobile wallets library

DOBEN:
## Purpose

Not-enforcing padding of the base64 schema when signing a binary message in the CryptoX wallets.

In the past, some tools added and/or required padding on the schema in the Concordium ecosystem while others did not. At some point, we relaxed it and removed that requirement:
https://github.com/Concordium/concordium-dapp-libraries/pull/63/files

Currently,
- `cargo-concordium` does not add padding
- `browser wallet` does not enforce padding
- `dapp-libraries` does not enforce padding

Padding means that there are zero, one, or two `=` characters added at the end of the base64 string representing the schema to make it dividable.

The mobile wallets still enforce padding which was incompatible with the schema produced by `cargo-concordium`.

## Changes

- Remove check on padding in mobile wallets library

### DIFF
--- a/mobile_wallet/CHANGELOG.md
+++ b/mobile_wallet/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.25.3
+
+- Changed the base64 schema decoding to allow padded and not padded strings.
+
 ## 0.25.2
 
 - Allowed use of functions deprecated due to the removal of encrypted transfers in protocol version 7.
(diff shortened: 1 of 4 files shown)


Radiokot (changes requested):
It does not remove check on padding but requires no padding, now values with padding get rejected

Radiokot (approved):

PR #569, opened 2024-11-11
Update Rust and Haskell versions.

td202:
## Purpose

Update the Rust version to 1.82 and the GHC version to 9.6.6 (LTS-22.39).

## Changes

- Update versions.
- Update base image tag to `rust-1.82_ghc-9.6.6-1`.
- Clippy fixes.
- Update rust library versions.

## Checklist

      hard-to-understand areas.


rust-src/concordium_base/src/id/account_holder.rs
```diff
@@ -704,11 +701,7 @@ where
 }
 
 #[allow(clippy::too_many_arguments)]
-pub fn create_unsigned_credential<
-    P: Pairing,
-    C: Curve<Scalar = P::ScalarField>,
-    AttributeType: Attribute<C::Scalar>,
->(
+pub fn create_unsigned_credential<P: Pairing, C: Curve<Scalar = P::ScalarField>, AttributeType>(
```
soerenbf, line 704: **Non-blocking**: Is this clippy? I'm wondering why the trait bound declarations are not consistent 🤷

soerenbf (approved):
LGTM! Good job going through all this 😅 

CI fails though so there's that 🤷

.github/workflows/build-test-sources.yaml
```diff
@@ -85,7 +85,7 @@ jobs:
     strategy:
       matrix:
         plan:
-        - rust: 1.73
+        - rust: 1.82
```
soerenbf, line 88: I'm wondering, is it necessary to test that we can build all this with the MSRV as well? Or is it enough that it is specified in `cargo.toml`?

rust-src/concordium_base/src/id/account_holder.rs
td202: Clippy doesn't like having bounds in both the parameter and the where clause, and suggested having them in the where clause. But I think it's probably better to have them in the parameter.

.github/workflows/build-test-sources.yaml
td202: You're right. The `check-base-compiles` job is supposed to do this, so I'll set that to the MSRV.

PR #574, opened 2024-11-20
Support timestamp JSON deserialization from strings.

td202:
## Purpose

Allow JSON deserializing `Timestamp` in Haskell from the format produced by the [Rust serialization](https://github.com/Concordium/concordium-base/blob/58f871e57ab8543ae499d3cccc14ff7ce7b4842a/smart-contracts/contracts-common/concordium-contracts-common/src/types.rs#L873). The divergence was introduced as a consequence of https://github.com/Concordium/concordium-base/pull/499/files, which unified two previously distinct Rust `Timestamp` implementations. The JSON serialization is in particular used by the transaction logger when ScheduleTransfer transactions are logged in the wallet proxy database. This change allows the wallet proxy to parse the String-formatted timestamps.

## Changes

- New `FromJSON` implementation for `Timestamp`.
- Testing

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-tests/Types/TimestampSpec.hs
+++ b/haskell-tests/Types/TimestampSpec.hs
@@ -1,0 +1,46 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+-- | Basic JSON serialization and deserialization tests for 'Timestamp'.
+module Types.TimestampSpec where
+
+import Control.Monad
+import qualified Data.Aeson as AE
+import qualified Data.ByteString.Lazy as LBS
+import Test.Hspec
+import Test.QuickCheck
+
+import Concordium.Common.Time
+
+import Generators
+
+testEncodeDecode :: Property
+testEncodeDecode = forAll genTimestamp $ \ts -> AE.decode (AE.encode ts) === Just ts
+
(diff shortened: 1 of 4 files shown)


DOBEN (approved):

haskell-tests/Types/TimestampSpec.hs
```diff
@@ -0,0 +1,46 @@
...
+import Generators
+
+testEncodeDecode :: Property
+testEncodeDecode = forAll genTimestamp $ \ts -> AE.decode (AE.encode ts) === Just ts
+
+decodeExamples :: [(LBS.ByteString, Maybe Timestamp)]
+decodeExamples =
+    [ ("0", Just $ Timestamp 0),
+      ("1", Just $ Timestamp 1),
+      ("1.0", Just $ Timestamp 1),
```
soerenbf, line 23: This would not be a valid timestamp though... right? I feel like it would be best to limit the tests to only expected positives/negatives.. or am I misunderstanding something?

soerenbf (approved):

drsk0 (approved):
Looks good!

PR #607, opened 2025-03-18
Add CDDL schema for PLT transactions.

td202:
## Purpose

This PR introduces a CDDL schema that defines the CBOR representation used by the protocol-level token transactions.

## Changes

- Define `token-holder-transaction`.
- Define `token-governance-transaction`.
- Define `token-intialization-parameters`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cbor-schema/plt.cddl
+++ b/cbor-schema/plt.cddl
@@ -139,8 +139,6 @@ token-remove-deny-list = {
 token-initialization-parameters = { 
     ; The name of the token
     "name": text,
-    ; The number of decimals in the representation of the token
-    "decimals": uint,
     ; A URL pointing to the token metadata
     "metadata": text,
     ; Whether the token supports an allow list
--- a/cbor-schema/plt.cddl
+++ b/cbor-schema/plt.cddl
@@ -11,16 +11,18 @@ token-holder-operation = token-transfer
 
 ; A token transfer operation. This transfers a specified amount of tokens from the sender account
 ; (implicit) to the recipient account.
-token-transfer = { 
+token-transfer = [
     ; The operation type is "transfer".
-    "type": "transfer",
-    ; The amount of tokens to transfer.
-    "amount": token-amount,
-    ; The recipient account.
-    "recipient": token-receiver,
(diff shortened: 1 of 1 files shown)


cbor-schema/plt.cddl
```diff
@@ -0,0 +1,129 @@
...
+token-update-list = {
+    ; The operation type is "update-list".
+    "type": "update-list",
+    ; The action to take: "add" or "remove".
+    "action": "add" / "remove",
+    ; The list to update: "allow" or "deny".
+    "list": "allow" / "deny",
+    ; The account to add or remove.
+    "target": token-receiver
+}
```
td202, line 105: I went for this representation, but arguably a more concise format would be appropriate.

cbor-schema/plt.cddl
```diff
@@ -0,0 +1,156 @@
...
+    "amount": token-amount,
+    ; The recipient account.
+    "recipient": token-receiver,
+    ; An optional memo.
+    ? "memo": memo
+}
+
+; We represent token amounts as a decimal fraction.
+; The range and precision may be constrained by the token itself.
+token-amount = decfrac
```
soerenbf, line 27: I feel like we might shoot ourselves in the foot with this representation of token amounts. In the protobuf types, token amounts are represented with the integer amount and a number of decimals. This allows the token amount to have a fractional number representation while also allowing seemless serialization and deserialization back into the same token amount representation.

Maybe we can use a tag here as well to ensure correct interpretation of the value?

soerenbf (commented):
Will there be definitions for the other dynamic values included in the protobuf definitions, i.e.
- Token module state
- Governance event
- Holder event

?

cbor-schema/plt.cddl
soerenbf: My assumptions about `decfrac` were wrong.. sorry 😅

limemloh (approved):

td202:
To be added in subsequent PRs:
- Token module state
- Token governance event
- Token holder event
- Reject reasons

PR #612, opened 2025-04-03
plt: protocol level token types

drsk0:
This adds basic types used for protocol level tokens. For now, only the bare minimum definitions have been added. Instances should be added when needed.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/haskell-src/Concordium/Types.hs
+++ b/haskell-src/Concordium/Types.hs
@@ -179,6 +179,13 @@ module Concordium.Types (
     -- * PartsPerHundredThousands
     PartsPerHundredThousands (..),
     partsPerHundredThousandsToRational,
+
+    -- * Protocol Level Tokens
+    TokenIndex (..),
+    TokenAmount (..),
+    TokenId (..),
+    TokenModuleRef (..),
+    Cbor (..),
 ) where
 
 import Data.Data (Data, Typeable)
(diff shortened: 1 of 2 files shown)


haskell-src/Concordium/Types.hs
```diff
@@ -1148,3 +1155,21 @@ createAlias (AccountAddress addr) count = AccountAddress ((addr .&. mask) .|. re
 -- Template haskell derivations. At the end to get around staging restrictions.
 $(deriveJSON defaultOptions{sumEncoding = TaggedObject{tagFieldName = "type", contentsFieldName = "address"}} ''Address)
 makeLenses ''CommissionRates
+
+-- | The index into the global token table.
+newtype TokenIndex = TokenIndex {theTokenIndex :: Word64}
+
+-- | A token amount. The encoding is  VLQ/ULEB128 (TODO: (drsk) tbd). To compute the
```
drsk0, line 1162: We need to decide which encoding we will be using.

td202: I have an implementation in this PR: https://github.com/Concordium/concordium-node/pull/1359/files#diff-1a728bc8e6dd831cf39af11900a8cc0d235e3df23716e929b4d95f270d83e11dR32

td202 (approved):
This is OK, but I feel like it may not be necessary to have `TokenIndex` and `TokenAmount` in base, as these types are probably not exposed externally from the node.

drsk0:
I wait for the merge until https://github.com/Concordium/concordium-node/pull/1359 has been merged and it becomes clearer which types are actually already defined in the node.

drsk0:
Not necessary, as types have been introduced in the meantime.

PR #614, opened 2025-04-08
Introduce types and implemenations for CreatePLT update payload

limemloh:
## Purpose

Base support for CreatePLT update payload.

Depends on https://github.com/Concordium/concordium-grpc-api/pull/86
and is related to https://github.com/Concordium/concordium-node/pull/1361

### DIFF
--- a/haskell-src/Concordium/Types/ProtocolVersion.hs
+++ b/haskell-src/Concordium/Types/ProtocolVersion.hs
@@ -238,6 +238,7 @@ module Concordium.Types.ProtocolVersion (
     supportsAccountSignatureChecks,
     supportsContractInspectionQueries,
     supportsEncryptedTransfers,
+    supportsProtocolLevelTokens,
 
     -- * Defunctionalisation symbols
     P1Sym0,
@@ -794,3 +795,17 @@ supportsEncryptedTransfers = \case
     SP7 -> False
     SP8 -> False
     SP9 -> False
+
+-- | Whether the protocol version supports Protocol Level Tokens (PLT).
+-- (Supported in 'P9' and onwards.)
+supportsProtocolLevelTokens :: SProtocolVersion pv -> Bool
+supportsProtocolLevelTokens = \case
+    SP1 -> False
+    SP2 -> False
+    SP3 -> False
+    SP4 -> False
+    SP5 -> False
+    SP6 -> False
+    SP7 -> False
+    SP8 -> False
+    SP9 -> True
--- a/haskell-src/Concordium/Types/ProtocolVersion.hs
+++ b/haskell-src/Concordium/Types/ProtocolVersion.hs
@@ -199,6 +199,7 @@ module Concordium.Types.ProtocolVersion (
     SupportsPLT,
     supportsPLT,
     sSupportsPLT,
+    protocolSupportsPLT,
     -- | Determine whether a specific account version supports protocol level tokens.
     AVSupportsPLT,
     -- | Determine whether a specific protocol version supports protocol level tokens.
@@ -238,7 +239,6 @@ module Concordium.Types.ProtocolVersion (
     supportsAccountSignatureChecks,
     supportsContractInspectionQueries,
     supportsEncryptedTransfers,
-    supportsProtocolLevelTokens,
 
     -- * Defunctionalisation symbols
     P1Sym0,
@@ -628,6 +628,13 @@ type AVSupportsPLT (av :: AccountVersion) =
 type PVSupportsPLT (pv :: ProtocolVersion) =
     AVSupportsPLT (AccountVersionFor pv)
 
+-- | Whether the protocol version supports Protocol Level Tokens (PLT).
+protocolSupportsPLT :: SProtocolVersion pv -> Bool
+{-# INLINE protocolSupportsPLT #-}
+protocolSupportsPLT spv = case sSupportsPLT (sAccountVersionFor spv) of
+    STrue -> True
+    SFalse -> False
+
 -- | Constraint that an account version supports flexible cooldown.
 --
 -- Note, we do not use 'Assert' here, since that results in a weaker constraint that requires
@@ -795,17 +802,3 @@ supportsEncryptedTransfers = \case
     SP7 -> False
     SP8 -> False
     SP9 -> False
-
--- | Whether the protocol version supports Protocol Level Tokens (PLT).
--- (Supported in 'P9' and onwards.)
-supportsProtocolLevelTokens :: SProtocolVersion pv -> Bool
-supportsProtocolLevelTokens = \case
-    SP1 -> False
-    SP2 -> False
-    SP3 -> False
-    SP4 -> False
-    SP5 -> False
-    SP6 -> False
-    SP7 -> False
-    SP8 -> False
-    SP9 -> True
--- a/haskell-src/Concordium/Types/ProtocolVersion.hs
+++ b/haskell-src/Concordium/Types/ProtocolVersion.hs
@@ -287,6 +287,7 @@ $( singletons
             | ChainParametersV1
             | ChainParametersV2
             | ChainParametersV3
+            | ChainParametersV4
             deriving (Eq, Ord)
 
         chainParametersVersionFor :: ProtocolVersion -> ChainParametersVersion
@@ -298,7 +299,7 @@ $( singletons
         chainParametersVersionFor P6 = ChainParametersV2
         chainParametersVersionFor P7 = ChainParametersV2
         chainParametersVersionFor P8 = ChainParametersV3
-        chainParametersVersionFor P9 = ChainParametersV3
+        chainParametersVersionFor P9 = ChainParametersV4
 
         -- \* Account versions
 
@@ -454,6 +455,7 @@ chainParameterVersionToWord64 ChainParametersV0 = 0
 chainParameterVersionToWord64 ChainParametersV1 = 1
 chainParameterVersionToWord64 ChainParametersV2 = 2
 chainParameterVersionToWord64 ChainParametersV3 = 3
+chainParameterVersionToWord64 ChainParametersV4 = 4
 
 instance Serialize ProtocolVersion where
     put = putWord64be . protocolVersionToWord64
--- a/haskell-src/Concordium/Types/Queries/Tokens.hs
+++ b/haskell-src/Concordium/Types/Queries/Tokens.hs
@@ -9,9 +9,11 @@ import qualified Concordium.Crypto.SHA256 as SHA256
 import qualified Concordium.Types as Types
 import qualified Concordium.Types.HashableTo as HashableTo
 import qualified Data.Aeson as Aeson
+import qualified Data.Aeson.Types as Aeson
 import qualified Data.ByteString.Short as BS
 import qualified Data.Hashable as Hashable
 import qualified Data.Serialize as Serialize
+import qualified Data.Text.Encoding as Text
 import Data.Word
 
 -- | The unique token identifier. This is given as a symbol unique across the
@@ -19,10 +21,13 @@ import Data.Word
 newtype TokenId = TokenId {symbol :: BS.ShortByteString} deriving (Eq, Show)
 
 instance Aeson.ToJSON TokenId where
-    toJSON TokenId{} = undefined
+    -- decodeUtf8 will throw an exception if it fails, but we should be safe since the TokenId
+    -- should enforce valid UTF-8.
+    toJSON TokenId{..} = Aeson.String $ Text.decodeUtf8 $ BS.fromShort symbol
 
 instance Aeson.FromJSON TokenId where
-    parseJSON = undefined
+    parseJSON (Aeson.String text) = return $ TokenId $ BS.toShort $ Text.encodeUtf8 text
+    parseJSON invalid = Aeson.prependFailure "parsing TokenId failed" (Aeson.typeMismatch "String" invalid)
 
 instance Serialize.Serialize TokenId where
     get = do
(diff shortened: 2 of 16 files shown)


haskell-src/Concordium/Types/Queries/Tokens.hs
```diff
@@ -1,13 +1,33 @@
...
 
 -- | The unique token identifier. This is given as a symbol unique across the
 --  whole chain.
 newtype TokenId = TokenId {symbol :: BS.ShortByteString} deriving (Eq, Show)
 
+instance Aeson.ToJSON TokenId where
+    toJSON TokenId{} = undefined
+
+instance Aeson.FromJSON TokenId where
+    parseJSON = undefined
```
limemloh, line 25: Forgot to add these, will do in a minute.

haskell-src/Concordium/Types/Queries/Tokens.hs
```diff
@@ -1,13 +1,33 @@
...
+
+instance Serialize.Serialize TokenId where
+    get = TokenId <$> Serialize.get
+    put (TokenId symbol) = Serialize.put symbol
```
limemloh, line 29: will include the length (word8) here as well

haskell-src/Concordium/Types/Queries/Tokens.hs
```diff
@@ -39,3 +59,78 @@ data TokenAccountState = TokenAccountState
...
     deriving (Eq, Show)
+
+-- | Parameter for a Token module.
+newtype TokenParameter = TokenParameter {parameterBytes :: BS.ShortByteString}
+    deriving (Eq)
+    deriving (Aeson.ToJSON, Aeson.FromJSON, Show) via ByteStringHelpers.ByteStringHex
+
+instance Serialize.Serialize TokenParameter where
+    get = TokenParameter <$> Serialize.get
+    put (TokenParameter parameter) = Serialize.put parameter
```
limemloh, line 70: Adding a length here as well (word16be)

haskell-src/Concordium/Types/Queries/Tokens.hs
limemloh: done

haskell-src/Concordium/Types/ProtocolVersion.hs
```diff
@@ -794,3 +795,17 @@ supportsEncryptedTransfers = \case
     SP7 -> False
     SP8 -> False
     SP9 -> False
+
+-- | Whether the protocol version supports Protocol Level Tokens (PLT).
+-- (Supported in 'P9' and onwards.)
+supportsProtocolLevelTokens :: SProtocolVersion pv -> Bool
+supportsProtocolLevelTokens = \case
```
drsk0, line 802: I already introduced `supportsPLT`. We should align the naming.

haskell-src/Concordium/Types/Updates.hs
```diff
@@ -978,6 +988,7 @@ extractKeysIndices p =
         BlockEnergyLimitUpdatePayload{} -> getLevel2KeysAndThreshold asParamConsensusParameters
         FinalizationCommitteeParametersUpdatePayload{} -> getLevel2KeysAndThreshold asPoolParameters
         ValidatorScoreParametersUpdatePayload{} -> getLevel2KeysAndThreshold asPoolParameters
+        CreatePLTUpdatePayload{} -> undefined
```
drsk0, line 991: Can you replace this with an error and a message referencing the issue for the implementation?

DOBEN (approved):

haskell-src/Concordium/Types/Updates.hs
```diff
@@ -840,6 +845,8 @@ data UpdatePayload
       FinalizationCommitteeParametersUpdatePayload !FinalizationCommitteeParameters
     | -- | Update the validator score parameters (chain parameters version 3)
       ValidatorScoreParametersUpdatePayload !ValidatorScoreParameters
+    | -- | Issue a new Protocol Level Token (PLT) (Concordium Protocol 9)
```
drsk0, line 848: Do we already have a new chain parameter version for P9? I'd probably reference this in the comment.

haskell-src/Concordium/Types/Queries/Tokens.hs
```diff
@@ -39,3 +68,82 @@ data TokenAccountState = TokenAccountState
...
+        Serialize.putWord16be (fromIntegral (BS.length parameter))
+        Serialize.putShortByteString parameter
+
+-- | Module reference for a token module.
+newtype TokenModuleRef = TokenModuleRef {tokenModuleRef :: SHA256.Hash}
+    deriving (Eq, Ord, Hashable.Hashable)
+    deriving (Aeson.FromJSON, Aeson.ToJSON, Read) via SHA256.Hash
+
+instance Show TokenModuleRef where
+    show (TokenModuleRef m) = show m
```
drsk0, line 91: Is it not possible to derive the show instance?

haskell-src/Concordium/Types/Updates.hs
```diff
@@ -907,6 +915,7 @@ getUpdatePayload spv =
         21 | GASRewardsVersion1 <- gasRewardsVersionFor cpv -> GASRewardsCPV2UpdatePayload <$> get
         22 | isSupported PTFinalizationCommitteeParameters cpv -> FinalizationCommitteeParametersUpdatePayload <$> get
         23 | isSupported PTValidatorScoreParameters cpv -> ValidatorScoreParametersUpdatePayload <$> get
+        24 | supportsProtocolLevelTokens spv -> CreatePLTUpdatePayload <$> get
```
drsk0, line 918: I think we also have a new chain parameter version for P9. I'd align the check with the other `isSupported` checks here depending on `cpv`.

haskell-src/Concordium/Types/Queries/Tokens.hs
```diff
@@ -39,3 +68,82 @@ data TokenAccountState = TokenAccountState
       memberDenyList :: !Bool
     }
     deriving (Eq, Show)
+
+-- | Parameter for a Token module.
+newtype TokenParameter = TokenParameter {parameterBytes :: BS.ShortByteString}
+    deriving (Eq)
+    deriving (Aeson.ToJSON, Aeson.FromJSON, Show) via ByteStringHelpers.ByteStringHex
```
drsk0, line 75: I'm seeing this the first time for the derivation of `ToJSON`. Are we always using `ByteStringHelpers.ByteStringHex` for the derivation of JSON instances for byte strings?

haskell-src/Concordium/Types/Queries/Tokens.hs
```diff
@@ -39,3 +68,82 @@ data TokenAccountState = TokenAccountState
       memberDenyList :: !Bool
     }
     deriving (Eq, Show)
+
+-- | Parameter for a Token module.
```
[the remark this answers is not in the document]
limemloh: I don't think we can assume that here, this will probably be used for user input and we probably don't want to check the content to be valid here, but as part of the token module.

haskell-src/Concordium/Types/Queries/Tokens.hs
limemloh: Can't say whether we always do this, but we do for smart contract parameters, contract events, memos, register data at least, which are all arbitrary bytestrings.

haskell-src/Concordium/Types/Queries/Tokens.hs
limemloh: Yes you are right, `Serialize` as well actually.

haskell-src/Concordium/Types/Updates.hs
limemloh: I can introduce another chain parameter version for P9, but currently we don't have one.
Do we expect any new parameters?

haskell-src/Concordium/Types/ProtocolVersion.hs
limemloh: anyway, I'll adjust this to fit with the naming currently used.

haskell-src/Concordium/Types/Updates.hs
limemloh: I'll add a message, but I was not able to find an/the issue tracking this.

haskell-src/Concordium/Types/Updates.hs
limemloh: We don't at the moment and it is not clear to me whether we will have it.

haskell-src/Concordium/Types/Queries/Tokens.hs
```diff
@@ -1,13 +1,42 @@
...
+    parseJSON invalid = Aeson.prependFailure "parsing TokenId failed" (Aeson.typeMismatch "String" invalid)
+
+instance Serialize.Serialize TokenId where
+    get = do
+        len <- Serialize.getWord8
+        TokenId <$> Serialize.getShortByteString (fromIntegral len)
+    put (TokenId symbol) = do
+        Serialize.putWord8 (fromIntegral (BS.length symbol))
+        Serialize.putShortByteString symbol
+
```
td202, line 39: I have a PR (https://github.com/Concordium/concordium-base/pull/615) refactoring `TokenId` which includes serialization and enforces a length limit of 255.

haskell-src/Concordium/Types/Updates.hs
td202: I think we probably will have one, but this is not an update to a chain parameter as such, so it would be correct to say "from protocol version 9".

haskell-src/Concordium/Types/Queries/Tokens.hs
td202, line 1: The additions here go beyond queries, so either they should be moved out to a separate file, or rename this one as a whole to `Concordium.Types.Tokens`.

haskell-src/Concordium/Types/Updates.hs
drsk0: I'm also not sure. But would be good to check.

haskell-src/Concordium/Types/Queries/Tokens.hs
limemloh: Merge it in now

haskell-src/Concordium/Types/Queries/Tokens.hs
limemloh: I'll create `Concordium.Types.Tokens`

limemloh: Well, I guess not due to cyclic dependencies

haskell-src/Concordium/Types/Queries/Tokens.hs
```diff
@@ -1,11 +1,13 @@
...
+{-# LANGUAGE OverloadedStrings #-}
 
 -- | Types for protocol level tokens (PLT).
 module Concordium.Types.Queries.Tokens where
 
-import Data.Word
+import qualified Concordium.Types as Types
 
-import Concordium.Types
+import Data.Word
```
td202, line 10: Style: Generally, we group imports as:
- External imports
- Imports from `concordium-base`
- Imports from the package (e.g. `concordium-node`)

haskell-src/Concordium/Types/Updates.hs
```diff
@@ -978,6 +987,7 @@ extractKeysIndices p =
         BlockEnergyLimitUpdatePayload{} -> getLevel2KeysAndThreshold asParamConsensusParameters
         FinalizationCommitteeParametersUpdatePayload{} -> getLevel2KeysAndThreshold asPoolParameters
         ValidatorScoreParametersUpdatePayload{} -> getLevel2KeysAndThreshold asPoolParameters
+        CreatePLTUpdatePayload{} -> error "Not implemented yet, authorization of this update type is planned for a later iteration"
```
td202, line 990: Suggestion: Add a TODO comment referencing https://linear.app/concordium/issue/NOD-701/add-authorization-and-sequence-number-for-createplt-chain-updates

haskell-tests/Generators.hs
```diff
@@ -1039,6 +1069,72 @@ genGASRewards = do
...
+-- | Generate a valid UTF-8 string of the specified length.
+genUtf8String :: Int -> Gen [Word8]
+genUtf8String len
+    | len <= 0 = return []
+    | otherwise = do
+        c <- resize len genUtf8Char
+        rest <- genUtf8String (len - length c)
+        return (c ++ rest)
+
+genTokenId :: Gen TokenId
```
td202, line 1124:
```suggestion
-- | Generate an arbitrary 'TokenId', consisting of up to 255 bytes that is a valid UTF-8 string.
genTokenId :: Gen TokenId
```

haskell-tests/Generators.hs
```diff
@@ -1039,6 +1069,72 @@ genGASRewards = do
     _gasChainUpdate <- makeAmountFraction <$> choose (0, 100000)
     return GASRewards{..}
 
+genTokenParameter :: Gen TokenParameter
+genTokenParameter = do
+    n <- choose (0, 1000)
+    TokenParameter . BSS.pack <$> vector n
+
+genTokenModuleRef :: Gen TokenModuleRef
```
td202, line 1077: Doc

haskell-tests/Generators.hs
```diff
@@ -1039,6 +1069,72 @@ genGASRewards = do
...
+        c <- resize len genUtf8Char
+        rest <- genUtf8String (len - length c)
+        return (c ++ rest)
+
+genTokenId :: Gen TokenId
+genTokenId = do
+    len <- chooseBoundedIntegral (0, 255)
+    TokenId . BSS.pack <$> genUtf8String len
+
+genCreatePLT :: Gen CreatePLT
```
td202, line 1129: Doc

td202 (approved):

haskell-src/Concordium/Types/Updates.hs
td202: Since this causes some tests to fail at present, I would suggest using a dummy implementation, such as `(Set.empty, 0)`.

PR #618, opened 2025-04-10
CBOR (de)serialization for token initialization parameter

td202:
## Purpose

Part of NOD-682

This implements CBOR encoding and decoding for the `TokenInitializationParameter` in line with the CDDL specification (https://github.com/Concordium/concordium-base/pull/607). The decoding is permissive (though it does not allow spurious additional map keys, etc.) and the encoding follows deterministic CBOR.

## Changes

- Add `cborg` package dependency.
- Helpers for decoding and encoding types as CBOR maps.
- Encoding and decoding of 'TokenAmount'.
- Encoding and decoding of 'TokenInitializationParameters'.
- Unit tests.
- JSON serialization for `TokenInitializationParameters` and `TokenAmount`.
- Modify the JSON serialization of `CreatePLT` to allow the `initializationParameters` to be given either as a hex-encoded byte string or as JSON-encoded `TokenInitializationParameters`.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Types.hs
+++ b/haskell-src/Concordium/Types.hs
@@ -1161,6 +1161,8 @@ createAlias (AccountAddress addr) count = AccountAddress ((addr .&. mask) .|. re
     rest = FBS.encodeInteger (toInteger (count .&. 0xffffff))
     mask = complement (FBS.encodeInteger 0xffffff) -- mask to clear out the last three bytes of the addr
 
+-- * Protocol level tokens
+
 -- | Parameter for a Token module.
 newtype TokenParameter = TokenParameter {parameterBytes :: BSS.ShortByteString}
     deriving (Eq)
@@ -1246,7 +1248,8 @@ instance AE.ToJSON CreatePLT where
               "tokenModule" AE..= _cpltTokenModule,
               "governanceAccount" AE..= _cpltGovernanceAccount,
               "decimals" AE..= _cpltDecimals,
-              "initializationParameters" AE..= _cpltInitializationParameters
+              "initializationParameters"
+                AE..= EncodedTokenInitializationParameters _cpltInitializationParameters
             ]
 
 instance AE.FromJSON CreatePLT where
@@ -1255,7 +1258,8 @@ instance AE.FromJSON CreatePLT where
         _cpltTokenModule <- o AE..: "tokenModule"
         _cpltGovernanceAccount <- o AE..: "governanceAccount"
         _cpltDecimals <- o AE..: "decimals"
-        _cpltInitializationParameters <- o AE..: "initializationParameters"
+        (EncodedTokenInitializationParameters _cpltInitializationParameters) <-
+            o AE..: "initializationParameters"
         return CreatePLT{..}
 
 -- Template haskell derivations. At the end to get around staging restrictions.
--- a/haskell-src/Concordium/Types.hs
+++ b/haskell-src/Concordium/Types.hs
@@ -211,8 +211,10 @@ import qualified Concordium.Crypto.VRF as VRF
 import Concordium.ID.Types
 import Concordium.Types.Block
 import Concordium.Types.HashableTo
+import qualified Concordium.Types.ProtocolLevelTokens.CBOR as CBOR
 import Concordium.Types.ProtocolVersion
 import Concordium.Types.SmartContracts
+import Concordium.Types.Tokens
 import qualified Data.FixedByteString as FBS
 
 import Control.Exception (assert)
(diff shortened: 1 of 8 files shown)


haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
```diff
@@ -0,0 +1,267 @@
...
+makeMapKeyEncoding :: Encoding -> MapKeyEncoding
+makeMapKeyEncoding meEncoding = MapKeyEncoding{meBytes = CBOR.toLazyByteString meEncoding, ..}
+
+-- | Encode a map deterministically. Specifically, the map length is definite and the keys are
+--  ordered in bytewise lexicographic order. It is assumed that all encodings in the map are
+--  deterministic.
+encodeMapDeterministic :: Map.Map MapKeyEncoding Encoding -> Encoding
+encodeMapDeterministic m =
+    encodeMapLen (fromIntegral $ Map.size m)
+        <> mconcat [meEncoding k <> v | (k, v) <- Map.toAscList m]
```
soerenbf, line 121: Are we doing [CDE](https://datatracker.ietf.org/doc/html/draft-ietf-cbor-cde-05) or [dCBOR](https://www.ietf.org/archive/id/draft-mcnally-deterministic-cbor-11)? The latter is a spec from BlockchainCommons which we use for address serialization, but it does not seem like there is a haskell library to do this?

Does cborg enforce "preferred serialization" when encoding?

td202: I think the main thing that `dCBOR` mandates beyond what we implement is NFC encoding of unicode strings. So I think we should be `CDE` compliant, and if the strings are NFC then we would also comply with `dCBOR`.

soerenbf: great 😄

haskell-src/Concordium/Types.hs
```diff
@@ -1219,6 +1174,30 @@ instance S.Serialize TokenParameter where
         S.putWord32be (fromIntegral (BSS.length parameter))
         S.putShortByteString parameter
 
+-- | A wrapper type for (de)-serializing an CBOR-encoded initialization parameter to/from JSON.
+--  This can parse either an JSON object representation of 'TokenInitializationParameters'
+--  (which is then re-encoded as CBOR) or a hex-encoded byte string. When rendering JSON,
+--  it will render as a JSON object if the contents can be decoded to a
+-- 'TokenInitializationParameters', or otherwise as the hex-encoded byte string.
+newtype EncodedTokenInitializationParameters = EncodedTokenInitializationParameters TokenParameter
```
DOBEN, line 1182: To use this wrapper, we need to use it instead of the `TokenParameter` here:

```
data CreatePLT = CreatePLT
    { -- | The symbol of the token.
      _cpltTokenSymbol :: !TokenId,
      -- | A SHA256 hash that identifies the token module implementation.
      _cpltTokenModule :: !TokenModuleRef,
      -- | The address of the account that will govern the token.
      _cpltGovernanceAccount :: !AccountAddress,
      -- | The number of decimal places used in the representation of amounts of this token. This determines the smallest representable fraction of the token.
      _cpltDecimals :: !Word8,
      -- | The initialization parameters of the token, encoded in CBOR.
      _cpltInitializationParameters :: !EncodedTokenInitializationParameters
    }
```

td202: I have fixed this - my idea is not to expose the wrapper type directly, but just to use it in the ToJSON and FromJSON definitions for CreatePLT.

limemloh (approved):

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
```diff
@@ -0,0 +1,313 @@
...
+                "Expected an array of length 2, but length was " ++ show len
+    exp10 <- decodeInt
+    mantissa <- decodeInteger
+    when (isNothing mLen) $ decodeBreakOr >>= \b -> unless b (fail "Expected end of array")
+    return $ scientific mantissa exp10
+
+-- * Encoding helpers
+
+-- | An 'Encoding' that represents a key in a CBOR map. The 'Ord' instance corresponds to
+--  lexicographic ording of the CBOR encodings, so that these can be used to order keys for
```
DOBEN, line 102:
```suggestion
--  lexicographic ordering of the CBOR encodings, so that these can be used to order keys for
```

DOBEN (approved):

PR #625, opened 2025-04-28
Expose the update sequence number for PLTs

limemloh:
## Purpose

Ref COR-1344.

Related PR for grpc: https://github.com/Concordium/concordium-grpc-api/pull/91
Related PR for node: https://github.com/Concordium/concordium-node/pull/1376

## Changes

- Moved `updateQueuesNextSequenceNumbers` concordium-node repo and made it inline.

### DIFF
--- a/haskell-src/Concordium/Types/Queries.hs
+++ b/haskell-src/Concordium/Types/Queries.hs
@@ -723,7 +723,9 @@ data NextUpdateSequenceNumbers = NextUpdateSequenceNumbers
       -- | Updates to the consensus version 2 finalization committee parameters
       _nusnFinalizationCommitteeParameters :: !U.UpdateSequenceNumber,
       -- | Updates to the consensus version 2 validator score parameters
-      _nusnValidatorScoreParameters :: !U.UpdateSequenceNumber
+      _nusnValidatorScoreParameters :: !U.UpdateSequenceNumber,
+      -- | Updates to protocol level tokens. Introduced in protocol version 9.
+      _nusnProtocolLevelTokensParameters :: !U.UpdateSequenceNumber
     }
     deriving (Show, Eq)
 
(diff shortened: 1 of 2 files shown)


haskell-src/Concordium/Types/Queries.hs
```diff
@@ -732,8 +734,14 @@ $(deriveJSON defaultOptions{fieldLabelModifier = firstLower . drop (length ("_nu
...
-updateQueuesNextSequenceNumbers :: UQ.PendingUpdates cpv -> NextUpdateSequenceNumbers
-updateQueuesNextSequenceNumbers UQ.PendingUpdates{..} =
+updateQueuesNextSequenceNumbers ::
+    -- | The queues with pending updates
+    UQ.PendingUpdates cpv ->
+    -- | The update sequence number for protocol level tokens.
+    -- No queue is associated with this chain update, so it is provided separately.
+    U.UpdateSequenceNumber ->
+    NextUpdateSequenceNumbers
+updateQueuesNextSequenceNumbers UQ.PendingUpdates{..} protocolLevelTokensSeqNumber =
```
DOBEN, line 744: It might be nicer to pass in the `updates` as input parameter to the function and then extract the `protocolLevelTokensSeqNumber` inside this function. Just thinking that this would be a better approach to handle more chain parameters in the future that do not have a queue.

limemloh: Yes, I agree. I did not do this because the `Updates` type lives in the `concordium-node` repo and did not want to move it into `concordium-base`.
Maybe this function could be moved to `concordium-node`

DOBEN (approved):

DOBEN (approved):

haskell-src/Concordium/Types/Queries.hs
limemloh: I tried [searching across all our repos](https://github.com/search?q=org%3AConcordium%20updateQueuesNextSequenceNumbers&type=code) and it is only used in one place in concordium-node, so I have made it inline in the node instead.

td202 (approved):

PR #626, opened 2025-04-28
Base support for P9 in concordium_base rust crate

limemloh:
## Purpose
Ref COR-1146.

Introduce basic types for plt in `rust-src/concordium-base`

Relevant PR in rust-sdk: https://github.com/Concordium/concordium-rust-sdk/pull/240 and https://github.com/Concordium/concordium-rust-sdk/pull/243 and https://github.com/Concordium/concordium-rust-sdk/pull/246

### DIFF
--- a/rust-src/concordium_base/src/protocol_level_tokens/token_amount.rs
+++ b/rust-src/concordium_base/src/protocol_level_tokens/token_amount.rs
@@ -9,6 +9,10 @@ pub struct TokenAmount {
 }
 
 impl TokenAmount {
+    /// Construct a [`TokenAmount`] representing an integer amount and zero
+    /// decimals.
+    pub fn from_integer(value: u64) -> Self { Self { value, decimals: 0 } }
+
     /// The number of decimals in the token amount.
     ///
     /// Together with the `raw_value` the token amount can be represented as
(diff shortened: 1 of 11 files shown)


rust-src/concordium_base/src/protocol_level_tokens/token_id.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    fn try_from(symbol: String) -> Result<Self, Self::Error> {
+        let symbol_byte_len = symbol.as_bytes().len();
+        if symbol_byte_len > TOKEN_ID_MAX_BYTE_LEN {
+            Err(TokenIdFromStringError {
+                actual_size: symbol_byte_len,
+            })
+        } else {
+            Ok(Self { symbol })
+        }
```
soerenbf, line 47: We'd need to check for utf8 validity here as well, I assume

rust-src/concordium_base/src/protocol_level_tokens/token_amount.rs
```diff
@@ -0,0 +1,235 @@
...
+impl TokenAmount {
+    /// Construct a [`TokenAmount`] representing an integer amount and zero
+    /// decimals.
+    pub fn from_integer(value: u64) -> Self { Self { value, decimals: 0 } }
+
+    /// The number of decimals in the token amount.
+    ///
+    /// Together with the `raw_value` the token amount can be represented as
+    /// `raw_value * 10^(-decimals)`
+    pub fn decimals(&self) -> u8 { self.decimals }
```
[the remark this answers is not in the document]
limemloh: Yes, the idea was to hide the inner details of the token amount, so the fields should be private...
I'll fix it

rust-src/concordium_base/src/protocol_level_tokens/token_amount.rs
```diff
@@ -0,0 +1,235 @@
...
+            std::cmp::Ordering::Less => {
+                let decimal_diff = other.decimals - self.decimals;
+                let scaled_self = u128::from(self.value) * 10u128.pow(decimal_diff.into());
+                scaled_self.cmp(&u128::from(other.value))
+            }
+            std::cmp::Ordering::Greater => {
+                let decimal_diff = self.decimals - other.decimals;
+                let scaled_other = u128::from(other.value) * 10u128.pow(decimal_diff.into());
+                u128::from(self.value).cmp(&scaled_other)
+            }
```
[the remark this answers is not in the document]
limemloh: I don't think it will panic with the current implementation, but it would overflow silently.
Not sure what the best approach would be. Maybe panicking is better in this case. Performance-wise it would probably be best to restrict the range of decimals, or we could use something larger than `u128` but I would expect it to be fine in practice.

rust-src/concordium_base/src/protocol_level_tokens/token_id.rs
soerenbf: great, rust ftw!

rust-src/concordium_base/src/protocol_level_tokens/token_id.rs
```diff
@@ -0,0 +1,74 @@
...
+
+impl common::Deserial for TokenId {
+    fn deserial<R: byteorder::ReadBytesExt>(source: &mut R) -> common::ParseResult<Self> {
+        let len = source.read_u8()?;
+        let mut buf = vec![0u8; len as usize];
+        source.read_exact(&mut buf)?;
+        let symbol = String::from_utf8(buf)?;
+        Ok(Self { symbol })
+    }
+}
```
DOBEN, line 74: Just curious, we we need `common:Serial/Deserial` for this type but not for the other types.

DOBEN (approved):

rust-src/concordium_base/src/protocol_level_tokens/token_id.rs
limemloh: Because it is used as part of the payload for a transaction

rust-src/concordium_base/src/protocol_level_tokens/token_amount.rs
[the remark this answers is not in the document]
limemloh: I should now support the entire token amount space.

PR #634, opened 2025-05-06
CBOR and schema for token module state.

td202:
## Purpose

Part of COR-1371
Required by https://github.com/Concordium/concordium-node/pull/1385

This defines a CDDL schema for representing token module state, and implements encoding and decoding CBOR from a Haskell type.

## Changes

- Add `token-module-state` to the CDDL schema.
- Define type `TokenModuleState` with CBOR encoding and decoding following the schema.
- Add tests.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cbor-schema/plt.cddl
+++ b/cbor-schema/plt.cddl
@@ -163,3 +163,34 @@ token-initialization-parameters = {
     ; Whether the token is burnable
     ? "burnable": bool .default false
 }
+
+; Token state
+
+; The Token Module state represents global state information that is maintained by the Token Module,
+; and is returned as part of a `GetTokenInfo` query. It does not include state that is managed by
+; the Token Kernel, such as the token identifier, global supply and governance account. It also
+; does not (typically) include account-specific state, which is returned as part of
+; `GetAccountInfo` instead.
+;
+; The "name" and "metadata" fields are required. Other fields are optional, and can be omitted if
+; the module implementation does not support them. The structure supports additional fields for
+; future extensibility. Non-standard fields (i.e. any fields that are not defined by a standard,
+; and are specific to the module implementation) may be included, and their tags should be
+; prefixed with an underscore ("_") to distinguish them as such.
+token-module-state = {
+    ; The name of the token
+    "name": text,
+    ; A URL pointing to the token metadata
+    "metadata": text,
+    ; Whether the token supports an allow list.
+    ? "allowList": bool,
+    ; Whether the token supports a deny list.
+    ? "denyList": bool,
+    ; Whether the token is mintable.
+    ? "mintable": bool,
+    ; Whether the token is burnable.
+    ? "burnable": bool,
+    ; Additional state information may be provided under further text keys, the meaning
+    ; of which are not defined the present specification.
+    * tstr => any
+}
--- a/cbor-schema/plt.cddl
+++ b/cbor-schema/plt.cddl
@@ -191,6 +191,6 @@ token-module-state = {
     ; Whether the token is burnable.
     ? "burnable": bool,
     ; Additional state information may be provided under further text keys, the meaning
-    ; of which are not defined the present specification.
-    * tstr => any
+    ; of which are not defined in the present specification.
+    * text => any
 }
--- a/haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
+++ b/haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
@@ -861,6 +861,11 @@ encodeTokenModuleState TokenModuleState{..} =
             ]
     k = at . makeMapKeyEncoding . encodeString
 
+-- | CBOR-encode a 'TokenModuleState' to a (strict) 'BS.ByteString'.
+--  This uses default values in the encoding.
+tokenModuleStateToBytes :: TokenModuleState -> BS.ByteString
+tokenModuleStateToBytes = CBOR.toStrictByteString . encodeTokenModuleState
+
 -- | Decode a CBOR-encoded 'TokenModuleState'.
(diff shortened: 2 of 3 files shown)


cbor-schema/plt.cddl
```diff
@@ -163,3 +163,34 @@ token-initialization-parameters = {
...
+; and is returned as part of a `GetTokenInfo` query. It does not include state that is managed by
+; the Token Kernel, such as the token identifier, global supply and governance account. It also
+; does not (typically) include account-specific state, which is returned as part of
+; `GetAccountInfo` instead.
+;
+; The "name" and "metadata" fields are required. Other fields are optional, and can be omitted if
+; the module implementation does not support them. The structure supports additional fields for
+; future extensibility. Non-standard fields (i.e. any fields that are not defined by a standard,
+; and are specific to the module implementation) may be included, and their tags should be
+; prefixed with an underscore ("_") to distinguish them as such.
```
soerenbf, line 179: I don't think I understand the distinction made between "standard" and "specific to module implementation" here. Are we not supposed to be concrete in the CBOR schema, i.e. _not_ allow flexibility? I feel like the "flexibility" part is already covered by choosing to CBOR encode the parts of the PLT data that is variable between token modules.

Maybe I'm misunderstanding something here, but I would imagine 1-1 relationship between token modules and CBOR schemas.

cbor-schema/plt.cddl
```diff
@@ -163,3 +163,34 @@ token-initialization-parameters = {
...
+    ? "allowList": bool,
+    ; Whether the token supports a deny list.
+    ? "denyList": bool,
+    ; Whether the token is mintable.
+    ? "mintable": bool,
+    ; Whether the token is burnable.
+    ? "burnable": bool,
+    ; Additional state information may be provided under further text keys, the meaning
+    ; of which are not defined the present specification.
+    * tstr => any
```
soerenbf, line 195: To me, this defeats the purpose of having the schema in the first place (related to the comment above)

cbor-schema/plt.cddl
soerenbf: After discussion in office, the conclusion is that the **cddl defines a plt standard** which can have **different implementations of the standard, i.e. plt module versions**.

In that case, it makes sense to me that there is support for arbitrary fields for different implementations.

td202: My intention is that the schema defines an interface for a PLT module that can have multiple implementations. The schema can allow for implementation defined behaviour (adding non-standard fields), but is also intended to still allow for an extension to the schema in a subsequent standard.

cbor-schema/plt.cddl
```diff
@@ -163,3 +163,34 @@ token-initialization-parameters = {
...
+    ; A URL pointing to the token metadata
+    "metadata": text,
+    ; Whether the token supports an allow list.
+    ? "allowList": bool,
+    ; Whether the token supports a deny list.
+    ? "denyList": bool,
+    ; Whether the token is mintable.
+    ? "mintable": bool,
+    ; Whether the token is burnable.
+    ? "burnable": bool,
```
drsk0, line 192: do we expect non-mintable/burnable tokens?

cbor-schema/plt.cddl
```diff
@@ -163,3 +163,34 @@ token-initialization-parameters = {
...
+    ; Additional state information may be provided under further text keys, the meaning
+    ; of which are not defined the present specification.
```
drsk0, line 194: typo: "defined _in_ the present specification"

cbor-schema/plt.cddl
drsk0, line 195: what does 'tstr' stand for?

cbor-schema/plt.cddl
```diff
@@ -163,3 +163,34 @@ token-initialization-parameters = {
...
+; The "name" and "metadata" fields are required. Other fields are optional, and can be omitted if
+; the module implementation does not support them. The structure supports additional fields for
+; future extensibility. Non-standard fields (i.e. any fields that are not defined by a standard,
+; and are specific to the module implementation) may be included, and their tags should be
+; prefixed with an underscore ("_") to distinguish them as such.
+token-module-state = {
+    ; The name of the token
+    "name": text,
+    ; A URL pointing to the token metadata
+    "metadata": text,
```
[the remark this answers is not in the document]
td202: You mean, for instance, add a `metadataHash`? I think that's reasonable, but we should also add it when creating a PLT. I have created [an issue](https://linear.app/concordium/issue/COR-1385/support-plt-metadata-hash) for it, rather than address it here.

cbor-schema/plt.cddl
drsk0: I'm also confused about this. If we have _standards_, should those be defined in CBOR at all? Or can we update the standard defined in CBOR without a protocol update?

cbor-schema/plt.cddl
td202: "text string" - it's one of the CDDL primitive types. Since elsewhere we use `text`, which is an alias for `tstr`, I'll switch it to `text` here.

cbor-schema/plt.cddl
td202: I think that's unclear at the moment. In any case, they're so far optional when creating the token.

cbor-schema/plt.cddl
td202: The idea is that eventually we will have WASM-implemented token modules, and we can deploy and update new ones without a protocol update. These can have additional functionality that is specified by a subsequent standard.

soerenbf (approved):
LGTM (at least the cddl part, as haskell does not fully compile in my brain yet)

PR #635, opened 2025-05-06
Update plt cddl with module specific event details

soerenbf:
## Purpose

Adds event detail definitions to the plt token module cddl

## Changes

- Factors out the details of each operation to be able to re-use them in the module specific event details

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cbor-schema/plt.cddl
+++ b/cbor-schema/plt.cddl
@@ -15,17 +15,14 @@ token-holder-operation = token-transfer
 ; (implicit) to the recipient account.
 token-transfer = {
     ; The operation type is "transfer".
-    "transfer": token-transfer-details
-}
-
-; The details of a token transfer operation.
-token-transfer-details = {
-    ; the amount of tokens to transfer.
-    "amount": token-amount,
-    ; the recipient account.
-    "recipient": token-receiver,
-    ; an optional memo.
-    ? "memo": memo
+    "transfer": {
+	; the amount of tokens to transfer.
+	"amount": token-amount,
+	; the recipient account.
+	"recipient": token-receiver,
+	; an optional memo.
+	? "memo": memo
+    }
 }
 
 ; We represent token amounts as a decimal fraction.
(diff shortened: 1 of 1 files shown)


cbor-schema/plt.cddl
```diff
@@ -11,18 +11,24 @@ token-holder-transaction = [ * token-holder-operation ]
...
-    }
+    "transfer": token-transfer-details
+}
+
+; The details of a token transfer operation.
+token-transfer-details = {
+    ; the amount of tokens to transfer.
+    "amount": token-amount,
+    ; the recipient account.
+    "recipient": token-receiver,
```
drsk0, line 29: I guess the sender is always clear from context?

drsk0 (approved):
Looks good.

DOBEN (approved):

cbor-schema/plt.cddl
td202: For the transfer operation, the sender is implicit. However, for the transfer event, I think it should be explicit.

cbor-schema/plt.cddl
```diff
@@ -81,22 +87,26 @@ token-governance-operation = token-mint
     / token-burn
     / token-update-list
 
+; A token governance event, which is the result of executing a token governance operation.
+token-governance-event = token-supply-update-details
+    / token-list-update-details
```
td202, line 92: I don't think we should define a union type for these, as the event `type` string will determine how the details should be parsed.

td202 (changes requested):
I would suggest putting all the events in their own section of the schema, common for both token-holder and governance operations. e.g.

```
; "transfer" event details.
; This is a Token Kernel event, and indicates funds were transferred from the "from" account to the "to" account.
token-event-transfer = {
    "amount": token-amount,
    "from": token-receiver,
    "to": token-receiver,
    ? "memo": memo
}

; "mint" event details.
; This is a Token Kernel event....
token-event-mint = {...}

; "addDenyList" event details.
; This is a Token Module event, and indicates the specified account was added to the deny list.
token-event-add-deny-list = token-list-update-details
```

cbor-schema/plt.cddl
soerenbf: yeah that makes sense, will remove this

cbor-schema/plt.cddl
soerenbf: Why do you think it should be explicitly defined at this level? It's already available from the surrounding structure, is it not? When looking at the corresponding "simple transfer" event, it does not explicitly define the sender either, but rather rely on the surrounding context for this.

limemloh: When we get sponsored transactions and smart contract interaction, the sender might be harder to derive from the context.

cbor-schema/plt.cddl
```diff
@@ -16,12 +16,12 @@ token-holder-operation = token-transfer
...
-        ; The recipient account.
-        "recipient": token-receiver,
-        ; An optional memo.
-        ? "memo": memo
+	; the amount of tokens to transfer.
+	"amount": token-amount,
+	; the recipient account.
+	"recipient": token-receiver,
+	; an optional memo.
+	? "memo": memo
```
td202, line 24: Suggestion: revert

cbor-schema/plt.cddl
```diff
@@ -194,3 +188,24 @@ token-module-state = {
     ; of which are not defined in the present specification.
     * text => any
 }
+
+; Token module events.
+
+; These events are limited to the events emitted by the token module. Events related to protocol-level tokens
+; balance changes are defined from the token kernel.
```
td202, line 195: I suggest de-emphasising the token module/token kernel a bit, and rather say that token balance change events are defined separately and not subject to CBOR encoding.

td202 (approved):

PR #637, opened 2025-05-08
Get token info query

DOBEN:
## Purpose

Related to:
https://github.com/Concordium/concordium-client/pull/356

### DIFF
--- a/haskell-src/Concordium/Types/Queries/Tokens.hs
+++ b/haskell-src/Concordium/Types/Queries/Tokens.hs
@@ -12,11 +12,10 @@ module Concordium.Types.Queries.Tokens (
 import qualified Data.ByteString as BS
 import Data.Word
 
+import Concordium.Crypto.ByteStringHelpers
 import Concordium.Types
 import Concordium.Types.Tokens
 import Data.Aeson as AE
-import qualified Data.ByteString.Base16 as BS16
-import qualified Data.Text.Encoding as Text
 
 -- | Protocol level token.
 data Token = Token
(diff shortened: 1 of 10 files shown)


haskell-src/Concordium/Types/Queries/Tokens.hs
```diff
@@ -77,14 +79,47 @@ data TokenState = TokenState
       -- | The governance account for the token.
       tsIssuer :: !AccountAddress,
       -- | The number of decimals in the token representation.
-      tsDecimals :: !Word8,
+      tsDecimals :: !Word32,
```
soerenbf, line 82: Why this change? As far as I know, the upper bound for token decimals is 255?

haskell-src/Concordium/Types/Queries/Tokens.hs
```diff
@@ -77,14 +79,47 @@ data TokenState = TokenState
...
+
+instance ToJSON HexByteString where
+    toJSON (HexByteString bs) =
+        String (Text.decodeUtf8 (BS16.encode bs))
+
+instance FromJSON HexByteString where
+    parseJSON = AE.withText "HexByteString" $ \txt ->
+        case BS16.decode (Text.encodeUtf8 txt) of
+            Right bs -> pure $ HexByteString bs
+            _ -> fail "Invalid Base16 encoding"
```
td202, line 101: Move this to `ByteStringHelpers` with the rest of these. I'm not sure it's great to have both `ByteStringHex` and `HexByteString`.

haskell-src/Concordium/Types/Queries/Tokens.hs
td202: It should be kept as `Word8`. Protobuf doesn't have a `uint8`, so it's encoded there as a `unit32`, but we require it to be within the 0-255 range.

td202 (changes requested):

haskell-src/Concordium/Types/Queries/Tokens.hs
DOBEN: Moved it and renamed the already existing type`ByteHexString` to `ShortByteHexString`.

haskell-src/Concordium/Types/Queries/Tokens.hs
```diff
@@ -12,9 +12,10 @@ module Concordium.Types.Queries.Tokens (
 import qualified Data.ByteString as BS
 import Data.Word
 
+import Concordium.Crypto.ByteStringHelpers
 import Concordium.Types
 import Concordium.Types.Tokens
-import Data.Aeson (FromJSON (..), ToJSON (..), object, withObject, (.:), (.=))
+import Data.Aeson as AE
```
td202, line 18:
```suggestion
import Data.Aeson
```
I don't see anywhere it's referred to here as `AE`.

td202 (approved):

PR #639, opened 2025-05-09
json encode plt token holder operation

allanbrondum:
## Purpose

Implement JSON serialization of CBOR encoded data inside token holder transaction

## Changes

Implemented JSON serialization for relevant types. Example of token holder transaction
```json
{
    "energy": 563,
    "expiryTime": 1747397416,
    "nonce": 6,
    "payload": {
        "operations": [
            {
                "transfer": {
                    "amount": 1,
                    "memo": {
                        "type": "cbor",
                        "value": "6474657374"
                    },
                    "recipient": {
                        "address": "3CBp9U5nJyLFk8r7Uh3mBHTB23r2pwRj5H3qcTtLtdribh8o1U",
                        "coininfo": "CCD",
                        "type": "account"
                    }
                }
            }
        ],
        "tokenSymbol": "TK1",
        "transactionType": "tokenHolder"
    },
    "signature": {
        "0": {
            "0": "8c59270581bfd9e2ed41ff04f70c48e4811272a61c98264e5b757bf996823d3e95ab4a038e3e16fbd1ea777dab3d8cb0e1ef66ecfa7d0a8ed63eeeb9495b610a"
        }
    },
    "signer": "37GP3ChyTTX2gjHdBrKfaA5tbmEqM53kkxsSHYN4FmsQEsaq8V",
    "version": 1
}
```

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Types.hs
+++ b/haskell-src/Concordium/Types.hs
@@ -201,6 +201,7 @@ module Concordium.Types (
     cpltGovernanceAccount,
     cpltDecimals,
     cpltInitializationParameters,
+    EncodedTokenOperations (..),
 ) where
 
 import Data.Data (Data, Typeable)
@@ -1375,6 +1376,30 @@ instance AE.FromJSON CreatePLT where
         (EncodedTokenInitializationParameters _cpltInitializationParameters) <-
             o AE..: "initializationParameters"
         return CreatePLT{..}
+
+-- | A wrapper type for (de)-serializing an CBOR-encoded token operations to/from JSON.
+--  This can parse either an JSON object representation of 'TokenHolderTransaction'
+--  (which is then re-encoded as CBOR) or a hex-encoded byte string. When rendering JSON,
+--  it will render as a JSON object if the contents can be decoded to a
+-- 'TokenHolderTransaction', or otherwise as the hex-encoded byte string.
+newtype EncodedTokenOperations = EncodedTokenOperations TokenParameter
+
+instance AE.ToJSON EncodedTokenOperations where
+    toJSON (EncodedTokenOperations tp@(TokenParameter sbs)) =
+        case CBOR.tokenHolderTransactionFromBytes
+            (BSBuilder.toLazyByteString $ BSBuilder.shortByteString sbs) of
+            Left _ -> AE.toJSON tp
+            Right v -> AE.toJSON v
+
+-- instance AE.FromJSON EncodedTokenOperations where
+--    parseJSON o@(AE.Object _) = do
+--        tip <- AE.parseJSON o
+--        return $
+--            EncodedTokenOperations $
+--                TokenParameter $
+--                    BSS.toShort $
+--                        CBOR.tokenHolderTransactionToBytes tip
+--    parseJSON val = EncodedTokenOperations <$> AE.parseJSON val
 
 -- Template haskell derivations. At the end to get around staging restrictions.
 $(deriveJSON defaultOptions{sumEncoding = TaggedObject{tagFieldName = "type", contentsFieldName = "address"}} ''Address)
--- a/haskell-src/Concordium/Types.hs
+++ b/haskell-src/Concordium/Types.hs
@@ -202,6 +202,7 @@ module Concordium.Types (
     cpltDecimals,
     cpltInitializationParameters,
     EncodedTokenOperations (..),
+    EncodedTokenInitializationParameters (..),
 ) where
 
 import Data.Data (Data, Typeable)
@@ -1297,6 +1298,7 @@ makeTokenModuleRejectReason tmrrTokenSymbol CBOR.EncodedTokenRejectReason{..} =
 --  it will render as a JSON object if the contents can be decoded to a
 -- 'TokenInitializationParameters', or otherwise as the hex-encoded byte string.
 newtype EncodedTokenInitializationParameters = EncodedTokenInitializationParameters TokenParameter
+    deriving newtype (Eq, Show)
 
 instance AE.ToJSON EncodedTokenInitializationParameters where
     toJSON (EncodedTokenInitializationParameters tp@(TokenParameter sbs)) =
@@ -1383,6 +1385,7 @@ instance AE.FromJSON CreatePLT where
 --  it will render as a JSON object if the contents can be decoded to a
 -- 'TokenHolderTransaction', or otherwise as the hex-encoded byte string.
 newtype EncodedTokenOperations = EncodedTokenOperations TokenParameter
+    deriving newtype (Eq, Show)
 
 instance AE.ToJSON EncodedTokenOperations where
     toJSON (EncodedTokenOperations tp@(TokenParameter sbs)) =
--- a/haskell-src/Concordium/Types.hs
+++ b/haskell-src/Concordium/Types.hs
@@ -1392,14 +1392,15 @@ instance AE.ToJSON EncodedTokenOperations where
             Right v -> AE.toJSON v
 
 instance AE.FromJSON EncodedTokenOperations where
-    parseJSON o@(AE.Object _) = do
-        tip <- AE.parseJSON o
+    parseJSON v@(AE.Array _) = do
+        tip <- AE.parseJSON v
         return $
             EncodedTokenOperations $
                 TokenParameter $
                     BSS.toShort $
                         CBOR.tokenHolderTransactionToBytes tip
-    parseJSON val = EncodedTokenOperations <$> AE.parseJSON val
+    parseJSON v@(AE.String _) = EncodedTokenOperations <$> AE.parseJSON v
+    parseJSON _ = fail "EncodedTokenOperations JSON must be either an array or a string"
 
 -- Template haskell derivations. At the end to get around staging restrictions.
 $(deriveJSON defaultOptions{sumEncoding = TaggedObject{tagFieldName = "type", contentsFieldName = "address"}} ''Address)
--- a/haskell-src/Concordium/Types.hs
+++ b/haskell-src/Concordium/Types.hs
@@ -1391,15 +1391,15 @@ instance AE.ToJSON EncodedTokenOperations where
             Left _ -> AE.toJSON tp
             Right v -> AE.toJSON v
 
--- instance AE.FromJSON EncodedTokenOperations where
---    parseJSON o@(AE.Object _) = do
---        tip <- AE.parseJSON o
---        return $
---            EncodedTokenOperations $
---                TokenParameter $
---                    BSS.toShort $
---                        CBOR.tokenHolderTransactionToBytes tip
---    parseJSON val = EncodedTokenOperations <$> AE.parseJSON val
+instance AE.FromJSON EncodedTokenOperations where
+    parseJSON o@(AE.Object _) = do
+        tip <- AE.parseJSON o
+        return $
+            EncodedTokenOperations $
+                TokenParameter $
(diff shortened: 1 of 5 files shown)


haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
```diff
@@ -592,6 +592,18 @@ encodeTokenTransfer TokenTransferBody{..} =
 newtype TokenHolderOperation = TokenHolderTransfer TokenTransferBody
     deriving (Eq, Show)
 
+instance AE.ToJSON TokenHolderOperation where
+    toJSON (TokenHolderTransfer TokenTransferBody{..}) = do
+        AE.object $
+            [ "amount" AE..= ttAmount,
+              "recipient" AE..= receiverAccountAddress ttRecipient
```
allanbrondum, line 599: I'm not sure how recipient and memo should be encoded to JSON. How should the different variants be encoded and when are they relevant?
```
-- | A destination that can receive and hold protocol-level tokens.
--  Currently, this can only be a Concordium account address.
data TokenReceiver = ReceiverAccount
    { -- | The account address.
      receiverAccountAddress :: !AccountAddress,
      -- | Although the account can only be a Concordium address, this specifies whether the
      --  address type should be explicit in the CBOR encoding.
      receiverAccountCoinInfo :: !(Maybe CoinInfo)
    }

-- | A 'TaggableMemo' represents a 'Memo' that may optionally be tagged as CBOR-encoded.
--  Memos are often assumed to be CBOR-encoded, but the tag can be used to make this explicit.
data TaggableMemo
    = -- | The memo is represented as a byte string with no tag.
      UntaggedMemo {untaggedMemo :: !Memo}
    | -- | The memo is represented as a byte string with a tag indicating CBOR-encoded data.
      CBORMemo {untaggedMemo :: !Memo}

-- | Token transfer operation. This transfers a specified amount of tokens from the sender account
--  (implicit) to the recipient account.
data TokenTransferBody = TokenTransferBody
    { -- | The amount to transfer.
      ttAmount :: !TokenAmount,
      -- | The recipient account address.
      ttRecipient :: !TokenReceiver,
      -- | An optional memo associated with the transfer.
      ttMemo :: !(Maybe TaggableMemo)
    }
```

td202: For `TokenReceiver`, I think we need to have a wrapper (because eventually this could include smart contract addresses). If we want to make it compatible with the JSON serialization of the `Address` type, then we should do something like `{"type":"AddressAccount", "address": "35RkB2UFwMdkNqB3RLtf5oaiF6AjAjK3C3k8n2it727BExTDBQ"}`. (Though I'm not entirely sure we want to do that.) I don't think there's any reason to retain the `CoinInfo` part in the JSON serialization.

For the memo, I think it would make sense to keep the tag information, and possibly try to convert the CBOR to JSON. I think this would need 3 tags, something like: "Untagged" (for the untagged case), "CBOR-Raw" (for the tagged case where it cannot be converted to JSON), and "CBOR-JSON" (for when it can be decoded). In the CBOR-Raw case, it will be represented as a hex-encoded string (also in the Untagged case). In the other case it will be whatever JSON value is decoded. We need to distinguish the cases, because a CBOR value can convert to a JSON string value.

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
```diff
@@ -592,6 +657,19 @@ encodeTokenTransfer TokenTransferBody{..} =
...
+        AE.object
+            [ "transfer" AE..= AE.toJSON body
+            ]
+
+instance AE.FromJSON TokenHolderOperation where
+    parseJSON = AE.withObject "TokenHolderOperation" $ \o -> do
+        transferBodyMaybe <- o AE..:? "transfer"
+        case transferBodyMaybe of
+            Just transferBody -> return $ TokenHolderTransfer transferBody
+            _ -> fail "Unknown TokenHolderOperation type"
```
DOBEN, line 671:
```suggestion
        transferBody <- o AE..: "transfer"
        pure $ TokenHolderTransfer transferBody
```

DOBEN:
The `Cbor` value is displayed as `hex-encoded string` currently. As Thomas pointed out, we might want to look into how to decode it further into JSON if possible.

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
```diff
@@ -366,6 +366,13 @@ tokenInitializationParametersToBytes =
 data CoinInfo = CoinInfoConcordium
     deriving (Eq, Show)
 
+instance AE.ToJSON CoinInfo where
+    toJSON CoinInfoConcordium = AE.String "CCD"
+
+instance AE.FromJSON CoinInfo where
+    parseJSON (AE.String "CCD") = return CoinInfoConcordium
```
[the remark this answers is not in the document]
allanbrondum: The symbol "CCD" is taken from the list https://github.com/satoshilabs/slips/blob/master/slip-0044.md referenced by https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-007-hdkey.md#cddl-for-coin-info

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
```diff
@@ -366,6 +366,13 @@ tokenInitializationParametersToBytes =
 data CoinInfo = CoinInfoConcordium
     deriving (Eq, Show)
 
+instance AE.ToJSON CoinInfo where
+    toJSON CoinInfoConcordium = AE.String "CCD"
```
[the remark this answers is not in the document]
allanbrondum: I agree it seems a bit confusing. Though it is an identifier for a crypto coin where the receiver address is resolved. Or rather is is a chain where the address is resolved, but the chain is identified by its native coin according to https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-009-address.md

allanbrondum:
> The `Cbor` value is displayed as `hex-encoded string` currently. As Thomas pointed out, we might want to look into how to decode it further into JSON if possible.

Is it the memo CBOR? We don't have a schema for that? So mapping it to and from JSON would not really be possible?

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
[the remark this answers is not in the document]
allanbrondum: What do you think @td202?

DOBEN:
> > The `Cbor` value is displayed as `hex-encoded string` currently. As Thomas pointed out, we might want to look into how to decode it further into JSON if possible.
> 
> Is it the memo CBOR? We don't have a schema for that? So mapping it to and from JSON would not really be possible?

Yes, the memo can be CBOR.
E.g. at the front-end we display it if possible CBOR decoded (here a number decoded but could be more complex objects; meaning something more human-readable JSON like)
https://ccdscan.io/transactions?dcount=1&dentity=transaction&dhash=6ec08f22b1a7b812166154b68dd82b383f54c91a8bd9b8845a830b6dc7491b61

but if the decoding is not working we display the byes as a hex string:
https://ccdscan.io/transactions?dcount=1&dentity=transaction&dhash=05f8e7946d84163f2533ac57d6bab4339c4736a85634062527d2207445ceb076

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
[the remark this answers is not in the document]
td202: I don't think it matters too much. I think either `CCD` or `Concordium` would be fine, as they are both unique in the [coin table](https://github.com/satoshilabs/slips/blob/master/slip-0044.md). `ccd` is also used for our namespace for [CAIP-2](https://namespaces.chainagnostic.org/ccd/caip2), so maybe that's a good option.  I would consider making parsing case insensitive.

td202:
Regarding the CBOR decoding for the memo, per my first comment, I think we need to handle three cases to ensure that JSON encoding then decoding is the identity:
1. It's not tagged as CBOR. `{"type": "raw", "value": "006a3e5d"}`
2. It's tagged as CBOR, and we can decode it to a JSON value. `{"type": "cbor", "value": {"foo": "bar"}}`
3. It's tagged as CBOR, but we can't decode it to a JSON value. `{"type": "cbor-raw", value": "A10163666F6F"}`

allanbrondum:
> Regarding the CBOR decoding for the memo, per my first comment, I think we need to handle three cases to ensure that JSON encoding then decoding is the identity:
> 
> 1. It's not tagged as CBOR. `{"type": "raw", "value": "006a3e5d"}`
> 2. It's tagged as CBOR, and we can decode it to a JSON value. `{"type": "cbor", "value": {"foo": "bar"}}`
> 3. It's tagged as CBOR, but we can't decode it to a JSON value. `{"type": "cbor-raw", value": "A10163666F6F"}`

I think case 2./3. can get really complex. We first need to define what it evens means for CBOR to be decodable to JSON. I don't think it is obvious? Also, preserving the same bitwise CBOR encoding will be very challenging if not impossible. So at least, would prefer not to do it in this PR.

haskell-src/Concordium/Types.hs
```diff
@@ -1376,6 +1377,31 @@ instance AE.FromJSON CreatePLT where
...
+instance AE.FromJSON EncodedTokenOperations where
+    parseJSON v@(AE.Array _) = do
+        tip <- AE.parseJSON v
+        return $
+            EncodedTokenOperations $
+                TokenParameter $
+                    BSS.toShort $
+                        CBOR.tokenHolderTransactionToBytes tip
+    parseJSON v@(AE.String _) = EncodedTokenOperations <$> AE.parseJSON v
+    parseJSON _ = fail "EncodedTokenOperations JSON must be either an array or a string"
```
td202, line 1403: Please add some tests for this.

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
```diff
@@ -501,9 +527,33 @@ data TaggableMemo
     = -- | The memo is represented as a byte string with no tag.
       UntaggedMemo {untaggedMemo :: !Memo}
     | -- | The memo is represented as a byte string with a tag indicating CBOR-encoded data.
-      CBORMemo {untaggedMemo :: !Memo}
+      CBORMemo {cborMemo :: !Memo}
```
td202, line 530: Changing this will be a breaking change for `concordium-node`. Personally, I would leave it, but if you really want to change it, then please open a PR to fix it there.

td202 (approved):
I'd like to see some tests for the the new instances. Otherwise, looks good.

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
allanbrondum: I believe I checked this, in `concordium-node` the field is used without naming it. So there was no compile error after the change

td202: It's used here: https://github.com/Concordium/concordium-node/blob/72d5708184be1d30a95db90a982751e1ca90067b/concordium-consensus/src/Concordium/Scheduler/ProtocolLevelTokens/Module.hs#L141

PR #647, opened 2025-05-21
Consistent cddl naming.

soerenbf:
Apply consistent naming of types and field names to cddl

- values are kebab-case, as this seems to be the standard
- field names are quoted camelCase to make it unambiguous that we're dealing with a map of string-keyed values

We want to represent string keys with camelCase as this makes it easier to use dynamic values decoded from CBOR in javascript (through dot notation, which is not possible when field names are kebab-cased).

## Checklist

      hard-to-understand areas.


soerenbf:
> Could you also update the corresponding kebab case in CBOR.hs?

Will do. Don't know if there's also rust code in base for cbor encoding/decoding, but if that's the case I will update that as well.

allanbrondum:
> > Could you also update the corresponding kebab case in CBOR.hs?
> 
> Will do. Don't know if there's also rust code in base for cbor encoding/decoding, but if that's the case I will update that as well.

There is no Rust code yet doing that, I'm working on it though

allanbrondum (approved):

PR #648, opened 2025-05-21
Add metadata url type to cddl

soerenbf:
Add a type describing metadata urls which are accompanied by an optional checksum.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cbor-schema/plt.cddl
+++ b/cbor-schema/plt.cddl
@@ -165,8 +165,10 @@ sha256-hash = bytes .size(32)
 metadata-url = {
     ; A string field representing the URL
     "url": text,
-    ; An optional 32-byte checksum value tied to the content of the URL
-    ? "checksum": sha256-hash
+    ; An optional sha256 checksum value tied to the content of the URL
+    ? "checksumSha256": sha256-hash
+    ; Additional fields may be included for future extensibility, e.g. another hash algorithm.
+    * text => any
 }
 
 ; Token state
--- a/cbor-schema/plt.cddl
+++ b/cbor-schema/plt.cddl
@@ -164,7 +164,7 @@ hash = bytes .size(32)
 ; Type alias for metadata with a URL and checksum
 metadata-url = {
     ; A string field representing the URL
-    "value": string,
+    "url": text,
     ; An optional 32-byte checksum value tied to the content of the URL
     ? "checksum": hash
 }
--- a/cbor-schema/plt.cddl
+++ b/cbor-schema/plt.cddl
@@ -145,7 +145,7 @@ token-initialization-parameters = {
     ; The name of the token
     "name": text,
     ; A URL pointing to the token metadata
-    "metadata": text,
+    "metadata": metadata-url,
     ; Whether the token supports an allow list
     ? "allowList": bool .default false,
     ; Whether the token supports a deny list
@@ -156,6 +156,17 @@ token-initialization-parameters = {
     ? "mintable": bool .default false,
     ; Whether the token is burnable
     ? "burnable": bool .default false
+}
+
+; Type alias for a 32-byte hash value
+hash = bytes .size(32)
+
+; Type alias for metadata with a URL and checksum
+metadata-url = {
+    ; A string field representing the URL
+    "value": string,
+    ; An optional 32-byte checksum value tied to the content of the URL
+    ? "checksum": hash
 }
 
 ; Token state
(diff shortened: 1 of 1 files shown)


DOBEN (approved):

drsk0 (approved):

cbor-schema/plt.cddl
```diff
@@ -158,6 +158,17 @@ token-initialization-parameters = {
     ? "burnable": bool .default false
 }
 
+; Type alias for a 32-byte hash value
+hash = bytes .size(32)
+
+; Type alias for metadata with a URL and checksum
+metadata-url = {
+    ; A string field representing the URL
+    "value": string,
```
td202, line 167:
```suggestion
    "url": string,
```

cbor-schema/plt.cddl
```diff
@@ -158,6 +158,17 @@ token-initialization-parameters = {
...
+    ; An optional 32-byte checksum value tied to the content of the URL
+    ? "checksum": hash
```
td202, line 169: I think this should be more explicit in how it's calculated. e.g. this is the SHA-256 hash of the URL contents.

cbor-schema/plt.cddl
```diff
@@ -175,7 +186,7 @@ token-module-state = {
     ; The name of the token
     "name": text,
     ; A URL pointing to the token metadata
-    "metadata": text,
+    "metadata": metadata-url,
```
td202, line 189: I am not 100% sure we should use a map for this, rather than having an additional key e.g. "metadata-sha256". This might be more future compatible if we want to use different hashing schemes in future (for instance, if someone breaks SHA-256).

soerenbf: Could also provide a hashing scheme with the `hash`, i.e.

```
hash = {
  "value": bytes .size(32),
  "scheme": "sha256",
}

metadata-url = {
  "url": text,
  "checksum": hash,
}
```

What do you think?

cbor-schema/plt.cddl
soerenbf: Also changed `string` to `text`

cbor-schema/plt.cddl
limemloh: We should probably not assume that the checksum is 32 bytes, if we want to be future-proof.
I like the current approach, since we can easily extend with more/different checksums for the metadata url.

limemloh (approved):

cbor-schema/plt.cddl
```diff
@@ -158,6 +158,17 @@ token-initialization-parameters = {
     ? "burnable": bool .default false
 }
 
+; Type alias for a 32-byte hash value
+hash = bytes .size(32)
```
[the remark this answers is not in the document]
soerenbf: Good idea, would go well with a discriminant field describing the algorithm used

cbor-schema/plt.cddl
soerenbf: I feel like we should either be specific from the start:

```
sha256-hash = bytes .size(32)
sha512-hash = bytes .size(64)

metadata-url = { 
  "url": text, 
  ? "checksum-sha256": sha256-hash, 
  ? "checksum-sha512": sha512-hash, ; added down the line when needed
  ...
}
```

Or add to a union:

```
hash = sha256-hash / sha512-hash

sha256-hash = {
  "value": bytes .size(32),
  "scheme": "sha256",
}

sha512-hash = {
  "value": bytes .size(64),
  "scheme": "sha512",
}

metadata-url = {
  "url": text,
  ? "checksum": hash,
}
```

I'm leaning towards the latter, as I feel like it's easier to work with at the consumer level. I don't have a strong preference though. Maybe you can elaborate on why you don't like a map approach @td202 ?

cbor-schema/plt.cddl
soerenbf: Resolving this, as the same discussion is taking place in https://github.com/Concordium/concordium-base/pull/648#discussion_r2100474810

cbor-schema/plt.cddl
soerenbf: After discussion with @td202 we landed on:

```
; Type alias for a sha256 hash
sha256-hash = bytes .size(32)

; Type alias for metadata with a URL and checksum
metadata-url = {
    ; A string field representing the URL
    "url": text,
    ; An optional sha256 checksum value tied to the content of the URL
    ? "checksumSha256": sha256-hash
    ; Additional fields may be included for future extensibility, e.g. another hash algorithm.
    * text => any
}
```

PR #650, opened 2025-05-22
Replace PLT token symbol with token ID in haskell and rust src

limemloh:
## Purpose

Consistently use token ID


allanbrondum (approved):

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
```diff
@@ -530,6 +530,11 @@ data TaggableMemo
       CBORMemo {cborMemo :: !Memo}
     deriving (Eq, Show)
 
+-- | Unwrap the 'TaggableMemo' into the inner 'Memo'
+taggableMemoInner :: TaggableMemo -> Memo
+taggableMemoInner UntaggedMemo{..} = untaggedMemo
+taggableMemoInner CBORMemo{..} = cborMemo
+
```
td202, line 537: I find this irritating, but whatever, it's fine.

td202 (approved):

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
limemloh: You rather want `untaggedMemo` in both variants?

PR #662, opened 2025-06-06
Add authorization for CreatePLT transaction

eb-concordium:
## Purpose

Relates to COR-701. Adds  authorization for the `CreatePLT` chain update.

## Changes

- `Authorizations` data structure extended with a `asCreatePLT` field, holding those level 2 keys that can authorize an `CreatePLT` chain update.
- `AuthorizationsVersion` extended with `AuthorizationsVersion2` (AUV2)
- `AuthorizationsVersion` and `ChainParamtersVersion` are now independent, so that `AuthorizationsVersion` is not determined by `ChainParamtersVersion` anymore, but by the protocol version directly (like `ChainParamtersVersion` is)
- Type family `IsCompatibleAuthorizationsVersion` introduced to constrain on the type level that an `AuthorizationsVersion` and `ChainParamtersVersion` are compatible.
- GRPC module updated to cover an `Authorizations 'AuthorizationsVersion2` into the protobuf type `AuthorizationsV1` with the newly introduced optional field `create_plt` set (whereas it is not set for `Authorizations 'AuthorizationsVersion1`)
- RootUpdate payload extended with `Level2KeysRootUpdateV2` for AUV2
- Level1Update payload extended with `Level2KeysLevel1UpdateV2` for AUV2

## Checklist

      hard-to-understand areas.


haskell-tests/Types/UpdatesSpec.hs
```diff
@@ -148,8 +151,7 @@ tests = parallel $ do
     versionedTests SP6
     versionedTests SP7
     versionedTests SP8
-    -- TODO: (NOD-701) Re-enable tests
-    xdescribe "P9 tests" $ versionedTests SP9
+    describe "P9 tests" $ versionedTests SP9
```
td202, line 154:
```suggestion
    versionedTests SP9
```

rust-src/concordium_base/src/updates.rs
```diff
@@ -426,6 +426,8 @@ pub struct AuthorizationsV1 {
     pub cooldown_parameters: AccessStructure,
     /// Keys for changing the lenghts of the reward period.
     pub time_parameters:     AccessStructure,
+    /// Keys for creating a protocol level token
+    pub create_plt:          AccessStructure,
```
td202, line 430: Shouldn't this be an Option?
```suggestion
    pub create_plt:          Option<AccessStructure>,
```

Abutalif: I am not sure this part. I used the same approach as in other `AccessStructures` - make it not optional, use `require()?` for when converting from generated types.
Where can I find out what is the correction approach from the business point of view?

eb-concordium: Thomas is right. It should be an Option, as we have discussed as well

PR #664, opened 2025-06-06
Ar/cor 1458 concordium base deterministic cbor map encoding rust refactor

allanbrondum:
## Purpose

Refactor CBOR serialization and deserialization traits to express the following structure/invariants:
1) `CborDeserialize::deserialize` can only deserialize a single CBOR data item. Same for `CborSerialize::serialize`.
2) A CBOR map consists of a number of entries prefixed by a header. Same for CBOR array

No behaviour is changes as such. The changes are made to facilitate deterministic map encoding (entry order) which relies on the CBOR encoder having this structure available. I generally also makes implementing serializers and deserializers safer, the compiler can help you check the comply with the expected CBOR structure.

## Changes

Ad 1): `CborDeserialize::deserialize` and `CborSerialize::serialize` and all `decode_xxx`/`encode_xxx` functions on `CborDecoder` and `CborEncoder` now take ownership of the decoder/encoder (instead of using it by reference), which means it can only be used once to call these functions.
Ad 2): `CborDecoder::decode_map` now returns a `CborMapDecoder` that decoder the map entries. Likewise for encoding.

Also, the `cbor` module has been split into multiple files.

## Checklist

      hard-to-understand areas.


rust-src/concordium_base/src/common/cbor/decoder.rs
allanbrondum, line 1: Moved here from `cbor` module. The new functionality is only `MapDecoder` and `ArrayDecoder`

rust-src/concordium_base/src/common/cbor/encoder.rs
allanbrondum, line 1: Moved here from `cbor` module. The new functionality is only `MapEncoder` and `ArrayEncoder`

allanbrondum:
> Everything public could use a doc comment, which seems like it's missing for a bunch of things that were moved from the `cbor` module.

Added docs where it was missing

> 
> I also think it would be good with some tests covering the map/array encoding/decoding, but maybe this is covered in a subsequent PR?

The map and array decoding is tested in `cbor::test`. it was already implemented and tested, this PR just changed to decoding/encoding traits to have `CborMapEncoder/Decoder` and `CborArrayEncoder/Decoder`

soerenbf (approved):

PR #672, opened 2025-06-16
Ar/cor 1458 Deterministic CBOR map entry order

allanbrondum:
## Purpose

Implement deterministic ordering of CBOR map entries

## Changes

## Checklist

      hard-to-understand areas.


rust-src/concordium_base/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Unreleased
+
+- `MetadataUrl` and `TokenModuleState` now supports decoding "additional" data matching the CDDL rule `* text => any`.
```
allanbrondum, line 3: From previous PR

rust-src/concordium_base/src/common/cbor/encoder.rs
```diff
@@ -2,74 +2,85 @@ use crate::common::cbor::{
     CborArrayEncoder, CborEncoder, CborMapEncoder, CborSerializationError, CborSerializationResult,
     CborSerialize,
 };
-use ciborium_io::Write;
 use ciborium_ll::Header;
+use std::{
+    io::Write,
```
allanbrondum, line 7: Change trait used to `std::io::Write` since it has the blanket implementation `std::io::Write for &mut impl std::io::Write`

rust-src/concordium_base/src/common/cbor/encoder.rs
```diff
@@ -151,3 +173,41 @@ where
...
+        let mut bytes = Vec::new();
+        let mut encoder = Encoder::new(&mut bytes);
+        let mut map_encoder = encoder.encode_map(3).unwrap();
+        map_encoder.serialize_entry(&256, &0).unwrap();
+        map_encoder.serialize_entry(&(256 * 256), &0).unwrap();
+        map_encoder.serialize_entry(&1, &0).unwrap();
+        map_encoder.end().unwrap();
+
+        assert_eq!(hex::encode(&bytes), "a30100190100001a0001000000");
+    }
```
soerenbf, line 212: I think it would be better if we tested with keys where the lexical ordering is different from the numeric ordering, as here they are the same, i.e. in both cases they will be `1 < 256 < 65536`. Also would avoid the `256*256` as it's just extra overhead when looking through the test to verify the ordering manually 😛 

And for good measure, I think it would make sense to have a test with string keys as well.

soerenbf (approved):
Looks good, just one small thing

PR #673, opened 2025-06-17
Use extendable tokenHolder types in plt events (COR-1506)

DOBEN:
## Purpose

We want to ensure the Protocol-level token (PLT) events (transfer, mint and burn) are structured in a forward-compatible way, such that we can expand the events in the future, with minimal breakage in the ecosystem. Specifically adding new token holder entities as there are already discussions on adding smart contract support. Here we need to consider the additional context as well, since memos might not be relevant for smart contracts, but instead a receive-hook name with some parsed input parameters would be needed in the events.

https://linear.app/concordium/issue/COR-1506/align-json-serialization-of-plt-events-across-rusthaskell-and-adhere

Related PRs:
https://github.com/Concordium/concordium-client/pull/382
https://github.com/Concordium/concordium-node/pull/1402
https://github.com/Concordium/concordium-rust-sdk/pull/275
https://github.com/Concordium/concordium-transaction-logger/pull/34
https://github.com/Concordium/concordium-wallet-proxy/pull/132

## Changes

- Clearly separate between the `CborTokenHolder` type (used in the transaction payload, in reject reasons, and in the `TokenModuleEvent`) from the `TokenHolder` type (emitted by the plt mint/burn/transfer events as defined in the proto buffer definition).

- The  plt mint/burn/transfer events are extendable now with additional optional fields in the future and the `tokenHolder` type can be extended to e.g. smart contracts. The event `add/remove account address from allow/deny list` (`TokenModuleEvent`) is not extendable but as `TokenModuleEvents` itself can be extended as their `details` part can be any cbor encoded future event that trade-off seems reasonable.

- Add a bitmask to each plt mint/burn/transfer event, so that we can toggle on/off the optional `memo` field and introduce additional optional fields in the future for smart contract token holder support.

- Align JSON serialization of the `TokenHolder` type emitted by the plt mint/burn/transfer event across the `haskell/rust` code base. Align `CborTokenHolder` type. Align `updateCreatePLT` transaction tag across the code base.

### DIFF
--- a/haskell-src/Concordium/Types.hs
+++ b/haskell-src/Concordium/Types.hs
@@ -182,6 +182,7 @@ module Concordium.Types (
 
     -- * Protocol-level tokens
     TokenId (..),
+    TokenHolderEvent (HolderAccountEvent),
     makeTokenId,
     unsafeGetTokenId,
     TokenParameter (..),
@@ -411,6 +412,47 @@ instance AE.FromJSON UrlText where
 
 emptyUrlText :: UrlText
 emptyUrlText = UrlText ""
+
+newtype TokenHolderEvent
+    = HolderAccountEvent AccountAddress
+    deriving (Eq)
+
+instance Show TokenHolderEvent where
+    show (HolderAccountEvent addr) =
+        show addr
+
+instance AE.ToJSON TokenHolderEvent where
+    toJSON (HolderAccountEvent address) = do
+        AE.object
+            [ -- Tag with type of `account`.
+              "type" AE..= AE.String "account",
+              "address" AE..= address
+            ]
+
+instance AE.FromJSON TokenHolderEvent where
+    parseJSON = AE.withObject "TokenHolderEvent" $ \o -> do
+        type_string <- o AE..: "type"
+        case (type_string :: String) of
+            "account" -> do
+                address <- o AE..: "address"
+                return (HolderAccountEvent address)
+            _ -> fail ("Unknown TokenHolderEvent type " ++ type_string)
+
+instance S.Serialize TokenHolderEvent where
+    get = getHolderAccount
+    put = putHolderAccount
+
+getHolderAccount :: S.Get TokenHolderEvent
+getHolderAccount =
+    S.getWord8 >>= \case
+        0 -> do
+            HolderAccountEvent <$> S.get
+        n -> fail $ "Unrecognized TokenHolderEvent tag: " ++ show n
+
+putHolderAccount :: S.Putter TokenHolderEvent
+putHolderAccount (HolderAccountEvent address) =
+    S.putWord8 0
+        <> S.put address
 
 -- | Due to limitations on the ledger, there has to be some restriction on the
 --  precision of the input for updating ElectionDifficult. For this purpose,
--- a/haskell-src/Concordium/Types.hs
+++ b/haskell-src/Concordium/Types.hs
@@ -182,7 +182,7 @@ module Concordium.Types (
 
     -- * Protocol-level tokens
     TokenId (..),
-    TokenHolderEvent (HolderAccountEvent),
+    TokenHolder (HolderAccount),
     makeTokenId,
     unsafeGetTokenId,
     TokenParameter (..),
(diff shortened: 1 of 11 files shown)


rust-src/concordium_base/src/protocol_level_tokens/token_event.rs
```diff
@@ -89,17 +90,24 @@ pub enum TokenModuleEventType {
 #[serde(rename_all = "camelCase")]
 pub struct TokenListUpdateEventDetails {
     /// The account that was added or removed from an allow or deny list
-    pub target: TokenHolder,
+    pub target: CborTokenHolder,
+}
+
+#[derive(Debug, Eq, PartialEq, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(tag = "type")]
+pub enum TokenHolderEvent {
```
allanbrondum, line 98: This is named as an event. If it is a token holder in an event, it should be named `EventTokenHolder` (unless there is a better name)

DOBEN: i am looking forward once we have no `tokenHolder` transaction type anymore in the next iteration. At that transaction type also cause confusion.

allanbrondum: Looking more closely at it, I'm not sure we should have this type. In the CDDL, there is only one token holder type. I only think we need one in the rust code also. What we need is rather a custom JSON serialization. You can do it still using derive like this
```rust
impl serde::Serialize for TokenHolder {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer
    {
        #[derive(serde::Serialize)]
        enum TokenHolderJson<'a> {
            #[serde(rename = "account")]
            HolderAccountEvent { address: &'a AccountAddress },
        }
        // map self to TokenHolderJson
        todo!()
    }
}
```

haskell-src/Concordium/Types.hs
```diff
@@ -412,6 +413,47 @@ instance AE.FromJSON UrlText where
 emptyUrlText :: UrlText
 emptyUrlText = UrlText ""
 
+newtype TokenHolderEvent
```
allanbrondum, line 416: Same thoughts about the naming as for the Rust code

allanbrondum: Can't we just use the already existing type `TokenHolder`? And then in the transfer operation, if that operation only works for accounts, then change the type used there from `TokenHolder` to a type specific for accounts (the single variant of `TokenHolder`):
```
data TokenTransferBody = TokenTransferBody
    { -- | The amount to transfer.
      ttAmount :: !TokenAmount,
      -- | The recipient account address.
      ttRecipient :: !TokenHolder,
      -- | An optional memo associated with the transfer.
      ttMemo :: !(Maybe TaggableMemo)
    }
    deriving (Eq, Show)
```
There seems to be a bit mismatch in the `TokenTransferBody` definition currently. Seems like `ttRecipient` should be `AccountTokenHolder` (type does not exist yet)

allanbrondum: Seems like we resolved the discussion of one or two types, so just the naming remains, maybe we can call it `TokenHolder` instead of `TokenHolderEvent`?

rust-src/concordium_base/src/protocol_level_tokens/token_event.rs
```diff
@@ -3,8 +3,9 @@ use crate::{
     transactions::Memo,
 };
 use concordium_base_derive::{CborDeserialize, CborSerialize};
+use concordium_contracts_common::AccountAddress;
 
-use super::{cbor::RawCbor, TokenAmount, TokenHolder, TokenId};
+use super::{cbor::RawCbor, TokenAmount, TokenHolder as CborTokenHolder, TokenId};
```
allanbrondum, line 8: I suggest to rename `TokenHolder` to `CborTokenHolder` in the Rust code also

allanbrondum (approved):

rust-src/concordium_base/src/protocol_level_tokens/token_event.rs
DOBEN: The `TokenModuleEvents (which when decoded include the `CborTokenHolder` type) haven't been tracked by `transaction-logger` indexer yet. Meaning we haven't discovered/tested yet any JSON serialization misalignment between haskell/rust.

rust-src/concordium_base/src/protocol_level_tokens/token_holder.rs
```diff
@@ -26,7 +26,7 @@ const CONCORDIUM_SLIP_0044_CODE: u64 = 919;
 )]
 #[serde(rename_all = "camelCase")]
 #[cbor(tagged)]
-pub enum TokenHolder {
```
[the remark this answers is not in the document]
DOBEN: I suspect that the `ccdscan` indexer might already use that serde serialization, so don't think we can remove it.

haskell-src/Concordium/Types.hs
```diff
@@ -412,6 +413,47 @@ instance AE.FromJSON UrlText where
 emptyUrlText :: UrlText
 emptyUrlText = UrlText ""
 
+newtype TokenHolder
+    = HolderAccount AccountAddress
+    deriving (Eq)
+
+instance Show TokenHolder where
+    show (HolderAccount addr) =
+        show addr
```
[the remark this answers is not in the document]
DOBEN: The custom show is in place because it gives a better result in the wallets and `concordium-client`.

Custom `show`:

```
1.00 PLT burned from 3VLYFiMYHxx62ym2Ze6iNEha5hWpyzXVAK6PSA2wEQPsEV9yh6
```

vs

derived `show`:

```
1.00 PLT burned from HolderAccount {thAccount = "3VLYFiMYHxx62ym2Ze6iNEha5hWpyzXVAK6PSA2wEQPsEV9yh6"}
```

haskell-src/Concordium/Types/Execution.hs
```diff
@@ -1923,18 +1932,30 @@ getEvent spv =
             etmeDetails <- S.get
             return TokenModuleEvent{..}
         39 | supportPlt -> do
+            -- The purpose of the bitmap is to make the event extendable with optional additional fields in the future.
+            bitmap <- S.getWord8
```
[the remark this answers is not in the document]
DOBEN: Very good point.

PR #682, opened 2025-06-25
Parallel module build

allanbrondum:
## Purpose

Parallelize package build (internally)

## Changes

Set `-j` option on GHC to enable compiling modules within the package in parallel.
See https://downloads.haskell.org/ghc/latest/docs/users_guide/flags.html#miscellaneous-compiler-options

## Checklist

      hard-to-understand areas.


allanbrondum:
Build servers does not seem to have the capacity that this change has any positive effect

PR #692, opened 2025-07-01
Add token-pause operation to plt.cddl

soerenbf:
Adds the `token-pause` along with a corresponding field on `token-module-state` operation to `plt.cddl`.

One important thing to note here, is that I did not include a reject reason specifically for this. My thought is that it would be fine to use the existing "reject-details-unsupported-operation" for this based on the assumption that it is not intended only for signalling operations being **permanently unsupported**.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cbor-schema/plt.cddl
+++ b/cbor-schema/plt.cddl
@@ -201,7 +201,7 @@ token-module-state = {
     ; Whether the token is burnable.
     ? "burnable": bool,
     ; Whether the execution of certain token operations has been paused.
-    ? "paused": bool,
+    "paused": bool,
     ; Additional state information may be provided under further text keys, the meaning
     ; of which are not defined in the present specification.
     * text => any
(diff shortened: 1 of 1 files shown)


limemloh (approved):
Looks good to me

cbor-schema/plt.cddl
```diff
@@ -129,12 +129,19 @@ token-list-update-details = {
     "target": token-holder
 }
 
+; Pause or unpause the execution of certain token operations. This includes
+; the operations "mint", "burn", and "transfer".
+token-pause = {
```
allanbrondum, line 134: Should this be added to the operation list?

soerenbf: Good catch 👍

drsk0 (approved):
Looks great!

cbor-schema/plt.cddl
```diff
@@ -129,12 +130,19 @@ token-list-update-details = {
...
+    ; Whether to pause or unpause token operations.
+    "pause": bool
```
td202, line 137: I think the body of all operations should be a map. I would suggest having separate `pause` and `unpause` operations with the empty map as the body.

soerenbf: 🙈 

I had a feeling you'd say that.. I'll draw up another proposal, and we can start a discussion there.

PR #718, opened 2025-07-24
Ar/cor 1678 cbor decoding fails on empty string rust

allanbrondum:
## Purpose

Fix CBOR decoding empty string

## Changes

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/src/common/cbor/primitives.rs
+++ b/rust-src/concordium_base/src/common/cbor/primitives.rs
@@ -414,6 +414,16 @@ mod test {
     }
 
     #[test]
+    fn test_bytes_empty() {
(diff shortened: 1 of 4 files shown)


rust-src/concordium_base/src/common/cbor/primitives.rs
```diff
@@ -447,6 +457,16 @@ mod test {
         assert_eq!(text_decoded, text);
     }
 
+    #[test]
+    fn test_text_empty() {
```
allanbrondum, line 461: This is the only test that actually failed before the fix. The other tests where added to cover the empty case for other major types than byte strings

td202 (approved):

PR #733, opened 2025-08-22
Ar/cor 1646 support decoding indefinite length cbor character strings

allanbrondum:
## Purpose

Support decoding indefinite length CBOR byte and text strings

## Changes

## Checklist

      hard-to-understand areas.


rust-src/concordium_base/src/common/cbor/primitives.rs
```diff
@@ -744,22 +744,29 @@ mod test {
...
+            "err: {}",
+            err
+        );
     }
 
-    /// Test where CBOR is not well-formed: Bytes length in header does not
-    /// match actual data. Test that we get an error and don't panic
     #[test]
-    fn test_bytes_length_invalid() {
-        let cbor = hex::decode("58ff0102030405").unwrap();
```
allanbrondum, line 754: moved to decoder test

squirer (approved):

rust-src/concordium_base/src/common/cbor/decoder.rs
```diff
@@ -213,53 +218,101 @@ where
...
     }
 
-    /// Decodes text data item into given destination. Length of text data item
-    /// must match the destination length.
-    ///
-    /// This function works only for text data items of definite length (which
-    /// means there is a single segment)
-    fn decode_definite_length_text(&mut self, dest: &mut [u8]) -> CborSerializationResult<()>
+    /// Decodes text data item into given destination.
+    fn decode_text_impl(
```
td202, line 296: From RFC 8949:

> If any definite-length text string inside an indefinite-length text string is invalid, the indefinite-length text string is invalid. Note that this implies that the UTF-8 bytes of a single Unicode code point (scalar value) cannot be spread between chunks: a new chunk of a text string can only be started at a code point boundary.

I don't think this does any checking that the contents are valid UTF-8. Therefore, I'm not sure if it accepts invalid CBOR that breaks text within a codepoint. e.g. `7f61ec629c84ff` should be invalid but `7f63ec9c84ff` should decode to "위".

rust-src/concordium_base/src/common/cbor.rs
```diff
@@ -539,19 +539,13 @@ pub trait CborDecoder {
...
     /// Decode bytes into given `destination`. The length of the bytes data item
     /// must match the `destination` length, else an error is returned.
-    ///
-    /// Works only for definite length bytes.
     fn decode_bytes_exact(self, destination: &mut [u8]) -> CborSerializationResult<()>;
 
     /// Decode text and return UTF8 encoding.
-    ///
-    /// Works only for definite length text.
     fn decode_text(self) -> CborSerializationResult<Vec<u8>>;
```
td202, line 549: It might be useful to keep around definite-length versions. In particular, when we move the token module implementation from Haskell to Rust, we may want to ensure that the parsing behaves identically, and thus need to use the definite-length versions in many cases.

allanbrondum: I'm not sure we need to imitate the Haskell limitations necessarily?

rust-src/concordium_base/src/common/cbor/decoder.rs
allanbrondum: The `ciborium_ll` `Segment` reader parses and returns `&str` so I think it will check that. But in any case, it will be good to cover this by automatic tests, will add some tests of UTF-8 corner cases.

rust-src/concordium_base/src/common/cbor.rs
allanbrondum: If we do at some point, I would add options to `SerializationOptions` that can disallow certain constructs. E.g. check that the headers specify definite length. I would still use the same "multisegment" decoder that is implemented now though, I would just add a check before calling it.

rust-src/concordium_base/src/common/cbor/decoder.rs
allanbrondum: @td202 I've updated the code a bit and added some tests of two byte code points

rust-src/concordium_base/src/common/cbor/decoder.rs
```diff
@@ -213,53 +218,103 @@ where
     fn options(&self) -> SerializationOptions { self.options }
 }
 
+trait CursorExt {
+    /// Advance the position of the cursor with by `len`, or as many positions
```
DOBEN, line 222:
```suggestion
    /// Advance the position of the cursor by `len`, or as many positions
```

PR #738, opened 2025-08-27
fix: add smart contract version for forwards compatibility to store it as a u8

squirer:
## Purpose

Add new type `SmartContractVersion` to store the version as a u8. This will be used in forwards compatibility when mapping the version

## Changes

Added new type so that the smart contract version can be stored as a u8

## Checklist

      hard-to-understand areas.


rust-src/concordium_base/src/smart_contracts.rs
```diff
@@ -258,3 +259,27 @@ impl ContractEvent {
...
+
+/// Represents the smart contract version as a u8
+#[derive(SerdeSerialize, SerdeDeserialize, Default, Debug, Clone, PartialEq, Eq)]
+#[serde(transparent)]
+#[repr(transparent)]
+pub struct SmartContractVersion(pub u8);
+
+
+/// Convert the smart contract version from i32 to u8
+impl TryFrom<i32> for SmartContractVersion {
```
allanbrondum, line 271: I would put the conversion from `i32` to `SmartContractVersion` where the conversion is needed and not by implementing `TryFrom<i32>`. I will be a bit arbitrary/confusing for users that there is a conversion from `i32`

rust-src/concordium_base/src/smart_contracts.rs
```diff
@@ -258,3 +259,27 @@ impl ContractEvent {
...
+    type Error = TryFromIntError;
+
+    fn try_from(value: i32) -> Result<Self, Self::Error> {
+        u8::try_from(value).map(SmartContractVersion)
+    }
+}
+
+/// Convert from WasmVersion to our Smart Contract Version
+impl From<WasmVersion> for SmartContractVersion {
```
allanbrondum, line 280: Suggestion: Also add conversion `TryFrom` the other way

rust-src/concordium_base/src/smart_contracts.rs
```diff
@@ -258,3 +259,27 @@ impl ContractEvent {
         Ok(res)
     }
 }
+
+/// Represents the smart contract version as a u8
+#[derive(SerdeSerialize, SerdeDeserialize, Default, Debug, Clone, PartialEq, Eq)]
```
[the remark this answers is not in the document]
allanbrondum: Probably also `Display`, see `ProtocolVersionInt` for an example

rust-src/concordium_base/src/smart_contracts.rs
```diff
@@ -258,3 +259,27 @@ impl ContractEvent {
...
+#[serde(transparent)]
+#[repr(transparent)]
+pub struct SmartContractVersion(pub u8);
```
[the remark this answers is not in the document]
squirer: resolving this now

rust-src/concordium_base/src/smart_contracts.rs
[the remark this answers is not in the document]
squirer: done

rust-src/concordium_base/src/smart_contracts.rs
```diff
@@ -258,3 +258,34 @@ impl ContractEvent {
         Ok(res)
     }
 }
+
+/// Represents the wasm version (smart contract version) as a u8
+/// Copy, PartialOrd, Ord and Hash
```
limemloh, line 263:
```suggestion
```

rust-src/concordium_base/src/smart_contracts.rs
```diff
@@ -190,7 +190,7 @@ impl ContractTraceElement {
 /// possibly some transfers.
 pub struct InstanceUpdatedEvent {
     #[serde(default)]
-    pub contract_version: WasmVersion,
+    pub contract_version: WasmVersionInt,
```
limemloh, line 193: Please add this change to the changelog

limemloh (approved):

rust-src/concordium_base/src/smart_contracts.rs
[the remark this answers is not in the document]
limemloh: You could change the `#[serde(default)]` in `InstanceUpdatedEvent`  to `#[serde(default = "wasm_version_fallback")]` and then define the `fn wasm_version_fallback() -> WasmVersionInt {WasmVersionInt(0)}` below `InstanceUpdatedEvent`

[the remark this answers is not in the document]
squirer: Please review the changes i made here and in the Rust SDK PR - implented the fallback method in concordium base where we are defining the WasmVersionInt, then in the rust SDK our mod file is now using the same fallback reference defined in concordium base. Hope this is correct, please check and let me know

rust-src/concordium_base/src/smart_contracts.rs
squirer: I have already added this line to the Rust SDK changelog:

- Using a newly defined type `WasmVersionInt` defined in concordium-base in order to make the Wasm Version (Smart contract version) forwards compatible by storing it there as a u8.

I seen concordium base does not seem to have its own changelog, just let me know if there is another location too?

rust-src/concordium_base/src/smart_contracts.rs
```diff
@@ -258,3 +258,31 @@ impl ContractEvent {
...
+/// Try convert from WasmVersionInt to WasmVersion
+impl TryFrom<WasmVersionInt> for WasmVersion {
+    type Error = U8WasmVersionConvertError;
+    fn try_from(value: WasmVersionInt) -> Result<Self, Self::Error> {
+        WasmVersion::try_from(value.0)
+    }
+}
+
+/// fallback version for the WasmVersionInt
+pub fn wasm_version_fallback() -> WasmVersionInt {
```
limemloh, line 286: Don't make it `pub`, users of this library should not see it

squirer: hmm if its not public, i cannot reuse it in the rust sdk - as I need to be able to access this function for the default definition for the `ContractInitializedEvent`, what would be the suggested approach for this?

rust-src/concordium_base/src/smart_contracts.rs
squirer: my bad, i was checking the root of the repository...updated it now

PR #742, opened 2025-09-03
Ar/cor 1781 clean up haskell cbor model to match cis 7 spec rename token holder

allanbrondum:
## Purpose

Rename the `CborTokenHolder` type to `CborHolderAccount` (same name as in the Rust code)

## Changes

## Checklist

      hard-to-understand areas.


haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
```diff
@@ -396,12 +396,12 @@ decodeCborTokenHolderHelper = \case
             Right coinInfo -> Just coinInfo
 
 -- | Decode a CBOR-encoded 'TokenHolder'.
```
td202, line 398:
```suggestion
-- | Decode a CBOR-encoded 'CborHolderAccount'.
```

td202 (commented):
I think `CborHolderAccount` is a bad name. I would suggest `CborAccountAddress` instead.

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
soerenbf: I agree, also aligns better with the CIS7 description

PR #753, opened 2025-10-07
Ar/cor 1970 identity credential attributes prover and verifier copied

allanbrondum:
## Purpose

Implement first (very) naive version of proving identity attribute commitments. 
It should just be a copy of the credential deployment prover and verifier.

The idea is to create a baseline, such that further changes can be compared with the existing credential deployment.

## Changes

Copied `account_holder::create_credential` and `chain::verify_cdi` and the associated types.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/src/id/account_holder.rs
+++ b/rust-src/concordium_base/src/id/account_holder.rs
@@ -486,11 +486,11 @@ fn generate_pio_common<'a, P: Pairing, C: Curve<Scalar = P::ScalarField>, R: ran
 /// Convenient data structure to collect data related to a single AR
 pub struct SingleArData<'a, C: Curve> {
     pub ar: &'a ArInfo<C>,
-    share: Value<C>,
+    pub share: Value<C>,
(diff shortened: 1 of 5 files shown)


rust-src/concordium_base/src/id/account_holder.rs
```diff
@@ -486,11 +486,11 @@ fn generate_pio_common<'a, P: Pairing, C: Curve<Scalar = P::ScalarField>, R: ran
 /// Convenient data structure to collect data related to a single AR
```
squirer: Just paying some more close attention to this, now im starting to work in this area. Originally when i seen AR here, I had not really a clue what this meant, and then when i checked the blue paper update, I noticed the comment:

"Major terminology updates across the codebase, particularly replacing "AR" (Anonymity Revoker) terminology with "PG" (Privacy Guardian)" 

I think we should be explicit with the naming now with the blue paper changes and update this to be `SinglePrivacyGuardianData`

PR #889, opened 2026-02-26
Allow both revealing attribute and using it in bullet proofs

allanbrondum:
## Purpose

Fix such that an attribute can both be used in bullet proofs and revealed at the same time. This was not possible for identity based presentations (actually, it depended on the order of the statements, if the prover would fail).

## Changes

If an attribute is both revealed and used on bullet proofs, it is now just committed to in the identity credential attributes proof. 

## Checklist

      hard-to-understand areas.


tschudid (approved):
Main change: Ensure that we commit to a value if it is used in a statement that is not `reveal`. 
Looks good to me.

PR #895, opened 2026-03-03
Rbac cddl

soerenbf:
Closes RBC-4

## Purpose

Add support for multiple role assignments in single admin roles update operation

## Checklist

      hard-to-understand areas.

### DIFF
--- a/cddl/cis-7.cddl
+++ b/cddl/cis-7.cddl
@@ -1,12 +1,18 @@
-; Common Types
+; Specification of CCDL related to protocol-level tokens.
+;
+; When changing it, please check the syntax first using
+;   cargo install cddl
+;   cddl compile-cddl --cddl cis-7.cddl
 
-token-id = text .size (1..128) .regexp "[a-zA-Z0-9\-.%]+"
+; Common Types 
+
+token-id = (text .size (1..128)) .regexp "[a-zA-Z0-9\\-.%]+"
 
 token-module-hash = sha256-hash
 
 sha256-hash = bytes .size(32)
 
-token-decimals = int .range (0..255)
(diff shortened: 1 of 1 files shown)


cddl/cis-7.cddl
```diff
@@ -344,7 +344,7 @@ reject-details-mint-would-overflow = {
 
 ; The authorizations query for a token. This query returns `token-authorizations-response`.
 token-authorizations-query = {
-    "roles": [ * token-admin-role ]
+    "roles": [ * token-admin-roles ]
```
allanbrondum:
```suggestion
    "roles": [ * token-admin-role ]
```

soerenbf: good catch

cddl/cis-7.cddl
```diff
@@ -344,7 +344,7 @@ reject-details-mint-would-overflow = {
 
 ; The authorizations query for a token. This query returns `token-authorizations-response`.
 token-authorizations-query = {
```
allanbrondum: Notice that on https://github.com/Concordium/concordium-node/pull/1566/changes#diff-cb962b74b66d51efff030e15a7e2a899d6cf2f69541f07b4abbbde28c9fe327bR10, there are no roles specified in the query? (I think it is fine and simpler not to specify any roles and just return all roles, that would be my preferrence, but as long as the documents are consistent, then it is ok)

allanbrondum (approved):

cddl/cis-7.cddl
soerenbf: yeah, forgot to remove this, thanks

squirer (approved):

PR #897, opened 2026-03-04
Add token authorizations type

soerenbf:
Ref RBC-8, Closes RBC-23

## Purpose

Depends on https://github.com/Concordium/concordium-grpc-api/pull/117.

Adds the authorizations model which will be used as the response for the token authorizations query for the GRPC service.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Types/Queries/Tokens.hs
+++ b/haskell-src/Concordium/Types/Queries/Tokens.hs
@@ -9,6 +9,7 @@ module Concordium.Types.Queries.Tokens (
     TokenAccountState (..),
     TokenState (..),
     TokenInfo (..),
+    TokenAuthorizations (..),
 ) where
 
 import Data.Aeson as AE
@@ -225,3 +226,12 @@ instance Serialize TokenInfo where
         tiTokenId <- get
         tiTokenState <- get
         return TokenInfo{..}
+
+-- | Role based authorizations structure for a given token. The authorizations are CBOR encoded
+-- inside the 'taDetails' field
+data TokenAuthorizations = TokenAuthorizations
+    { -- | The symbol uniquely identifying the protocol-level token.
+      taTokenId :: !TokenId,
+      -- | The CBOR encoded authorizations details
+      taDetails :: !BS.ByteString
+    }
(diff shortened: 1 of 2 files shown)


haskell-src/Concordium/Types/Queries/Tokens.hs
```diff
@@ -225,3 +226,12 @@ instance Serialize TokenInfo where
         tiTokenId <- get
         tiTokenState <- get
         return TokenInfo{..}
+
+-- | Role based authorizations structure for a given token. The authorizations are CBOR encoded
+-- inside the 'taDetails' field
+data TokenAuthorizations = TokenAuthorizations
+    { -- | The symbol uniquely identifying the protocol-level token.
+      taTokenId :: !TokenId,
```
squirer, line 234: Just a question, for my naiive knowledge of this repo - is there a reason we need `ta` in front of this? same for details below - would it be better if it was just `tokenId` and `details` as we are already inside a `TokenAuthorizations` data structure?

squirer (approved):
I just added a question

haskell-src/Concordium/Types/Queries/Tokens.hs
limemloh: In Haskell, each field in a datatype results in a function being defined as part of the module.
So it means there is now a getter function `taTokenId :: TokenAuthorizations -> TokenId` in this module.

This has the downside of two datatypes within the same module cannot use the same name for the field, since it would introduce getters functions with the same name, therefore the convention is to prefix fields like above.

td202 (approved):

haskell-src/Concordium/Types/Queries/Tokens.hs
squirer: ah makes sense! Thanks

soerenbf: This is convention is haskell to uniquely identify the functions pulling out e.g. details of a data structure. As a concrete example, `taTokenId` is a function available in the `Tokens.hs` module that returns the `TokenId` inside a `TokenAuthorizations` value, as in haskell you don't have other ways (AFAIK) to access properties of a structure.

I'm still a haskell novice, so take this with a grain of salt 😅

PR #899, opened 2026-03-09
Feature/rbc 7 update token metadata url support

squirer:
## Purpose

Add support for updating metadata token information for a protocol level token

## Changes

Added token operation and event details for updating token metadata

## Checklist

      hard-to-understand areas.


rust-src/concordium_base/src/protocol_level_tokens/token_event.rs
```diff
@@ -82,7 +86,7 @@ impl TokenModuleEventType {
 }
 
 /// Token module event parsed from type and CBOR
-#[derive(Debug, Clone, Eq, PartialEq)]
```
squirer, line 85: Note: had to remove Eq here because the HashMap defined in the UpdateMetadataEvent which is a String to cbor value map, means it cannot derive Eq due to the support there for cbor values that are f64.

PR #907, opened 2026-03-13
fix: dead code flagedd by node release pipeline. Adding deadcode tag …

squirer:
## Purpose

fix dead code errors on node release pipeline for mac OS

## Changes

The node release pipeline has flaggged the following errors for building concordium base, during the `node-macos` build step. 

Added deadcode for now to these 3 structures as they are never constructed.

image

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/src/sigma_protocols/dlogaggequal.rs
+++ b/rust-src/concordium_base/src/sigma_protocols/dlogaggequal.rs
@@ -16,11 +16,13 @@ use crate::{
 use itertools::izip;
 use std::rc::Rc;
 
+#[allow(dead_code)]
 pub struct DlogAndAggregateDlogsEqual<C: Curve> {
     pub dlog: Dlog<C>,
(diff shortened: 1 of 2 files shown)


td202 (approved):

allanbrondum:
@squirer We have to revert this. If the code is dead, it needs to be properly fixed by being removed, not by allowing the warning. If the build fails on warnings in base, it is because it is running the wrong build command (we will just have the same problem in the future, the build must be resilient to clippy warnings from base). It is this action in the workflow that makes the warnings turn into errors:
```
      - uses: actions-rust-lang/setup-rust-toolchain@v1
        with:
          toolchain: ${{ env.RUST_VERSION }}
```
We should remove this action where ever we see it, it is just causing problems and bad behaviour. You can simply replace the action with settings the default toolchain (rustup is installed already on our builders):
`rustup default ${{ env.RUST_VERSION }}`. If you do this, the warnings in base will appear as warnings only, and not make the build fail.

PR #911, opened 2026-03-18
CDDL schema for locks

td202:
## Purpose

Define the CDDL schema used for lock transactions, events and queries.

## Changes

- Add 'locks.cddl' defining the schema extensions for locks.
- Update GRPC pointer to include lock queries and events.

## Checklist

      hard-to-understand areas.


cddl/locks.cddl
allanbrondum, line 1: I think it would be better to merge this together with in a CDDL shared with the access control changes (which are currently in `cis-7.cddl`)

cddl/locks.cddl
```diff
@@ -0,0 +1,451 @@
...
+
+; The simple lock controller initialization parameters.
+lock-controller-simple-init = {
+    "simpleV0": {
+        ; The set of capability grants for the simple lock.
+        ; At least one account must be granted the "fund" capability.
+        "grants": [ + lock-controller-simple-grant ],
+        ; If set, then the lock will not be destroyed before expiry except
+        ; by a cancel operation.
+        ? "keepAlive": bool .default false,
```
soerenbf: Not sure what this means.. For an encode/decode roundtrip to be the same, we need to make sure that there is no ambiguity wrt. how `false` can be described here. 

Does it mean that 
1. if `"keepAlive"` is not defined it should be interpreted as `false`, or 
2. if `"keepAlive"` is not defined, decode it as `false`.

I feel like we cannot use 2 as it is non-deterministic, but I'm wondering if it's better to drop the `.default`?

Alternatively, we could 
```suggestion
        ; If set, then the lock will not be destroyed before expiry except
        ; by a cancel operation.
        ? "keepAlive": true,
```

td202: For deterministic encoding, the CDDL spec implies that `.default` fields should be absent rather than specify the default value.

soerenbf (approved):
Looks good!

allanbrondum (approved):

PR #912, opened 2026-03-19
Add type TokenAuthorizations with CBOR encoding

limemloh:
## Purpose

Closing [COR-2340](https://linear.app/concordium/issue/COR-2340/implement-cbor-token-authorizations-in-base).

## Changes

- Add the `TokenAuthorizations` struct corresponding to the CDDL.
- Derive `Hash` for enum `TokenAdminRole` .

### DIFF
--- a/rust-src/concordium_base/src/protocol_level_tokens/token_authorizations.rs
+++ b/rust-src/concordium_base/src/protocol_level_tokens/token_authorizations.rs
@@ -1,0 +1,57 @@
+use crate::protocol_level_tokens::CborHolderAccount;
+use concordium_base_derive::{CborDeserialize, CborSerialize};
+
+/// Describes the authorizations structure for protocol level tokens.
+#[derive(Debug, Clone, PartialEq, Eq, CborSerialize, CborDeserialize, Default)]
+pub struct TokenAuthorizations {
+    /// Gives authority to perform `token-assign-admin-roles` and `token-revoke-admin-roles` operations.
+    pub update_admin_roles: TokenRoleAuthorizationsCbor,
+    /// Gives authority to perform `token-mint` operations.
+    pub mint: TokenRoleAuthorizationsCbor,
+    /// Gives authority to perform `token-burn` operations.
+    pub burn: TokenRoleAuthorizationsCbor,
+    /// Gives authority to perform `token-add-allow-list` and `token-remove-allow-list` operations.
+    pub update_allow_list: TokenRoleAuthorizationsCbor,
+    /// Gives authority to perform `token-add-deny-list` and `token-remove-deny-list` operations.
+    pub update_deny_list: TokenRoleAuthorizationsCbor,
+    /// Gives authority to perform `token-pause` and `token-unpause` operations.
+    pub pause: TokenRoleAuthorizationsCbor,
+    /// Gives authority to perform `token-update-metadata` operations.
+    pub update_metadata: TokenRoleAuthorizationsCbor,
+}
+
+/// Authorizations details applicable to any admin role.
+#[derive(Debug, Clone, PartialEq, Eq, CborSerialize, CborDeserialize, Default)]
+pub struct TokenRoleAuthorizationsCbor {
+    /// The accounts that have the role assigned.
+    pub accounts: Vec<CborHolderAccount>,
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::common::cbor;
+    use concordium_contracts_common::AccountAddress;
+
+    #[test]
+    fn test_token_authorizations_cbor() {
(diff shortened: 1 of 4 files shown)


rust-src/concordium_base/src/protocol_level_tokens/token_authorizations.rs
```diff
@@ -0,0 +1,57 @@
...
+    pub burn: TokenRoleAuthorizationsCbor,
+    /// Gives authority to perform `token-add-allow-list` and `token-remove-allow-list` operations.
+    pub update_allow_list: TokenRoleAuthorizationsCbor,
+    /// Gives authority to perform `token-add-deny-list` and `token-remove-deny-list` operations.
+    pub update_deny_list: TokenRoleAuthorizationsCbor,
+    /// Gives authority to perform `token-pause` and `token-unpause` operations.
+    pub pause: TokenRoleAuthorizationsCbor,
+    /// Gives authority to perform `token-update-metadata` operations.
+    pub update_metadata: TokenRoleAuthorizationsCbor,
+}
```
td202, line 21: I think that it's possible for roles to be absent (as distinct from empty). For instance, if a token does not support an allow list, then there will be no `update_allow_list` authorizations.

limemloh: I have changed the implementation to fit this, also to only include the non-empty fields

td202: This is still not really what I would expect. If the token supports an allow list but there are currently no accounts wiht `update_allow_list` role, I would expect the authorizations for that role to be present but empty. But if the token does not support an allow list, I would expect it to be absent. That way, one can infer quite directly from getting the authorizations which roles are applicable to the token, rather than having to deduce it from the token configuration.

td202 (approved):

rust-src/concordium_base/src/protocol_level_tokens/token_authorizations.rs
allanbrondum: I'm not sure this data structure must carry the information whether a feature is enabled or not. I think it can be fine that roles for disabled features and enabled features are represented the same way? (currently, by an empty list for disabled features) It is simpler, and does not couple enablement of features and assignment hard together in our data model (in the future, we could e.g. allow assigning roles before enabling a feature, that would not be possible if we bind enabling a feature hard together with allowing assignments)

allanbrondum (approved):

rust-src/concordium_base/src/protocol_level_tokens/token_authorizations.rs
soerenbf: Will just point out that in the design document for the node implementation, the solution described matches what @td202 describes in his comment.

I think there's also advantages to the coupling, i.e. being able to blanket revoke/assign all roles from an account without needing to figure out which features are enabled elsewhere, i.e. iterate through the keys in the authorizations to know which roles are available to assign. That (as far as I recall) was the main reason for structuring the data this way.

It's also similar to how we represent allow-/deny-list information on the account state.

rust-src/concordium_base/src/protocol_level_tokens/token_authorizations.rs
```diff
@@ -0,0 +1,123 @@
+use crate::common::cbor::{CborDeserialize, CborSerialize};
+use crate::protocol_level_tokens::CborHolderAccount;
+use concordium_base_derive::{CborDeserialize, CborSerialize};
+use concordium_contracts_common::AccountAddress;
+
+/// Describes the authorizations structure for protocol level tokens.
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct TokenAuthorizations {
+    /// Gives authority to perform `token-assign-admin-roles` and `token-revoke-admin-roles` operations.
+    pub update_admin_roles: Vec<AccountAddress>,
```
soerenbf: How come we're not using a struct to represent the data for each role, similar to the CBOR encoding? Would that not be better in terms of forward compatibility if we introduce more auth information per role?

soerenbf (approved):

rust-src/concordium_base/src/protocol_level_tokens/token_authorizations.rs
limemloh: okay, I'll adjust accordingly

rust-src/concordium_base/src/protocol_level_tokens/token_authorizations.rs
limemloh: I changed to follow the cbor encoding, but the reason is that it is a really noisy representation in rust, so simplified the rust representation.

PR #915, opened 2026-03-23
MetaUpdate transaction

td202:
## Purpose

Part of: [COR-2301](https://linear.app/concordium/issue/COR-2301/locks-basic-support-for-new-transaction-type)

Implement the types and serialization for the MetaUpdate transaction.

## Changes

- Haskell types and serialization
- Rust types and serialization
- `supportsMetaUpdate` protocol version predicate; determines deserializability of `MetaUpdate` in scheduler.
- Protobuf conversions for events.
- Tests

## Checklist

      hard-to-understand areas.


rust-src/concordium_base/src/protocol_level_tokens/meta_operations.rs
```diff
@@ -0,0 +1,695 @@
+use crate::{
+    common::cbor::{self, CborSerializationResult},
+    protocol_level_tokens::{
+        CborHolderAccount, CborMemo, RawCbor, TokenAmount, TokenId, TokenOperation,
+    },
+};
+use concordium_base_derive::{CborDeserialize, CborSerialize};
+
+pub mod meta_operations {
```
allanbrondum: Would we rather change the existing operation constructors to create the meta update operation instead? Such that we make the SDK's opinionated about preferring using the meta update operations. It seems unnecessary that users must decide between them.

allanbrondum (commented):
I think we should consider if we can reduce the cognitive load on the users of our SDK's by making them opinionated about using the token update or the meta update operation. Right now, as a user, I would be a bit confused about the difference, and why I have to choose.

rust-src/concordium_base/src/protocol_level_tokens/meta_operations.rs
td202: Possibly. I think there's value in having a preferred way of doing things, but I also think if the node supports both transaction types, then they should also be supported in the SDK.

soerenbf: We have a client in the rust SDK for the usability concerns, no?

rust-src/concordium_base/src/protocol_level_locks/lock_id.rs
```diff
@@ -0,0 +1,43 @@
+use concordium_base_derive::{CborDeserialize, CborSerialize};
+
+/// This tag ident
+const LOCK_ID_TAG: u64 = 40920;
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Clone, CborSerialize, CborDeserialize, serde::Serialize, serde::Deserialize)]
+#[cbor(tag = LOCK_ID_TAG, tuple)]
```
soerenbf, line 19: Nice addition with the tuple!

I already merged an implementation of this type with #919. Would be good if we could merge these?

I added a new `protocol_level_locks` module to hold types and functionality related to locks specifically. I imagine there will be a `plt-locks-module` as well in the plt scheduler implementation in the node which would align well with this module. What do you think?

rust-src/concordium_base/src/protocol_level_tokens/token_id.rs
```diff
@@ -107,6 +108,25 @@ impl common::Deserial for TokenId {
...
+    where
+        Self: Sized,
+    {
+        let encoded = decoder.decode_text()?;
+        let string = String::from_utf8(encoded).map_err(CborSerializationError::invalid_data)?;
+        string
+            .try_into()
+            .map_err(CborSerializationError::invalid_data)
+    }
+}
```
soerenbf, line 128: Also added in #919

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
soerenbf, line 1: Are these even needed in haskell? is it to support `concordium-client`?

rust-src/concordium_base/src/protocol_level_locks/lock_id.rs
td202: I have moved the file to the same place, and we can resolve the difference when we merge the branches.

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
td202: Yes, `concordium-client` will need these. They are also used for constructing test meta-update payloads (see: https://github.com/Concordium/concordium-node/pull/1605/changes#diff-cd939690035bb80b72f7d25276e0b8ff0fd780f338a8c69f7235edd073029b71R207).

rust-src/concordium_base/src/protocol_level_tokens/token_id.rs
td202: Will resolve on merge.

soerenbf (approved):
Looks good. Are we sure we want to merge this into the p11 branch though? I see the rationale in doing it as it is a more or less self contained piece of work, but maybe it makes sense to release with the rest of the milestone defining the minimal functionality for locks instead?

PR #919, opened 2026-03-26
Rust types for node locks state

soerenbf:
Ref COR-2213

## Purpose

Add types needed to represent locks state in the node block state

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/src/lib.rs
+++ b/rust-src/concordium_base/src/lib.rs
@@ -7,6 +7,7 @@ pub mod cis4_types;
 pub mod constants;
 pub mod hashes;
 mod internal;
+pub mod protocol_level_locks;
 pub mod protocol_level_tokens;
 pub mod smart_contracts;
 pub mod transactions;
--- a/rust-src/concordium_base/src/protocol_level_locks/lock_config.rs
+++ b/rust-src/concordium_base/src/protocol_level_locks/lock_config.rs
@@ -8,11 +8,6 @@ use concordium_base_derive::{CborDeserialize, CborSerialize};
 /// Contains the list of recipients, the expiry time, and the controller
 /// configuration for a lock.
 #[derive(Debug, Clone, Eq, PartialEq, CborSerialize, CborDeserialize)]
-#[cfg_attr(
-    feature = "serde_deprecated",
-    derive(serde::Serialize, serde::Deserialize)
-)]
-#[cfg_attr(feature = "serde_deprecated", serde(rename_all = "camelCase"))]
 pub struct LockConfig {
     /// Accounts that can receive funds from this lock.
     pub recipients: Vec<CborHolderAccount>,
--- a/rust-src/concordium_base/src/protocol_level_locks/lock_config.rs
+++ b/rust-src/concordium_base/src/protocol_level_locks/lock_config.rs
@@ -1,0 +1,153 @@
+use super::LockController;
+use crate::common::types::TransactionTime;
+use crate::protocol_level_tokens::CborHolderAccount;
+use concordium_base_derive::{CborDeserialize, CborSerialize};
+
+/// Top-level lock configuration.
+///
+/// Contains the list of recipients, the expiry time, and the controller
+/// configuration for a lock.
+#[derive(Debug, Clone, Eq, PartialEq, CborSerialize, CborDeserialize)]
+#[cfg_attr(
+    feature = "serde_deprecated",
+    derive(serde::Serialize, serde::Deserialize)
+)]
+#[cfg_attr(feature = "serde_deprecated", serde(rename_all = "camelCase"))]
+pub struct LockConfig {
+    /// Accounts that can receive funds from this lock.
+    pub recipients: Vec<CborHolderAccount>,
+    /// Expiry time of the lock (seconds since epoch).
+    pub expiry: TransactionTime,
+    /// Controller configuration for the lock.
+    pub controller: LockController,
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::common::cbor;
+    use crate::protocol_level_locks::{
+        LockControllerSimpleV0, LockControllerSimpleV0Capability, LockControllerSimpleV0Grant,
+    };
+    use crate::protocol_level_tokens::test_fixtures::ADDRESS;
+
+    /// Build a full `LockConfig` for testing.
+    fn example_lock_config() -> LockConfig {
+        LockConfig {
+            recipients: vec![CborHolderAccount::from(ADDRESS)],
+            expiry: TransactionTime::from_seconds(1804806000),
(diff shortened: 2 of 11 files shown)


rust-src/concordium_base/src/lib.rs
```diff
@@ -7,6 +7,7 @@ pub mod cis4_types;
 pub mod constants;
 pub mod hashes;
 mod internal;
+pub mod protocol_level_locks;
```
soerenbf, line 10: Need to amend the changelog

rust-src/concordium_base/src/protocol_level_locks/lock_controller_simple_v0.rs
```diff
@@ -0,0 +1,436 @@
...
+#[cfg_attr(feature = "serde_deprecated", serde(rename_all = "camelCase"))]
+pub struct LockControllerSimpleV0 {
+    /// Capability grants to accounts.
+    pub grants: Vec<LockControllerSimpleV0Grant>,
+    /// Tokens affected by this lock controller.
+    pub tokens: Vec<TokenId>,
+    /// Whether the lock should be kept alive after all funds are
+    /// returned. Defaults to `false` when omitted from CBOR.
+    #[cfg_attr(feature = "serde_deprecated", serde(default))]
+    pub keep_alive: bool,
```
soerenbf: Maybe this should be 
```suggestion
    pub keep_alive: Option<bool>,
```

depending on the resolution of https://github.com/Concordium/concordium-base/pull/911/changes#r2996707367. It's important that any decode/encode roundtrip results in the same encoding as the original.

In that case, the cbor (de)serialize implementation should be changed accordingly.

soerenbf: Keeping this open, but going with `Option<bool>` for now to keep the model deterministic wrt. serialization.

rust-src/concordium_base/src/protocol_level_locks/lock_config.rs
```diff
@@ -0,0 +1,153 @@
...
+
+/// Top-level lock configuration.
+///
+/// Contains the list of recipients, the expiry time, and the controller
+/// configuration for a lock.
+#[derive(Debug, Clone, Eq, PartialEq, CborSerialize, CborDeserialize)]
+#[cfg_attr(
+    feature = "serde_deprecated",
+    derive(serde::Serialize, serde::Deserialize)
+)]
```
allanbrondum: I suggest not adding serde serialization here

rust-src/concordium_base/src/protocol_level_locks/lock_controller.rs
```diff
@@ -0,0 +1,101 @@
...
+
+/// Top-level lock controller type.
+///
+/// Each variant represents a different controller version.
+#[derive(Debug, Clone, Eq, PartialEq, CborSerialize, CborDeserialize)]
+#[cbor(map)]
+#[cfg_attr(
+    feature = "serde_deprecated",
+    derive(serde::Serialize, serde::Deserialize)
+)]
```
allanbrondum: Same here, and all the other places also.

rust-src/concordium_base/src/protocol_level_tokens/token_id.rs
```diff
@@ -107,8 +108,32 @@ impl common::Deserial for TokenId {
...
+/// CBOR deserialization of [`TokenId`] from a CBOR text string.
+impl cbor::CborDeserialize for TokenId {
+    fn deserialize<C: cbor::CborDecoder>(decoder: C) -> cbor::CborSerializationResult<Self>
+    where
+        Self: Sized,
+    {
+        let text_bytes = decoder.decode_text()?;
+        let text = String::from_utf8(text_bytes).map_err(|_| {
+            CborSerializationError::invalid_data(format_args!("TokenId text is not valid UTF-8"))
+        })?;
```
allanbrondum:
```suggestion
        let text = String::deserialize(decoder)?;
```

rust-src/concordium_base/src/protocol_level_locks/lock_controller_simple_v0.rs
```diff
@@ -0,0 +1,436 @@
...
+/// operations.
+#[derive(Debug, Clone, Eq, PartialEq, CborSerialize, CborDeserialize)]
+#[cfg_attr(
+    feature = "serde_deprecated",
+    derive(serde::Serialize, serde::Deserialize)
+)]
+#[cfg_attr(feature = "serde_deprecated", serde(rename_all = "camelCase"))]
+pub struct LockControllerSimpleV0Grant {
+    /// The account receiving the grant.
+    pub account: CborHolderAccount,
```
allanbrondum, line 75: The type `CborHolderAccount` has cost so much time and head ache for me. I suggest (or insist?) that we create a new one `CborHolderAccountV1` where coin info is mandatory inside: `pub coin_info: CoinInfo`. And then we use that in all new functionality that let it be reflected in the CDDL also for new functionality.

allanbrondum (commented):
I have two general comments: That we should not use `CborHolderAccount` in new functionality, we should use a variant where coin info is mandatory. And also that we don't add serde serialization to new types .

rust-src/concordium_base/src/protocol_level_locks/lock_controller_simple_v0.rs
soerenbf: What does not work? why is it an issue that it is optional?

allanbrondum: Because the code has to handle it. We have to have tests variants where it is present and where it is not for all our operations (scheduler input). And we have to check manually (without compiler assistance), that it is filled out in all values returned by the scheduler: reject reasons, events, query results (scheduler output). If it is mandatory, there is not variance in the input and output. I've spend at least a week, maybe two, in effective time to handle this in our existing code. So speaking out of experience :)

allanbrondum (approved):

rust-src/concordium_base/src/protocol_level_locks/lock_controller_simple_v0.rs
soerenbf: I don't know.. this is what we agreed on in the cbor schema which is the source of truth for this. If we need to change this, the change needs to start there and we should have a discussion on it in the group, IMO. I'll use it as is for now.

allanbrondum: @soerenbf No, the CCDL has not been reviewed and merged in: https://github.com/Concordium/concordium-base/pull/911

soerenbf: Will make sure to adapt to any changes to that.

PR #922, opened 2026-04-05
Rust PLT block state feature

allanbrondum:
## Purpose

Rust block state feature branch

## Changes

## Checklist

      hard-to-understand areas.


rust-src/concordium_base/src/common/serialize.rs
```diff
@@ -707,12 +683,12 @@ impl<R: ReadBytesExt, A: Deserial> Get<A> for R {
 /// `sink.put(value)` in contrast to `value.serial(sink)`. It is less important
 /// than `Get`.
 pub trait Put<A> {
-    fn put(&mut self, _v: &A);
+    fn put(&mut self, _v: A);
```
limemloh: Why is the reference removed here? I imagine it leads to blanket implementations for `Serial` and do we need to be able to drop the value here?

PR #923, opened 2026-04-16
Ar/psr 84 use smart contract trie for key value store

allanbrondum:
## Purpose

Make tree iterator etc. public such that it can be used in the Rust scheduler in the node.

## Changes

## Checklist

      hard-to-understand areas.


smart-contracts/wasm-chain-integration/src/v1/trie/low_level.rs
```diff
@@ -2270,7 +2270,7 @@ impl MutableTrie {
         }
     }
 
-    pub(crate) fn iter(
+    pub fn iter(
```
td202, line 2275: Can we add some documentation?

td202 (approved):

PR #924, opened 2026-04-16
Merge changes from release/p11

td202:
## Purpose

Merge the `release/p11` branch in to `feature/p11-locks`.


rust-src/concordium_base/src/protocol_level_locks/lock_id.rs
```diff
@@ -1,88 +1,88 @@
...
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Hash,
+    Clone,
+    CborSerialize,
+    CborDeserialize,
+    serde::Serialize,
+    serde::Deserialize,
```
soerenbf: I also had serde defined under the `serde_deprecated` flag, but at that point we agreed to leave it out. 
```suggestion
```

and consequently also remove it anywhere else where it's introduced as a result. Sorry I did not catch this in the initial PR 😅

soerenbf (approved):
Looks good to me!

rust-src/concordium_base/src/protocol_level_locks/lock_id.rs
```diff
@@ -1,88 +1,88 @@
...
+    Serialize,
```
soerenbf: also, maybe we should find a more space-efficient way to serialize this? 🤷 maybe a VLQ triple would make sense?

rust-src/concordium_base/src/protocol_level_locks/lock_id.rs
td202: I think we need it, at least under `serde_deprecated`, because that's how the JSON serialization of events is implemented, which is used by the transaction logger.

rust-src/concordium_base/src/protocol_level_locks/lock_id.rs
td202: I think that depends on what we use the serialization for. If it's for passing across FFI, then we're probably better off using the space over dealing with the encoding/decoding. For storing in state, using VLQs could make sense.

soerenbf: yeah, I was also thinking for storing it in the block state in the node

rust-src/concordium_base/src/protocol_level_locks/lock_id.rs
soerenbf: Alright, I'm also fine with that

rust-src/concordium_base/src/protocol_level_locks/lock_id.rs
soerenbf: We agreed to go with the simple option for now, i.e. keep it as is.

PR #926, opened 2026-04-17
MetaUpdatePayload decoding should disallow unknown fields.

td202:
## Purpose

Decoding of `MetaUpdatePayload` should not permit unknown fields.

### DIFF
--- a/rust-src/concordium_base/src/protocol_level_tokens/meta_operations.rs
+++ b/rust-src/concordium_base/src/protocol_level_tokens/meta_operations.rs
@@ -111,9 +111,13 @@ pub struct MetaUpdatePayload {
 }
 
 impl MetaUpdatePayload {
-    /// Decode meta-update operations from CBOR
+    /// Decode meta-update operations from CBOR. This disallows extraneous fields
+    /// in the encoded CBOR.
     pub fn decode_operations(&self) -> CborSerializationResult<MetaUpdateOperations> {
-        cbor::cbor_decode(&self.operations)
+        let decode_options = cbor::SerializationOptions {
+            unknown_map_keys: cbor::UnknownMapKeys::Fail,
+        };
+        cbor::cbor_decode_with_options(&self.operations, decode_options)
     }
 }
 


rust-src/concordium_base/src/protocol_level_tokens/meta_operations.rs
```diff
@@ -111,9 +111,13 @@ pub struct MetaUpdatePayload {
...
 impl MetaUpdatePayload {
-    /// Decode meta-update operations from CBOR
+    /// Decode meta-update operations from CBOR. This disallows extraneous fields
+    /// in the encoded CBOR.
     pub fn decode_operations(&self) -> CborSerializationResult<MetaUpdateOperations> {
-        cbor::cbor_decode(&self.operations)
+        let decode_options = cbor::SerializationOptions {
+            unknown_map_keys: cbor::UnknownMapKeys::Fail,
+        };
+        cbor::cbor_decode_with_options(&self.operations, decode_options)
```
allanbrondum, line 120: I think this logic (opinion on how to decode) belongs in the node. For the existing transaction, it is implemented in the line
```
let operations: Vec<TokenOperation> = util::cbor_decode(&token_operations)...
```

soerenbf: Yeah I agree with @allanbrondum here

td202: I think that it is absurd to have any decoding of the operations that is not strict in terms of the allowed fields. CIS-7 states:

> The semantics of each token operation SHOULD be the same across all Token Modules which implements it. In particular, implementations MUST conform to the schema for the token operations defined in this document. An implementation MUST NOT use the operation types transfer, mint, burn, addAllowList, removeAllowList, addDenyList, removeDenyList, pause, or unpause for any other operation than those defined below.

In particular, none of those transactions have `* text => any` very purposely. If we want to add fields, then that MUST be a different operation (with a distinct name).

allanbrondum: I don't think CIS-7 is a formal specification that we cannot change or break. That said, it is true that is probably not likely, that we will add a field to any of the operations. On the other hand, I don't think we can rule it out entirely, that we could add fields, that don't change the essentail meaning of the operations. So I don't think we should bake into the SDK, that it should fail if there are unknown fields. That would not be in the spirit of having forwards runtime compatibility where possible. Also, I think consistency with how `TokenOperationsPayload` is handled should be weighted high, it does not seem right that they are handled differently. It also seems the function `MetaUpdatePayload::decode_operations_maybe_known` that helps clients to be forwards compatible is missing.

td202: I think the equivalent function for `TokenOperationsPayload` is also bad for being non-strict and we should fix it.

Adding fields would inherently change the meaning of the operation, because in one (protocol) version an operation would fail with the extra field, and in another version it would not. The node is strict on which fields are allowed. I also think ignoring additional fields is bad for clients, since it could decode two operations to the same thing - one that is valid (and doesn't have extraneous fields) and one that is invalid (and does). This would make it unclear why one fails on serialization but the other doesn't.

td202: Also, I think it doesn't really benefit us to have flexibility to change the payloads, when we have a mechanism for extension that is adding new operation types.

allanbrondum: > Adding fields would inherently change the meaning of the operation, because in one (protocol) version an operation would fail with the extra field, and in another version it would not. The node is strict on which fields are allowed. I also think ignoring additional fields is bad for clients, since it could decode two operations to the same thing - one that is valid (and doesn't have extraneous fields) and one that is invalid (and does).

I think this is true. But I would would rather leave that as a future decision on which (if any) changes we can make to the operations. Rather than "hardcoding" it in the SDK.

td202: This is not deferring the decision. It is choosing to have two different slightly incompatible definitions of what the transaction payload can be on the grounds that we *might* change it in the node at some point in the future, for some reason that we can't imagine at the moment. And that if we do make such a change, then we definitely want tools built with the SDK before the change to handle it by ignoring the extra fields that are added. This seems to be putting the cart before the horse - changes to the node should be largely conservative, but they shouldn't be wholly constrained by how the SDK works. Yes, we should build the SDK to be forwards compatible, but first of all it should be compatible with the current version.

td202:
Closing as agreement could not be reached.

PR #927, opened 2026-04-20
Add token info query types

soerenbf:
Ref COR-2309

## Purpose

Add the types required to query lock information from the node. This does not include the addition to account state; this will be part of another PR.

## Checklist

      hard-to-understand areas.


haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
```diff
@@ -2086,6 +1955,314 @@ tokenAuthorizationsMapFromBytes =
 tokenAuthorizationsMapToBytes :: TokenAuthorizationsMap -> BS.ByteString
 tokenAuthorizationsMapToBytes = encodeToBytes . encodeTokenAuthorizationsMap
 
+-- * Lock identifiers
```
allanbrondum: Can you remind me why we need the lock model in Haskell? One of the benefits of the Rust rewrite is that we should not need a Haskell representation of this model. If we still need it, maybe there is something we have not accounted for.

rust-src/concordium_base/src/common/cbor.rs
```diff
@@ -279,6 +279,25 @@ pub mod __private {
             }
         }
     }
+
+    pub trait StructMapCbor: Sized {
```
allanbrondum: Suggest to rename to `CborDeserializeAsMap` and `CborSerializeAsMap` (putting the two functions in different traits) to be consistent with `CborDeserialize` and `CborSerialize`.

allanbrondum: And then I would just move it out of `__private` module. Same for `decode_cbor_from_value` if we actually need that function.

rust-src/concordium_base/src/common/cbor.rs
allanbrondum, line 1: Remember to add documentation of the flatten attribute at the module level documentation in this file

rust-src/concordium_base/src/common/cbor.rs
```diff
@@ -942,6 +961,175 @@ mod test {
         assert_eq!(value_decoded, value);
     }
 
+    #[test]
+    fn test_struct_as_map_derived_flatten() {
```
allanbrondum: Would be nice with a test of nested flattening. Of a documentation of that it is not possible, if it does not work.

allanbrondum: And a test of flattening two fields in the same struct

allanbrondum (commented):
My take would be that we should not implement the `[cbor(flatten)]` macro. Serialization macros that transform structure just tends to be a mess, especially when combined with other local effects/macros. You e.g. have to deserialize into a dynamic model `cbor::Value`. You see the same in serde, where the structure transforming macros are also a mess and a cause of a number of bugs. Alternative is just to inline `LockConfig` where ever it is used. Or just inline it is a temporary struct used during serialization/deserialization:
```
struct A {
    a: String,
    config: LockConfig,
}

impl CborSerialize for A {
    fn serialize<C: CborEncoder>(&self, encoder: C) -> Result<(), C::WriteError> {
        #[derive(CborSerialize)]
        struct ACbor<'a> {
            a: &'a String,
            recipients: &'a Vec<CborHolderAccount>,
            expiry: &'a TransactionTime,
            controller: &'a LockController,
        };
        
        let val_cbor = ACbor {
            a: &self.a,
            recipients: &self.config.recipients,
            expiry: &self.config.expiry,
            controller: &self.config.controller,
        };
        
        val_cbor.serialize(encoder)
    }
}
```

haskell-src/Concordium/Types/ProtocolLevelTokens/CBOR.hs
soerenbf: concordium-client is the only thing that comes to mind, but this is still in scope for the locks feature.. maybe you could argue that we can wait with implementing the haskell types until that point, but it's easier to do them now alongside the rust changes IMO.

allanbrondum: My take would be that we should postpone any work on tooling, that we are not 100% sure we need. Especially if it is in Haskell, because we know we might very well rewrite very soon.

soerenbf: Right.. that could be a strategy, but now the code is already there to support the tooling implementation. Can we agree that we include it now and and make a note that it can be worthwhile to defer implementation of the rest until we need it? 😅

haskell-src/Concordium/Types/Locks.hs
```diff
@@ -0,0 +1,44 @@
...
+    { liAccountIndex :: !Word64,
+      liSequenceNumber :: !Word64,
+      liCreationOrder :: !Word64
+    }
+    deriving (Eq, Show)
+
+instance AE.ToJSON LockId where
+    toJSON LockId{..} =
+        AE.object
+            [ "accountIndex" AE..= liAccountIndex,
```
limemloh, line 23: Should we maybe introduce a more compact string representation.
If the lock ID will be shown in UI anyway then maybe we should define and use the representation now.

Like it could be the base58check representation of the BE encoding of the u64s or simply `"<index>/<seq>/<order>"` so `"45/0/0"`

haskell-src/Concordium/Types/Locks.hs
```diff
@@ -0,0 +1,44 @@
...
+
+-- | Lock identifier: a trio of numbers that together uniquely identify a lock.
+data LockId = LockId
+    { liAccountIndex :: !Word64,
+      liSequenceNumber :: !Word64,
+      liCreationOrder :: !Word64
+    }
+    deriving (Eq, Show)
+
+instance AE.ToJSON LockId where
```
limemloh, line 20: Where do we actually need the JSON representation?

limemloh (approved):

haskell-tests/Types/CBOR.hs
```diff
@@ -1217,7 +1222,318 @@ testTokenAuthorizationsMapJSON = describe "TokenAuthorizationsMap JSON" $ do
...
+                        Seq.singleton
+                            LockedTokenAmount
+                                { ltaToken = ccdTokenId,
+                                  ltaAmount = TokenAmount (TokenRawAmount 12300) 3
+                                }
+                    }
+        }
+  where
+    acc = AccountAddress $ FBS.pack [1 .. 32]
+    ccdTokenId = case makeTokenId (BSS.toShort "CCD") of
```
td202: I would recommend against using `CCD` as a Token ID in any examples. Firstly, it's potentially confusing simply because CCD isn't a PLT. Secondly, in JSON encoding the string `"CCD"` is used for `chainInfo` to indicate a Concordium account.

td202 (approved):

haskell-src/Concordium/Types/Locks.hs
soerenbf: We would need it in concordium-client for doing e.g. multis-sig/sponsored transactions and wallet proxy for events

haskell-tests/Types/TokensSpec.hs
```diff
@@ -151,6 +152,36 @@ testTokenAmountEncodeDecode :: Property
...
     decodeFull get (encode a) == Right a
 
+-- | Test the binary serialization and deserialization of 'LockId'.
+-- The encoding is 3 x Word64 big-endian, 24 bytes total.
+testLockIdSerialize :: Spec
+testLockIdSerialize = describe "LockId" $ do
+    -- Fixture: account=0x1234, sequence=0x5678, order=0x9abc
+    -- Each Word64 big-endian: 8 bytes
+    it "encode fixture" $
+        BS16.encode (encode (LockId 0x1234 0x5678 0x9abc))
```
td202: Please use the record syntax for constructing `LockId`s for the tests, because we want to be sure that the field names are correct.

haskell-src/Concordium/Types/Locks.hs
soerenbf: yeah, @td202 and I had a discussion on the binary serialization of the lock ID where an idea was also to have a triple VLQ encoding, which could be represented as hex. Your suggestion is better in terms of human interpretation obviously, but I doubt that any of the 3 values will have any meaning for any user though 🤔

Asked claude for an analysis on representation efficiency, and it got back with these examples:

 ```text
   (0, 0, 0)
   VLQ hex:              000000                    len 6
   VLQ Base58Check:      11146EAsf                 len 9
   u64be Base58Check:    111111111111111111111111P5Gqu
                                                   len 29

   (128, 128, 128)
   VLQ hex:              800180018001              len 12
   VLQ Base58Check:      8C7YmwExwNx5ec            len 14
   u64be Base58Check:    11111118sWmo4FLbjHEbjEnzXMHmERa5NBCa
                                                   len 36

   (4294967295, 4294967295, 4294967295)
   VLQ hex:              ffffffff0fffffffff0fffffffff0f
                                                   len 30
   VLQ Base58Check:      osEoy2KbuedH57FLM2iefgHze1
                                                   len 26
   u64be Base58Check:    1111QLbz7EirSWaaadRpzxGK7bVKT4xj9qbgb
                                                   len 37

   (18446744073709551615, 18446744073709551615, 18446744073709551615)
   VLQ hex:              ffffffffffffffffff01ffffffffffffffffff01ffffffffffffffffff01
                                                   len 60
   VLQ Base58Check:      6nfFzYQPHsjA2eG4KnSfivbwwR5Y2P3ht5eCHVFojh81VZ1
                                                   len 47
   u64be Base58Check:    3diW7pnisfdBvHTXP7wvW5k5Ky1e5DVuF4qQEbh
                                                   len 39
 ```

base58check of VLQ apparently becomes more efficient when the total byte size is > 9 bytes.

soerenbf: let's continue the discussion here. I don't think it's crucial to do anything, and it's simpler to decode as is, so there's also that 🤷 

Will track this thread and create an issue if we want to change something as a result.

td202: I am not sure that the binary serialization should have any bearing on the JSON representation of a LockID. Personally, I think having the JSON being an object as we have is fine (`{"accountIndex"=76,"sequenceNumber"=4,"creationOrder"=1}`), with the other reasonable JSON representation being an array: `[75,4,1]`. I think it would also make sense to define a standard user-facing representation, but I'm not sure that we necessarily have to use that as the representation for JSON. For instance, smart contract addresses are represented in JSON as `{"index"=12,"subindex"=0}` but for the user as `<12,0>`.

Also, I would really rather we don't use Base58Check for anything other than account addresses. I don't particularly like it, but one virtue that it has is that it's pretty easy to see when something's supposed to be an account address.

limemloh: I am actually not sure why we even need a json representation for the lock ID

soerenbf: wallet proxy, as the lock related events are have proto-encoded lock ids which are represented as json in the DB.

PR #928, opened 2026-04-20
Default handling for CBOR

td202:
## Purpose

Allow specifying fields with default value that will be used when the field is not present. The field will not be included in the serialization if it has the default value.

## Changes

- CBOR default support in deriving macros
- `keep_alive` field of `LockController::SimpleV0` is treated as default `false`.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/rust-src/concordium_base/src/common/cbor.rs
+++ b/rust-src/concordium_base/src/common/cbor.rs
@@ -1234,6 +1234,29 @@ mod test {
     }
 
     #[test]
+    fn test_struct_default_derived() {
+        #[derive(Debug, Eq, PartialEq, CborSerialize, CborDeserialize)]
+        struct TestStruct {
+            field1: u64,
+            #[cbor(default = 5)]
+            field2: u64,
+        }
+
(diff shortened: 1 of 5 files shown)


soerenbf (approved):

allanbrondum:
Haven't looked at the details yet, will look at it tomorrow morning, but we should implement tests covering default values in the `cbor.rs` file in `concordium_base` crate.

allanbrondum (approved):
Nice. Technically, if we want to match CDDL `.default` control, we should actually disallow decoding values that is equal to the default value: https://datatracker.ietf.org/doc/html/rfc8610#section-3.8.6. Though I don't think our `[cbor(default=...)]` attribute needs to follow those semantics that strict. I think it is fine that we allow deserializing the default value if it is set in the CBOR.

rust-src/concordium_base/src/common/cbor.rs
allanbrondum, line 1: Remember to add documentation of the attribute at the module level documentation in this file.

PR #929, opened 2026-04-22
Fix rust 1.94 clippy warnings

limemloh:
## Purpose

Building and linting rust code in the node results in a number of warnings from here, so this PR fixes these warnings for rust toolchain version 1.94.


rust-src/concordium_base/src/sigma_protocols/dlogaggequal.rs
```diff
@@ -5,6 +5,7 @@
 //! NB:
 //! This module is currently not used, and is only here as a reference.
 //! When using the code needs to be thoroughly reviewed.
+#![allow(dead_code)]
```
td202: I remember @allanbrondum not being happy with this solution. Perhaps we could use #![cfg(any())] instead?

td202 (approved):

rust-src/concordium_base/src/sigma_protocols/dlogaggequal.rs
allanbrondum: We should just remove this modules. I've talked with Daniel about it also, we should not have protocols laying around that we don't use "just in case". So we can remove these two modules:
```
// the following two modules are only there for reference if we ever need them,
// they are not used.
mod dlogaggequal;
mod dlogeq;
```

limemloh: done

allanbrondum (approved):

PR #931, opened 2026-04-27
Support for createLock

td202:
## Purpose

Support createLock.

## Changes

- Replace `TokenParameter` and `MetaUpdateParameter` with a `RawCbor` type. This is changed to use `ByteString` instead of `ShortByteString` to reduce the amount of copying.
- Add `LockCreated` and `LockDestroyed` events, with binary and JSON serialization.
- Update tests
- Rust CBOR representation of lock operations and constructors
- Remove some unnecessary serde instances for meta-update operations.
- Tweak the Rust `CoinInfo` serialization to match Haskell, with tests.

## Checklist

      hard-to-understand areas.


rust-src/concordium_base/src/protocol_level_tokens/meta_operations.rs
soerenbf: This is slightly late, I admit.. but would this not be better suited for the node repository? is there any use case outside of the node for this?

soerenbf (approved):
LGTM

rust-src/concordium_base/src/protocol_level_tokens/meta_operations.rs
td202: Done.

PR #932, opened 2026-04-27
Extend token account state cbor with locks state

soerenbf:
Ref COR-2316

## Purpose

Extend the account-token cbor state with locks data.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/src/protocol_level_tokens/token_module_account_state.rs
+++ b/rust-src/concordium_base/src/protocol_level_tokens/token_module_account_state.rs
@@ -1,7 +1,14 @@
+use crate::{
+    common::cbor::{
+        CborDecoder, CborDeserialize, CborEncoder, CborMapDecoder, CborMapEncoder,
+        CborSerializationResult, CborSerialize,
+    },
+    protocol_level_locks::LockId,
+    protocol_level_tokens::TokenAmount,
+};
 use concordium_base_derive::{CborDeserialize, CborSerialize};
 
-#[derive(Debug, Clone, PartialEq, CborSerialize, CborDeserialize, Default)]
-///  The account state represents account-specific information that is
+/// The account state represents account-specific information that is
 /// maintained by the Token Module, and is returned as part of a
 /// `GetAccountInfo` query. It does not include state that is managed by
 /// the Token Kernel, such as the token identifier and account balance.
@@ -12,6 +19,7 @@ use concordium_base_derive::{CborDeserialize, CborSerialize};
 /// a standard, and are specific to the module implementation) may be included,
 /// and their tags should be prefixed with an underscore ("_") to distinguish
 /// them as such.
+#[derive(Debug, Clone, PartialEq, Default)]
 pub struct TokenModuleAccountState {
     /// Whether the account is on the allow list.
     /// If `None`, the token does not support an allow list.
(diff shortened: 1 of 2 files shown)


rust-src/concordium_base/src/protocol_level_tokens/token_module_account_state.rs
```diff
@@ -12,42 +19,167 @@ use concordium_base_derive::{CborDeserialize, CborSerialize};
...
+        let mut map_decoder = decoder.decode_map()?;
+        let mut account_state = TokenModuleAccountState::default();
+
+        while let Some(key) = map_decoder.deserialize_key::<String>()? {
+            match key.as_str() {
+                "allowList" => account_state.allow_list = Some(map_decoder.deserialize_value()?),
+                "denyList" => account_state.deny_list = Some(map_decoder.deserialize_value()?),
+                "locks" => account_state.locks = map_decoder.deserialize_value()?,
+                "available" => account_state.available = Some(map_decoder.deserialize_value()?),
+                _ => map_decoder.skip_value()?,
```
td202: I think it's important to document in the type that this is happening - i.e. we are losing information from the CBOR. I think this is also a bit different from how in other places the choice of strict or not is made as a parameter when decoding - though I think this is fine. Personally, I would probably have the optional extra fields.

td202 (approved):

rust-src/concordium_base/src/protocol_level_tokens/token_module_account_state.rs
```diff
@@ -12,42 +19,167 @@ use concordium_base_derive::{CborDeserialize, CborSerialize};
...
+    /// The amount of the token controlled by the lock.
+    pub amount: TokenAmount,
 }
 
 #[cfg(test)]
 mod test {
     use super::*;
     use crate::common::cbor;
 
+    fn lock_amount() -> AccountLockAmount {
```
limemloh: I would prefer to have this inlined.
It makes the tests more local and faster to read

rust-src/concordium_base/src/protocol_level_tokens/token_module_account_state.rs
```diff
@@ -12,42 +19,167 @@ use concordium_base_derive::{CborDeserialize, CborSerialize};
...
+        AccountLockAmount {
+            lock: LockId::new(10001, 5, 0),
+            amount: TokenAmount::from_raw(12300, 3),
+        }
+    }
+
+    fn assert_round_trip(account_state: TokenModuleAccountState) {
```
limemloh: I think this should be defined using a macro_rules, because now when this errors out, the error message will point to this function and not in the line in the specific test calling it. 
Ex. the test below, if any of these fails it is hard to figure out which one of them are the one failing.

limemloh (approved):

rust-src/concordium_base/src/protocol_level_tokens/token_module_account_state.rs
soerenbf: good point 🙂

rust-src/concordium_base/src/protocol_level_tokens/token_module_account_state.rs
soerenbf: yeah good point, will have another look

PR #933, opened 2026-05-06
Locks rust sdk

soerenbf:
Ref COR-2307

## Purpose

Updates functionality to enable creation of `LockClient` in the rust SDK.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/haskell-src/Concordium/Types/Locks.hs
+++ b/haskell-src/Concordium/Types/Locks.hs
@@ -17,9 +17,9 @@ import qualified Data.ULEB128 as ULEB128
 
 -- | Lock identifier: a trio of numbers that together uniquely identify a lock.
 data LockId = LockId
-    { liAccountIndex :: !Word64
-    , liSequenceNumber :: !Word64
-    , liCreationOrder :: !Word64
+    { liAccountIndex :: !Word64,
+      liSequenceNumber :: !Word64,
+      liCreationOrder :: !Word64
     }
     deriving (Eq)
 
@@ -34,9 +34,9 @@ instance Read LockId where
 instance AE.ToJSON LockId where
     toJSON LockId{..} =
         AE.object
-            [ "accountIndex" AE..= liAccountIndex
-            , "sequenceNumber" AE..= liSequenceNumber
-            , "creationOrder" AE..= liCreationOrder
+            [ "accountIndex" AE..= liAccountIndex,
+              "sequenceNumber" AE..= liSequenceNumber,
+              "creationOrder" AE..= liCreationOrder
             ]
 
 instance AE.FromJSON LockId where
--- a/haskell-src/Concordium/Types/Locks.hs
+++ b/haskell-src/Concordium/Types/Locks.hs
@@ -5,46 +5,38 @@
 -- | Types associated with protocol-level locks.
 module Concordium.Types.Locks where
 
+import Control.Monad (guard)
 import qualified Data.Aeson as AE
+import qualified Data.ByteString as BS
+import qualified Data.ByteString.Char8 as BS8
 import qualified Data.Serialize as S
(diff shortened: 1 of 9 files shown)


rust-src/concordium_base/src/transactions.rs
```diff
@@ -2695,23 +2715,60 @@ pub mod construct {
...
+                    | MetaUpdateOperation::RevokeAdminRoles(_) => cost::PLT_ASSIGN_REVOKE_ROLES,
+                    MetaUpdateOperation::UpdateMetadata(_) => cost::PLT_UPDATE_TOKEN_METADATA,
+                    MetaUpdateOperation::LockFund(_) => cost::PLT_LOCK_FUND,
+                    MetaUpdateOperation::LockSend(_) => cost::PLT_LOCK_SEND,
+                    MetaUpdateOperation::LockReturn(_) => cost::PLT_LOCK_RETURN,
+                    MetaUpdateOperation::LockCreate(_) => cost::PLT_LOCK_CREATE,
+                    MetaUpdateOperation::LockCancel(_) => cost::PLT_LOCK_CANCEL,
+                })
+                .sum()
+    }
```
td202, line 2745: No. Smart contract interaction means the energy cost of operations will not be fixed, so we shouldn't be doing this kind of thing.

haskell-src/Concordium/Types/Locks.hs
```diff
@@ -8,14 +8,38 @@ module Concordium.Types.Locks where
...
+
+instance Show LockId where
+    show LockId{..} =
+        "<"
+            ++ show liAccountIndex
+            ++ ", "
+            ++ show liSequenceNumber
+            ++ ", "
+            ++ show liCreationOrder
+            ++ ">"
```
td202: Are we sure about this format for displaying Lock Ids? It seems a bit close to the contract address format, so could be confusing.

rust-src/concordium_base/src/transactions.rs
soerenbf: The same can be said for phase 2 mint/burn, right? I feel like we could start with this simple representation, and then look at how it needs to change when we have a better idea of what we're doing wrt. contracts. Could have probably mentioned this approach somewhere, but is that reasonable?

haskell-src/Concordium/Types/Locks.hs
soerenbf: For reference, a discussion on this has been opened elsewhere. Will update with the result of that.

soerenbf: We decided to go with `P0L1T2L`.

rust-src/concordium_base/src/protocol_level_locks/lock_id.rs
```diff
@@ -26,6 +30,55 @@ pub struct LockId {
...
+        leb128::write::unsigned(&mut bytes, self.creation_order).map_err(|_| std::fmt::Error)?;
+        f.write_str(&bs58::encode(bytes).with_check_version(3).into_string())
+    }
+}
+
+impl FromStr for LockId {
+    type Err = LockIdFromStrError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let bytes = bs58::decode(s.trim())
```
limemloh: I would expect the caller to trim the string if necessary, and not this implementation.
It might be desirable to fail for strings which contain spaces

limemloh (approved):

rust-src/concordium_base/src/protocol_level_locks/lock_id.rs
soerenbf: true, this (the `.trim`) was a remnant from something else, good catch 🙂

PR #948, opened 2026-07-31
Ar/cor 2467 make concordium std compile with wasm32v1 2

allanbrondum:
## Purpose

Changes related to using the target wasm32v1-none for smart contracts. The PR is mostly update of Rust docs and changing MSRV to 1.85. Technically the target wasm32-unknown-unkown is still supported, but the change in MSRV means we cannot use it in smart contracts.

## Changes

* Update Rust docs to refer to the the new target wasm32v1-none
* Rewrite imports to not use `std` but `alloc` and `core` crates (simplifies feature flag conditionals)
* Fixed some small clippy warnings
* Put all crates in `smart_contracts` folder into a workspace build with Rust 1.85
* Removed examples in `rust-contracts/example-contracts` since they were not maintained and not used
* Set MSRV for concordium contracts common and concordium base to 1.85

## Checklist

      hard-to-understand areas.


smart-contracts/wasm-chain-integration/Cargo.toml
```diff
@@ -68,8 +68,5 @@ harness = false
 name = "trie_benches"
 harness = false
 
-[profile.release]
-codegen-units = 1
```
allanbrondum, line 72: Is applied in `concordium-node/plt/node-rust-library`

smart-contracts/README.md
```diff
@@ -3,54 +3,14 @@
...
-simple `Wasm` traps on any panic that occurs. This option can be specified
-either in `.cargo/config` as exemplified in
-[counter/.cargo/config](./rust-contracts/example-contracts/counter/.cargo/config),
-or in the `Cargo.toml` file as
+### Compilation options for smart contracts
 
-```
-[profile.release]
-# Don't unwind on panics, just trap.
-panic = "abort"
```
allanbrondum, line 41: Removed description of panic in profile, wasm32v1-none does not support unwinding at all, so the parameter is no longer relevant.

smart-contracts/README.md
```diff
@@ -157,26 +117,19 @@ trait SchemaType {
 }
 ```
 
-To build the schema, the `Cargo.toml` must include the `build-schema` feature, which is used by the contract building tool.
-```toml
-...
-[features]
-build-schema = []
```
allanbrondum, line 164: Must be some old documentation, the feature is not needed in the smart contract

smart-contracts/contracts-common/concordium-contracts-common/Cargo.toml
```diff
@@ -14,7 +14,7 @@ readme = "../README.md"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [dependencies]
 arbitrary = { version = "1.3", features = ["derive"], optional = true }
-base64 = "0.21"
+base64 = { version = "0.21", optional = true }
```
allanbrondum, line 17: Change to optional, `concordium-contracts-common` could not be compiled with no_std before

.github/workflows/build-test-contracts-common.yaml
allanbrondum, line 1: Jobs now part of .github/workflows/build-test-smart-contracts.yaml

smart-contracts/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[workspace]
+resolver = "3"
+
+members = [
+   "contracts-common/concordium-contracts-common",
+   "contracts-common/concordium-contracts-common-derive",
```
soerenbf, line 6: Looking through the code, there's also crates in `testdata/contracts`. Do we know what these are used for? Currently if I try to build them in the context of this PR, they fail with "current package believes it's in a workspace when it's not"

smart-contracts/rust-contracts/example-contracts/context-test/src/lib.rs
soerenbf, line 1: I assume it was agreed to remove these? It makes more sense to have examples in the smart contract repo, which we also already have. So removing it from here in any case seems like a good choice to me as well.

soerenbf (approved):
Looks good IMO.

smart-contracts/rust-contracts/example-contracts/context-test/src/lib.rs
allanbrondum: I could see that they had not been updated for a while (when old test infrastructure was removed e.g.). And they did not seems to have any role either. So I discussed with Emil and we decided to remove them.

smart-contracts/Cargo.toml
allanbrondum: I've excluded them from the workspace now. But they generally don't compile, I think they have just been used to generate the binaries in testdata.

PR #955, opened 2026-08-26
Add nested CBOR limit

soerenbf:
## Purpose

Add a nesting limit to the cbor decoder. 128 was chosen as the value, as this is also used as a nesting limit in serde_json, and it seems like a reasonable upper bound.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/src/common/cbor.rs
+++ b/rust-src/concordium_base/src/common/cbor.rs
@@ -329,15 +329,38 @@ pub enum UnknownMapKeys {
     Fail,
 }
 
-/// Options applied when serializing and deserializing
-#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Default)]
+/// Options applied when serializing and deserializing.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
 pub struct SerializationOptions {
     pub unknown_map_keys: UnknownMapKeys,
+    /// Maximum depth of nested CBOR arrays, maps, and tags.
+    pub max_nesting_depth: usize,
+}
+
+impl Default for SerializationOptions {
+    fn default() -> Self {
+        Self {
+            unknown_map_keys: UnknownMapKeys::default(),
+            max_nesting_depth: 128,
+        }
+    }
 }
 
 impl SerializationOptions {
+    /// Sets how unknown keys in decoded CBOR maps are handled.
     pub fn unknown_map_keys(self, unknown_map_keys: UnknownMapKeys) -> Self {
-        Self { unknown_map_keys }
+        Self {
+            unknown_map_keys,
+            ..self
+        }
+    }
+
+    /// Sets the finite maximum depth of nested CBOR arrays, maps, and tags.
+    pub fn max_nesting_depth(self, max_nesting_depth: usize) -> Self {
+        Self {
+            max_nesting_depth,
+            ..self
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


rust-src/concordium_base/src/common/cbor.rs
```diff
@@ -491,6 +519,21 @@ pub trait CborDeserialize {
     where
         Self: Sized;
 
+    /// Deserialize while preserving the current CBOR structural nesting depth.
+    ///
+    /// Container decoders use this to propagate the current depth when recursively
+    /// decoding dynamic values. The default preserves compatibility for types that
+    /// do not track nesting depth.
```
allanbrondum, line 526: I think the doc can be a lot clearer if we don't let Claude write it :)
```suggestion
    /// Deserialize while constraining the nesting depth of CBOR composites like arrays, maps and tags. The parameter `nesting_depth` is the current nesting depth of the deserialization. If the max nesting depth is exceeded, deserialization must stop and an error must be returned.
    ///
    /// This function should be implemented by types that can represent dynamic CBOR, like the types in the module [`value`]. It is generally not needed for types that represent a "static" CBOR model, since they have an inherent nesting depth in the type itself.
```

allanbrondum: I suggest we update the doc on the map and array decode functions below also.

rust-src/concordium_base/src/common/cbor.rs
```diff
@@ -941,6 +1017,29 @@ mod test {
...
+        assert_eq!(options.unknown_map_keys, UnknownMapKeys::Fail);
+    }
+
+    #[test]
+    fn nesting_limit_error_includes_limit() {
+        assert_eq!(
+            CborSerializationError::nesting_limit_exceeded(64).to_string(),
+            "maximum nesting depth of 64 exceeded"
+        );
+    }
```
allanbrondum, line 1041: I'm not sure these tests add much value? I assume Claude just added them just to add tests?

rust-src/concordium_base/src/common/cbor.rs
```diff
@@ -765,6 +830,17 @@ pub trait CborArrayDecoder {
     /// The total number of elements that can be deserialized is equal
     /// to [`Self::size`].
     fn deserialize_element<T: CborDeserialize>(&mut self) -> CborSerializationResult<Option<T>>;
+
+    /// Deserialize an array element while propagating the current structural nesting depth.
+    ///
+    /// Implementations that support depth-aware dynamic values should forward
+    /// `nesting_depth` to [`CborDeserialize::deserialize_with_nesting_depth`].
+    fn deserialize_element_with_nesting_depth<T: CborDeserialize>(
```
allanbrondum, line 838: I don't understand why we need two variants here? I can understand it for `CborDeseralize` that is implemented on types representing CBOR. But why does the decoder need multiple variants?

allanbrondum (commented):
I think we could just have a depth as a state in `Decoder` that is incremented and checked when `decode_map` and `decode_array` is called? The returned array and map decoders defines the scope of the incremented depth and will hence decrement again when dropped. And also apply the depth check to `Decoder::skip_data_item` as done in this PR. That seems more simple, but not sure if I'm missing something. Then the only gap left open I think is tag decoding. But maybe `decode_tag` can return an decoder like like for arrays and maps, that defines a scope of incremented depth. Alternatively, tag nesting depth can be handled as a special case in `value` module.

rust-src/concordium_base/src/common/cbor.rs
soerenbf: Nope, you're right

soerenbf:
> I think we could just have a depth as a state in `Decoder` that is incremented and checked when `decode_map` and `decode_array` is called? The returned array and map decoders defines the scope of the incremented depth and will hence decrement again when dropped. And also apply the depth check to `Decoder::skip_data_item` as done in this PR. That seems more simple, but not sure if I'm missing something. Then the only gap left open I think is tag decoding. But maybe `decode_tag` can return an decoder like like for arrays and maps, that defines a scope of incremented depth. Alternatively, tag nesting depth can be handled as a special case in `value` module.

Will try to revise to this and see if it's feasible. Thanks for the feedback 🙂

PR #957, opened 2026-08-28
Nested CBOR accounting v2

soerenbf:
Replaces #955 

## Purpose

Add a nesting limit to the cbor decoder. 128 was chosen as the value, as this is also used as a nesting limit in `serde_json`, and it seems like a reasonable upper bound.

## Changes

- Added a configurable nesting limit for CBOR arrays, maps, and tags, defaulting to 128.
- Added scoped tag decoding through `CborTagDecoder`.
- Applied the limit to derived, custom, dynamic `Value`, and skipped-item decoding.

`Decoder` could not carry the nesting depth directly because the decoding flow shares the same decoder reference across nested scopes. Instead, a `DecoderContext` wraps the decoder reference and carries the depth for each scope. DecoderContext implements the same `CborDecoder` API, limiting user-facing changes.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/src/common/cbor/decoder.rs
+++ b/rust-src/concordium_base/src/common/cbor/decoder.rs
@@ -1028,8 +1028,7 @@ mod test {
             nested_tag(129),
         ] {
             let mut decoder = Decoder::new(cbor.as_slice(), SerializationOptions::default());
-            let error = decoder.skip_data_item().unwrap_err();
-            assert_eq!(error.to_string(), "maximum nesting depth of 128 exceeded");
+            assert!(decoder.skip_data_item().is_err());
         }
     }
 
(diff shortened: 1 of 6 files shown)


rust-src/concordium_base/src/common/cbor/decoder.rs
```diff
@@ -22,15 +22,37 @@ impl<R: Read> Decoder<R> {
     }
 }
 
-impl<'a, R: Read> CborDecoder for &'a mut Decoder<R>
+/// A CBOR decoder borrow with its immutable nesting depth.
+pub struct DecoderContext<'a, R: Read> {
+    decoder: &'a mut Decoder<R>,
+    depth: usize,
+}
```
soerenbf, line 29: It was not possible to simply extend `Decoder` to carry the `depth` data, as deserialization shares the same `Decoder` reference throughout. Doing this would require implementing custom `Drop` functionality and would require users to manage scopes themselves, i.e.

```rs
{
  let map_decoder = decoder.decode_map()?;
  ..
}
..
```

allanbrondum: But isn't that also what callers are already doing? I'll just try the alternative and get back.

PR #959, opened 2026-09-03
Csec 68 concordium nodeplt limit cbor nesting while decoding

allanbrondum:
## Purpose

Implement CBOR deserialization nesting limit.

## Changes

## Checklist

      hard-to-understand areas.

### DIFF
--- a/rust-src/concordium_base/src/common/cbor/decoder.rs
+++ b/rust-src/concordium_base/src/common/cbor/decoder.rs
@@ -12,13 +12,37 @@ use std::{io::Cursor, iter};
 pub struct Decoder<R: Read> {
     inner: ciborium_ll::Decoder<R>,
     options: SerializationOptions,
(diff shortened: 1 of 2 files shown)


rust-src/concordium_base/src/common/cbor/decoder.rs
```diff
@@ -502,6 +544,7 @@ mod test {
         let elm3: Option<u32> = array_decoder.deserialize_element().unwrap();
         assert_eq!(elm3, None);
         assert_eq!(array_decoder.size(), Some(2));
+        drop(array_decoder);
```
soerenbf, line 547: I don't think relying on consumers to manually drop the decoder to decrement the nesting depth is feasible though, which is why I went with the `DecoderContext` approach instead. If you implement `CborDeserialize` without derive, you need to manage this manually.

